PR #5, opened 2018-12-20
chore(Dockerfile): standardize dockerfile and add rhel7 dockerfile

ericavonb:


enj:
/lgtm

Alfred and machine approver probably need the same treatment.

ericavonb:
@enj if you could give another lgtm :)

PR #7, opened 2018-12-21
add auth oauth resources to manifests

sallyom:
This PR adds the authentication and oauth CRDs
@enj is this what you'll need? or no CRs?

### DIFF
--- a/manifests/0000_09_cluster-osin-operator_06_authentication_crd.yaml
+++ b/manifests/0000_09_cluster-osin-operator_06_authentication_crd.yaml
@@ -1,0 +1,18 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: authentications.config.openshift.io
+spec:
+  group: config.openshift.io
+  names:
+    kind: Authentication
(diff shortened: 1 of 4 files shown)


enj:
I accidentally lost the run level changes when I rebased so you will want to use the 658607fddb875a9c37b6313a26460e321e7b066f naming format.

Otherwise this is fine for now, we will want to figure out the final resting place for the CRDs and CRs at some point before beta.

manifests/0000_09_cluster-osin-operator_08_auth_cr.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+items:
```
enj: uh just realized this is a list instead of a single item, same with other CR

sallyom: : ) /hold

sallyom: yup, cluster is much happier now,  i guess it wasn't the flaky CI, it was my flakiness, time for a break :)

PR #12, opened 2019-01-08
[WIP] Conversions of config->osin identity provider types

stlaz:
This PR adds the conversions of config IDP types to osin config types

### TODO
- [ ] add tests
- [ ] move the code that's supposed to be moved


pkg/operator2/oauth.go
```diff
@@ -136,3 +148,172 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+			CA:       getLocalConfigMapFilename(gitlabConfig.CA),
+			URL:      gitlabConfig.URL,
+			ClientID: gitlabConfig.ClientID,
+			ClientSecret: configv1.StringSource{
+				// TODO: this could be Env, Value, File, either can be encrypted
+				StringSourceSpec: configv1.StringSourceSpec{
+					File: getLocalSecretFilename(gitlabConfig.ClientSecret),
+				},
+			},
+			Legacy: &gitlabConfig.LegacyOAuth2, // TODO: better mapping of 2 to 3 values?
```
stlaz: I guess this was turned into a pointer in OSIN config for a reason but there is probably no way to reach the `nil` value?

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,172 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+			},
+			ExtraScopes:              openIDConfig.ExtraScopes,
+			ExtraAuthorizeParameters: openIDConfig.ExtraAuthorizeParameters,
+			URLs: osinv1.OpenIDURLs{ // TODO: do we need this to be different from configv1?
+				Authorize: openIDConfig.URLs.Authorize,
+				Token:     openIDConfig.URLs.Token,
+				UserInfo:  openIDConfig.URLs.UserInfo,
+			},
+			Claims: osinv1.OpenIDClaims{ // TODO: do we need this to be different from configv1?
+				ID:                []string{"sub"}, // TODO: this is missing in configv1 version?
```
stlaz: Is this the correct default? It's supposed to be required but there is no such field in the `config` types.

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,172 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+		serializer.Encode(&p, bytesWriter)
+
+	case configv1.IdentityProviderTypeLDAP:
+		ldapConfig := providerConfig.LDAP
+		p := osinv1.LDAPPasswordIdentityProvider{
+			URL:    ldapConfig.URL,
+			BindDN: ldapConfig.BindDN,
+			BindPassword: configv1.StringSource{
+				StringSourceSpec: configv1.StringSourceSpec{
+					// TODO: this could be Env, Value, File, either can be encrypted
```
stlaz: We have a bunch of options for `StringSources`, do we just go with the secrets being files?

pkg/operator2/oauth.go
```diff
@@ -15,6 +19,10 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
+func init() { // TODO: probably should be moved somewhere else
+	osinv1.AddToScheme(scheme.Scheme)
```
enj:
```go
var (
	scheme = runtime.NewScheme()
	codecs = serializer.NewCodecFactory(scheme)
)

func init() {
	utilruntime.Must(osinv1.Install(scheme))
}
```

pkg/operator2/oauth.go
```diff
@@ -37,22 +45,26 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
-			ProviderSelection: "",
-			Error:             "",
+			Login:             getLocalSecretFilename(oauthConfig.Spec.Templates.Login),
+			ProviderSelection: getLocalSecretFilename(oauthConfig.Spec.Templates.ProviderSelection),
+			Error:             getLocalSecretFilename(oauthConfig.Spec.Templates.Error),
 		}
 	}
 
 	identityProviders := make([]osinv1.IdentityProvider, 0, len(oauthConfig.Spec.IdentityProviders))
 	for _, idp := range oauthConfig.Spec.IdentityProviders {
```
enj: `len(oauthConfig.Spec.IdentityProviders) == 0` needs to imply `DenyAll`

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
 		},
 	}, nil
 }
+
+func convertProviderConfigToOsinBytes(providerConfig *configv1.IdentityProviderConfig) []byte {
```
enj: Move to `idp.go` in the same folder for now.

pkg/operator2/oauth.go
enj: Hard code this to `new(bool)` and add a comment saying `we require OIDC for GitLab now`

pkg/operator2/oauth.go
enj: Yes, expect all secrets to be files on disk.  The config needs to be safe to put in a config map.  Make a helper to reduce `configv1.StringSource` boilerplate.

pkg/operator2/oauth.go
enj: Set this to `[]string{configv1.UserIDClaim}` - we no longer support setting this because it is unsafe to do so.

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+				ID:                []string{"sub"}, // TODO: this is missing in configv1 version?
+				PreferredUsername: openIDConfig.Claims.PreferredUsername,
+				Name:              openIDConfig.Claims.Name,
+				Email:             openIDConfig.Claims.Email,
+			},
+		}
+		serializer.Encode(&p, bytesWriter)
+	} // switch
+
+	bytesWriter.Flush()
```
enj: `defer` this at the top.

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+				CA:  getLocalConfigMapFilename(basicAuthConfig.CA),
+				CertInfo: configv1.CertInfo{
+					CertFile: getLocalSecretFilename(basicAuthConfig.TLSClientCert),
+					KeyFile:  getLocalSecretFilename(basicAuthConfig.TLSClientKey),
+				},
+			},
+		}
+		serializer.Encode(&p, bytesWriter)
+
+	case configv1.IdentityProviderTypeAllowAll:
```
enj: Remove.

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
 	}, nil
 }
+
+func convertProviderConfigToOsinBytes(providerConfig *configv1.IdentityProviderConfig) []byte {
+	var providerConfigBytes bytes.Buffer
+
+	bytesWriter := bufio.NewWriter(&providerConfigBytes)
+	serializer := kubejson.NewYAMLSerializer(kubejson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+
+	switch providerConfig.Type {
```
enj: `case`s need to be in alphabetical order for my sanity.

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+				},
+			},
+		}
+		serializer.Encode(&p, bytesWriter)
+
+	case configv1.IdentityProviderTypeAllowAll:
+		// TODO: is the type necessarily set or would it rather be just the bool?
+		serializer.Encode(&osinv1.AllowAllPasswordIdentityProvider{}, bytesWriter)
+
+	case configv1.IdentityProviderTypeDenyAll:
```
enj: Remove, the use of this is implied as noted in other comment.

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+
+func convertProviderConfigToOsinBytes(providerConfig *configv1.IdentityProviderConfig) []byte {
+	var providerConfigBytes bytes.Buffer
+
+	bytesWriter := bufio.NewWriter(&providerConfigBytes)
+	serializer := kubejson.NewYAMLSerializer(kubejson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+
+	switch providerConfig.Type {
+	case configv1.IdentityProviderTypeBasicAuth:
+		basicAuthConfig := providerConfig.BasicAuth
```
enj: Add comment `// TODO we need validation to make sure this cannot be nil (this applies to all IDPs)`

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+			},
+			Claims: osinv1.OpenIDClaims{ // TODO: do we need this to be different from configv1?
+				ID:                []string{"sub"}, // TODO: this is missing in configv1 version?
+				PreferredUsername: openIDConfig.Claims.PreferredUsername,
+				Name:              openIDConfig.Claims.Name,
+				Email:             openIDConfig.Claims.Email,
+			},
+		}
+		serializer.Encode(&p, bytesWriter)
+	} // switch
```
enj: Needs a default `case` which errors.

pkg/operator2/oauth.go
enj: This will need to be able to `error`.  At the caller, for now, just log the error and keep going to the next IDP.

pkg/operator2/oauth.go
```diff
@@ -136,3 +148,173 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+
+	bytesWriter.Flush()
+	return providerConfigBytes.Bytes()
+}
+
+// TODO: move to some kind of helpers file, possibly branch to Local/Remote NS versions if necessary
+// TODO: the logic of naming the CMs and Secrets to be mounted to the OSIN container
+// TODO: we need to keep track of everything that needs to be mounted in the containers
+func getLocalConfigMapFilename(ref configv1.ConfigMapReference) string {
```
enj: Both of these helper functions should return a string:

`/var/config/user/idp/<index_of_idp>/<configmap_or_secret>/<name_of_key>`

I believe that is enough for us to generally identify everything.  We would mount it at `/var/config/user/idp/<index_of_idp>/<configmap_or_secret>`.

I could see some argument for replacing `<index_of_idp>` with `<name_of_idp>`.  Either way, enough to get started.

As a test, I would like to see what the generated config looks like.

enj: Hmm actually it needs the name of the config map or secret as well since they could contain conflicting keys.

pkg/operator2/oauth.go
stlaz: You need to do the flushing before you return, otherwise the bytes buffer to which the `Writer` is connected won't get filled on return.
`defer` would perform the flush after the return is done. I suspect (but did not check the implementation) that `providerConfigBytes.Bytes()` returns a copy of its byte slice rather than the pointer to its internal slice, which would mean that if we deferred the `Flush()`, we'd always return an empty slice (unless we reached the default `Writer` buffer size - 4096.

https://play.golang.org/p/TfhNxdUrYjS

stlaz:
First bunch of fixes.

pkg/operator2/oauth.go
stlaz: I thought we would rather have a more general map of (config_map, key) -> (filename) which we supply to the method taking care of creating the Deployment object so that this one can use it to add proper mounting mapping.

I think the `name/index_of_idp` might be irrelevant. All you want is your config maps to be mounted into right filenames, irregardless of their purpose?

pkg/operator2/oauth.go
stlaz: Sorted. I originally chose the same order as in the openshift/api version of `types_oauth.go`

stlaz:
... and second bunch of fixes. I added some comments and ideas above.

stlaz:
Creating a libvirt cluster fails for me so I added a unit test for `BasicAuth` to prove the conversions should work.
I have changes stashed for after the https://github.com/openshift/api/pull/150 gets merged.

pkg/operator2/helpers.go
```diff
@@ -0,0 +1,21 @@
...
+
+func moveLocalSecretToFileStringSource(ref configv1.LocalSecretReference) configv1.StringSource {
+	return configv1.StringSource{
+		StringSourceSpec: configv1.StringSourceSpec{
+			File: getLocalSecretFilename(ref),
+		},
+	}
+}
+
+// TODO: the logic of naming the CMs and Secrets to be mounted to the OSIN container
```
ericavonb: Since we're using local references only, we don't have to worry about naming collisions between any of the secrets or configmaps referenced.
It's possible there's an existing secret or configmap in the target namespace. I don't see an easy way to guarantee there won't be any collisions; the ones I can think of would make syncing updates and deletions more annoying and difficult. Some OK solutions would be like appending the name with some moniker, `operator-config/mycm` -> `cluster-integrated-auth/mycm-clusterauth-v4` or blacklisting certain names.

It might be good to go the content-addressable route. Use a consistent hash of the data in the name + use ownerrefs to GC unused cm & secrets. @enj thoughts?

enj: I will explore options.

enj:
/retest

Just trying to get this thing to rerun tests because #15 merged.

stlaz:
The types of arguments of
```go
func getLocalConfigMapFilename(ref configv1.ConfigMapReference) string {
	return ""
}

func getLocalSecretFilename(ref configv1.LocalSecretReference) string {
	return ""
}
```
will need to change type. You should be able to push to my branch here.

enj:
@stlaz just fix it up tomorrow and we will merge.

PR #13, opened 2019-01-08
validate session secrets

sallyom:
@enj this PR adds validation for session secrets

### DIFF
--- a/pkg/operator2/secret.go
+++ b/pkg/operator2/secret.go
@@ -3,6 +3,7 @@ package operator2
 import (
 	"crypto/rand"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/json"
 
 	"github.com/golang/glog"
@@ -13,33 +14,61 @@ import (
 	legacyconfigv1 "github.com/openshift/api/legacyconfig/v1"
 )
 
-func (c *osinOperator) expectedSessionSecret() *v1.Secret {
+func (c *osinOperator) expectedSessionSecret() (*v1.Secret, error) {
 	secret, err := c.secrets.Secrets(targetName).Get(targetName, metav1.GetOptions{})
-	if err != nil || isInvalidSessionSecret(secret) {
+	if err != nil || !isValidSessionSecret(secret) {
 		glog.V(4).Infof("failed to get secret %s: %v", targetName, err)
-		return randomSessionSecret()
+		generatedSessionSecret, err := randomSessionSecret()
+		if err != nil {
+			return nil, err
+		}
+		return generatedSessionSecret, nil
 	}
-	return secret
+	return secret, nil
 }
 
-func isInvalidSessionSecret(secret *v1.Secret) bool {
-	// TODO add validation
-	return false
+func isValidSessionSecret(secret *v1.Secret) bool {
+	// TODO add more validation?
+	var sessionSecretsBytes [][]byte
+	for _, v := range secret.Data {
+		sessionSecretsBytes = append(sessionSecretsBytes, v)
+	}
+	for _, ss := range sessionSecretsBytes {
+		var sessionSecrets *legacyconfigv1.SessionSecrets
+		err := json.Unmarshal(ss, &sessionSecrets)
+		if err != nil {
+			return false
+		}
+		for _, s := range sessionSecrets.Secrets {
+			if len(s.Authentication) != 64 {
+				return false
+			}
+
+			if len(s.Encryption) != 32 {
+				return false
+			}
+		}
+	}
+	return true
 }
 
-func randomSessionSecret() *v1.Secret {
+func randomSessionSecret() (*v1.Secret, error) {
+	skey, err := newSessionSecretsJSON()
+	if err != nil {
+		return nil, err
+	}
 	return &v1.Secret{
 		ObjectMeta: defaultMeta(),
 		Data: map[string][]byte{
-			sessionKey: newSessionSecretsJSON(),
+			sessionKey: skey,
 		},
-	}
+	}, nil
 }
 
-func newSessionSecretsJSON() []byte {
+func newSessionSecretsJSON() ([]byte, error) {
 	const (
-		sha256KeyLenBits = sha256.BlockSize * 8 // max key size with HMAC SHA256
-		aes256KeyLenBits = 256                  // max key size with AES (AES-256)
+		sha256KeyLenBytes = sha256.BlockSize // max key size with HMAC SHA256
+		aes256KeyLenBytes = 32               // max key size with AES (AES-256)
 	)
 
 	secrets := &legacyconfigv1.SessionSecrets{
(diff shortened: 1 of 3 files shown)


pkg/operator2/secret.go
```diff
@@ -17,14 +17,35 @@ func (c *osinOperator) expectedSessionSecret() *v1.Secret {
...
+		if err != nil {
+			return false
+		}
+		for _, s := range sessionSecrets.Secrets {
+			if len(s.Authentication) != len(s.Encryption) {
+				return false
+			}
+		}
+	}
+	return true
```
sallyom: I'm testing this validation locally, now, hold off on reviewing for now so i don't feel dumb : ) thanks.

pkg/operator2/secret.go
```diff
@@ -49,11 +70,12 @@ func newSessionSecretsJSON() []byte {
 		},
 		Secrets: []legacyconfigv1.SessionSecret{
 			{
-				Authentication: string(randomBits(sha256KeyLenBits)), // TODO these lengths are probably wrong
-				Encryption:     string(randomBits(aes256KeyLenBits)),
+				Authentication: string(randomBits(sha256KeyLenBits)), // 32 chars
```
enj: I think the issue here is that `randomBits` is just noise - it is not valid UTF8 so that cast to `string` is not valid.  I believe we have to base64 encode the value for it to be safe - and the length of the final base64 encoded string needs to be correct (so when osin casts it to `[]byte` it works out correctly).

pkg/operator2/secret.go
```diff
@@ -17,14 +17,35 @@ func (c *osinOperator) expectedSessionSecret() *v1.Secret {
...
+		sessionSecretsBytes = append(sessionSecretsBytes, v)
+	}
+	for _, ss := range sessionSecretsBytes {
+		var sessionSecrets *legacyconfigv1.SessionSecrets
+		err := json.Unmarshal(ss, &sessionSecrets)
+		if err != nil {
+			return false
+		}
+		for _, s := range sessionSecrets.Secrets {
+			if len(s.Authentication) != len(s.Encryption) {
```
enj: I do not think this is required.  We want `Authentication` to be `64` in length and `Encryption` to be `32`.

sallyom: I thought they had to be the same length, but now I see i'm looking at outdated ansible docs, so not surprised i'm wrong, k

pkg/operator2/secret.go
sallyom: > the length of the final base64 encoded string needs to be correct

and that is Authentication: 64chars and Encryption: 32 chars?
Where is that requirement coming from?

enj:
> and that is Authentication: 64chars and Encryption: 32 chars?
> Where is that requirement coming from?

Master config validation logic.

pkg/operator2/secret.go
```diff
@@ -17,14 +17,47 @@ func (c *osinOperator) expectedSessionSecret() *v1.Secret {
...
+		var sessionSecrets *legacyconfigv1.SessionSecrets
+		err := json.Unmarshal(ss, &sessionSecrets)
+		if err != nil {
+			return false
+		}
+		for _, s := range sessionSecrets.Secrets {
+			switch {
+			case len(s.Authentication) == 0:
+				return false
+			case len(s.Authentication) < 32:
```
enj: We are generating this ourselves so let us be very opinionated and require `64` as the length (this is the max key size before the key gets hashed).

pkg/operator2/secret.go
```diff
@@ -17,14 +17,47 @@ func (c *osinOperator) expectedSessionSecret() *v1.Secret {
...
+				return false
+			}
+
+			switch len(s.Encryption) {
+			case 0:
+				return false
+			case 16, 24, 32:
```
enj: Same here, require `32`.

pkg/operator2/secret.go
```diff
@@ -17,14 +17,47 @@ func (c *osinOperator) expectedSessionSecret() *v1.Secret {
 	secret, err := c.secrets.Secrets(targetName).Get(targetName, metav1.GetOptions{})
 	if err != nil || isInvalidSessionSecret(secret) {
 		glog.V(4).Infof("failed to get secret %s: %v", targetName, err)
-		return randomSessionSecret()
+		generatedSessionSecret := randomSessionSecret()
+		if isInvalidSessionSecret(generatedSessionSecret) {
+			panic(err) // should never happen
```
enj: `panic("invalid session secret generated")`

pkg/operator2/secret.go
```diff
@@ -49,11 +82,12 @@ func newSessionSecretsJSON() []byte {
 		},
 		Secrets: []legacyconfigv1.SessionSecret{
 			{
-				Authentication: string(randomBits(sha256KeyLenBits)), // TODO these lengths are probably wrong
-				Encryption:     string(randomBits(aes256KeyLenBits)),
+				Authentication: string(randomBits(sha256KeyLenBits)), // 64 chars
```
enj: Still need to use `base64.RawURLEncoding.EncodeToString` instead of blind cast to `string`.

pkg/operator2/secret.go
sallyom: ok, fixed

pkg/operator2/secret.go
sallyom: got it, fixed

pkg/operator2/secret.go
sallyom: done

pkg/operator2/secret.go
sallyom: got it

pkg/operator2/secret.go
```diff
@@ -17,13 +18,37 @@ func (c *osinOperator) expectedSessionSecret() *v1.Secret {
 	secret, err := c.secrets.Secrets(targetName).Get(targetName, metav1.GetOptions{})
 	if err != nil || isInvalidSessionSecret(secret) {
 		glog.V(4).Infof("failed to get secret %s: %v", targetName, err)
-		return randomSessionSecret()
+		generatedSessionSecret := randomSessionSecret()
+		if isInvalidSessionSecret(generatedSessionSecret) {
+			panic("invalid session secret generated") // should never happen
```
ericavonb: We shouldn't panic here. Return any errors and have the top-level sync function handle it.

pkg/operator2/secret.go
```diff
@@ -17,13 +18,37 @@ func (c *osinOperator) expectedSessionSecret() *v1.Secret {
 	secret, err := c.secrets.Secrets(targetName).Get(targetName, metav1.GetOptions{})
 	if err != nil || isInvalidSessionSecret(secret) {
 		glog.V(4).Infof("failed to get secret %s: %v", targetName, err)
-		return randomSessionSecret()
+		generatedSessionSecret := randomSessionSecret()
+		if isInvalidSessionSecret(generatedSessionSecret) {
```
ericavonb: Can we reverse this? (`if !isValidSessionSecret(...`). I almost missed that it was checking if it's *_In_valid*.

pkg/operator2/secret.go
```diff
@@ -63,14 +88,19 @@ func newSessionSecretsJSON() []byte {
...
 // needs to be in lib-go
-func randomBits(bits int) []byte {
-	size := bits / 8
-	if bits%8 != 0 {
-		size++
-	}
+func randomBytes(size int) []byte {
 	b := make([]byte, size)
 	if _, err := rand.Read(b); err != nil {
 		panic(err) // rand should never fail
```
ericavonb: Bubble up error

pkg/operator2/secret.go
sallyom: yes, I also messed myself up with this in the test, for a few minutes I was all 'wait, why is this failing?? ...oooh, it's bc I'm testing if it's _in_ valid, not valid' : )

pkg/operator2/secret.go
enj: This is supposed to die.  A failure here means the crypto part of your kernel is broken.  This panic has existed in Openshift since the very beginning with no reported failures (and it would be very obvious if a panic crashed the master).

pkg/operator2/secret.go
enj: I am not really sure why we even have this check.  Seems like it should be in a unit test or a panic.

pkg/operator2/secret.go
sallyom: ok : )

pkg/operator2/secret.go
sallyom: yea, i see your point, ok

pkg/operator2/secret.go
sallyom: kept that panic, but return the error from `json.Marshal(secrets)` in `newSessionSecretsJSON()`

PR #18, opened 2019-01-11
Honor route name assigned to our service in our config

stlaz:
This PR makes the route that we create be honored in our OAuth configuration in order for OSIN to be reachable.

I also did a check of the Route object we create according to https://docs.okd.io/latest/architecture/networking/routes.html. The behavior of the Route is described in the commit message of the second commit.

### DIFF
--- a/pkg/operator2/oauth.go
+++ b/pkg/operator2/oauth.go
@@ -13,11 +13,12 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	kubecontrolplanev1 "github.com/openshift/api/kubecontrolplane/v1"
 	osinv1 "github.com/openshift/api/osin/v1"
+	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
-func (c *authOperator) handleOAuthConfig(configOverrides []byte) (*corev1.ConfigMap, error) {
+func (c *authOperator) handleOAuthConfig(route *routev1.Route, configOverrides []byte) (*corev1.ConfigMap, error) {
 	oauthConfig, err := c.oauth.Get(configName, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
@@ -108,8 +109,8 @@ func (c *authOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
 			// which needs to direct the user to the real public URL (MasterPublicURL)
 			// that means we still need to get that value from the installer's config
 			// TODO ask installer team to make it easier to get that URL
-			MasterURL:                   "https://127.0.0.1:443",
-			MasterPublicURL:             "https://127.0.0.1:443",
+			MasterURL:                   fmt.Sprintf("https://%s", route.Spec.Host),
+			MasterPublicURL:             fmt.Sprintf("https://%s", route.Spec.Host),
 			AssetPublicURL:              "", // TODO do we need this?
 			AlwaysShowProviderSelection: false,
 			IdentityProviders:           identityProviders,
--- a/pkg/operator2/route.go
+++ b/pkg/operator2/route.go
@@ -18,11 +18,51 @@ func (c *authOperator) handleRoute() (*routev1.Route, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	if len(route.Spec.Host) == 0 {
 		return nil, fmt.Errorf("route has no host: %#v", route)
 	}
-	// TODO make sure route is correct
+
+	if err := isValidRoute(route); err != nil {
+		// delete the route so that it is replaced with the proper one in next reconcile loop
+		c.route.Delete(route.Name, &metav1.DeleteOptions{})
+		return nil, err
+	}
+
 	return route, nil
+}
+
+func isValidRoute(route *routev1.Route) error {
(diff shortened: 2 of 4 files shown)


pkg/operator2/route.go
```diff
@@ -21,7 +21,6 @@ func (c *osinOperator) handleRoute() (*routev1.Route, error) {
 	if len(route.Spec.Host) == 0 {
 		return nil, fmt.Errorf("route has no host: %#v", route)
 	}
-	// TODO make sure route is correct
```
enj: This means validate the route from the API, which we still do not do.

pkg/operator2/oauth.go
```diff
@@ -108,8 +109,8 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
 			// which needs to direct the user to the real public URL (MasterPublicURL)
 			// that means we still need to get that value from the installer's config
 			// TODO ask installer team to make it easier to get that URL
-			MasterURL:                   "https://127.0.0.1:443",
-			MasterPublicURL:             "https://127.0.0.1:443",
+			MasterURL:                   fmt.Sprintf("https://%s:%s", route.Spec.Host, route.Spec.Port.TargetPort.StrVal),
```
enj: No port.

pkg/operator2/oauth.go
```diff
@@ -108,8 +109,8 @@ func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.Config
...
+			MasterPublicURL:             fmt.Sprintf("https://%s:%s", route.Spec.Host, route.Spec.Port.TargetPort.StrVal),
```
enj: No port.

pkg/operator2/oauth.go
```diff
@@ -13,11 +13,12 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	kubecontrolplanev1 "github.com/openshift/api/kubecontrolplane/v1"
 	osinv1 "github.com/openshift/api/osin/v1"
+	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
-func (c *osinOperator) handleOAuthConfig(configOverrides []byte) (*corev1.ConfigMap, error) {
+func (c *osinOperator) handleOAuthConfig(route *routev1.Route, configOverrides []byte) (*corev1.ConfigMap, error) {
```
enj: Fix the lack of a scheme in `getMetadataConfigMap`

pkg/operator2/route.go
stlaz: Oh boy, I thought this was a weird request the way I grasped it
figure

pkg/operator2/route.go
```diff
@@ -18,13 +18,43 @@ func (c *authOperator) handleRoute() (*routev1.Route, error) {
 	if err != nil {
 		return nil, err
 	}
-	if len(route.Spec.Host) == 0 {
-		return nil, fmt.Errorf("route has no host: %#v", route)
+
+	if err := isValidRoute(route); err != nil {
```
stlaz: Should we try to reconcile the route if it does not meet our expectations?

pkg/operator2/route.go
enj: The message you had in your git commit would be reasonable comment.

pkg/operator2/route.go
enj: You could just add a delete for the route.  The validation error will cause the loop to retried and the route will be fixed on the next iteration.

pkg/operator2/route.go
```diff
@@ -18,13 +18,43 @@ func (c *authOperator) handleRoute() (*routev1.Route, error) {
...
 	return route, nil
 }
 
+func isValidRoute(route *routev1.Route) error {
+	// TODO: return all errors at once
+	if route.Spec.To.Name != targetName {
+		return fmt.Errorf("route targets a wrong service - needs %s: %#v", targetName, route)
+	}
+
+	if len(route.Spec.Host) == 0 {
```
enj: Move this check out, this is "route is not ready", not "this route is invalid"

pkg/operator2/route.go
```diff
@@ -18,13 +18,43 @@ func (c *authOperator) handleRoute() (*routev1.Route, error) {
...
+
+	if err := isValidRoute(route); err != nil {
+		return nil, err
 	}
-	// TODO make sure route is correct
+
 	return route, nil
 }
 
+func isValidRoute(route *routev1.Route) error {
```
enj: It might be simpler to do some form of deep equal.

pkg/operator2/route.go
stlaz: Found it in reflog, I'll add it back with next push.

pkg/operator2/route.go
stlaz: Did not find a deep-equal that would only check certain fileds. We might possibly add a for-loop which would use `reflect.ValueOf()` comparison, although this here seems a bit more explicit.

Nevertheless, I modified the check a bit so that it does not use hard-coded values and uses whatever is in `defaultRoute()` instead.

stlaz:
Thank you for your review and comments, I hope to address them in the latest patch.

enj:
Wake up bot.

/lgtm

PR #24, opened 2019-01-16
Get secret/configmaps information from config for all IdPs

stlaz:
Grabs identity providers' configmaps and secrets from the
top-level config and stores their mapping to volumes/volumeMounts
so that these can later be added to the deployment running OSIN

### DIFF
--- a/pkg/operator2/configsync.go
+++ b/pkg/operator2/configsync.go
@@ -91,64 +91,191 @@ type sourceData struct {
 	mount  corev1.VolumeMount
 }
 
+// TODO: new source data could be generalized
+func newSourceDataIDPSecret(index int, secretName, idpType string) (string, sourceData) {
+	dest := getIDPName(index, secretName, idpType)
+
+	volume, mount, path := secretVolume(index, dest, idpType)
+	ret := sourceData{
+		src:    secretName,
+		path:   path,
+		volume: volume,
+		mount:  mount,
+	}
+
+	return dest, ret
+}
+
+func newSourceDataIDPConfigMap(index int, cmName, idpType string) (string, sourceData) {
+	dest := getIDPName(index, cmName, idpType)
+
+	volume, mount, path := configMapVolume(index, dest, idpType)
+	ret := sourceData{
+		src:    cmName,
+		path:   path,
+		volume: volume,
+		mount:  mount,
+	}
+
+	return dest, ret
+}
+
 // TODO this should be combined with convertProviderConfigToOsinBytes as it would simplify how the data is shared
 func convertToData(idps []configv1.IdentityProvider) []idpSyncData {
 	out := make([]idpSyncData, 0, len(idps))
 	for i, idp := range idps {
 		pc := idp.IdentityProviderConfig
 		switch pc.Type {
+		case configv1.IdentityProviderTypeBasicAuth:
+			p := pc.BasicAuth
+
+			ca := p.CA.Name
+			caDest, caData := newSourceDataIDPConfigMap(i, ca, corev1.ServiceAccountRootCAKey)
+
+			clientCert := p.TLSClientCert.Name
+			clientCertDest, clientCertData := newSourceDataIDPSecret(i, clientCert, corev1.TLSCertKey)
+
+			clientKey := p.TLSClientKey.Name
+			clientKeyDest, clienKeyData := newSourceDataIDPSecret(i, clientKey, corev1.TLSPrivateKeyKey)
+
+			out = append(out,
+				idpSyncData{
+					configMaps: map[string]sourceData{caDest: caData},
+					secrets: map[string]sourceData{
+						clientCertDest: clientCertData,
+						clientKeyDest:  clienKeyData,
+					},
+				},
+			)
+
+		case configv1.IdentityProviderTypeGitHub:
+			p := pc.GitHub
+			ca := p.CA.Name
+			caDest, caData := newSourceDataIDPConfigMap(i, ca, corev1.ServiceAccountRootCAKey)
+
+			clientSecret := p.ClientSecret.Name
+			clientSecretDest, clientSecretData := newSourceDataIDPSecret(i, clientSecret, configv1.ClientSecretKey)
+
+			out = append(out,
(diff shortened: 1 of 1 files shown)


stlaz:
Note to self: get rid of `idpSyncData`, it does not seem to be bringing any value, only overhead

pkg/operator2/configsync.go
```diff
@@ -91,64 +91,189 @@ type sourceData struct {
...
+// TODO: new source data could be generalized
+func newSourceDataIDPSecret(index int, secretName, idpType string) (string, sourceData) {
+	dest := getIDPName(index, secretName, idpType)
+	var ret sourceData
+
+	volume, mount, path := secretVolume(index, dest, idpType)
+	ret.src = secretName
+	ret.path = path
+	ret.volume = volume
+	ret.mount = mount
```
ericavonb: Nit: just do 
```go
ret := sourceData{
	src: secretName,
	path: path,
	volume: volume,
	mount: mount,
}
```

pkg/operator2/configsync.go
```diff
@@ -91,64 +91,189 @@ type sourceData struct {
...
+
+func newSourceDataIDPConfigMap(index int, cmName, idpType string) (string, sourceData) {
+	dest := getIDPName(index, cmName, idpType)
+	var ret sourceData
+
+	volume, mount, path := secretVolume(index, dest, idpType)
+	ret.src = cmName
+	ret.path = path
+	ret.volume = volume
+	ret.mount = mount
```
ericavonb: Same nit about removing var and mutations in favor of setting in new struct directly

pkg/operator2/configsync.go
```diff
@@ -91,64 +91,189 @@ type sourceData struct {
...
+
+			out = append(out,
+				idpSyncData{
+					configMaps: map[string]sourceData{clientCADest: clientCAData},
+				},
+			)
+
 		default:
-			panic("TODO")
+			return nil // TODO: some erroring
```
ericavonb: `return fmt.Errorf(“unknown IdentityProvider type: %s”, pc.Type)`

pkg/operator2/configsync.go
```diff
@@ -91,64 +91,189 @@ type sourceData struct {
...
+			p := pc.BasicAuth
+
+			ca := p.CA.Name
+			caDest, caData := newSourceDataIDPConfigMap(i, ca, corev1.ServiceAccountRootCAKey)
+
+			clientCert := p.TLSClientCert.Name
+			clientCertDest, clientCertData := newSourceDataIDPSecret(i, clientCert, corev1.TLSCertKey)
+
+			clientKey := p.TLSClientKey.Name
+			clientKeyDest, clienKeyData := newSourceDataIDPSecret(i, clientKey, corev1.TLSCertKey)
```
ericavonb: Should be `corev1.TLSPrivateKeyKey`

pkg/operator2/configsync.go
```diff
@@ -91,64 +91,189 @@ type sourceData struct {
...
 // TODO this should be combined with convertProviderConfigToOsinBytes as it would simplify how the data is shared
 func convertToData(idps []configv1.IdentityProvider) []idpSyncData {
 	out := make([]idpSyncData, 0, len(idps))
 	for i, idp := range idps {
 		pc := idp.ProviderConfig
 		switch pc.Type {
+		case configv1.IdentityProviderTypeBasicAuth:
+			p := pc.BasicAuth
+
+			ca := p.CA.Name
```
ericavonb: nit: unnecessary var declaration. Use `p.CA.Name` directly. Same for `clientCert` and `clientKey`

pkg/operator2/configsync.go
ericavonb: And thank you, no panics!

pkg/operator2/configsync.go
stlaz: Good catch, I was afraid something might have gotten lost in those copy-pastes.

pkg/operator2/configsync.go
stlaz: The function currently only returns `[]idpSyncData`, I'd maybe fix this when gets merged with `convertProviderConfigToOsinBytes()`

stlaz:
Fixed comments, rebased on master.

I also added a refactoring to `convertToData()` which now only returns a pointer to the `idpSyncData` structure. Lets' hope nothing breaks!

pkg/operator2/configsync.go
ericavonb: icky, ok. for another pr

pkg/operator2/configsync.go
```diff
@@ -91,67 +90,133 @@ type sourceData struct {
...
+			p := pc.BasicAuth
+
+			caDest, caData := newSourceDataIDPConfigMap(i, p.CA.Name, corev1.ServiceAccountRootCAKey)
+			configMaps[caDest] = caData
+
+			clientCertDest, clientCertData := newSourceDataIDPSecret(i, p.TLSClientCert.Name, corev1.TLSCertKey)
+			secrets[clientCertDest] = clientCertData
+
+			clientKeyDest, clienKeyData := newSourceDataIDPSecret(i, p.TLSClientKey.Name, corev1.TLSPrivateKeyKey)
+			secrets[clientKeyDest] = clienKeyData
```
ericavonb: typo - `s/clienKeyData/clientKeyData/` (missing the 't' in 'client')

stlaz:
typo fixed

stlaz:
latest push fixes writing to `nil` maps

stlaz:
I chipped away the refactoring so that we can merge this earlier. See https://github.com/openshift/cluster-osin-operator/pull/26 for the follow-up.

stlaz:
Fixed bug in `newSourceDataIDPConfigMap()`, it had `secretVolume()` instead of the proper volume func.

enj:
Tested with 4.0 cluster, htpasswd still works so this hopefully fine.

/retest

enj:
`kube:admin` still works.  Time for the :hammer:

PR #25, opened 2019-01-16
Ensure default custom resources exist for top level config

enj:

@openshift/sig-auth this takes the simplest approach to getting these default custom resources created.

### DIFF
--- a/pkg/operator2/starter.go
+++ b/pkg/operator2/starter.go
@@ -6,11 +6,13 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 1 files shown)


pkg/operator2/starter.go
```diff
@@ -27,16 +29,42 @@ import (
...
 spec:
   managementState: Paused
 `
+
+	// TODO figure out the permanent home for top level CRDs and default CRs
+	defaultAuthentication = `
+apiVersion: config.openshift.io/v1
+kind: Authentication
+metadata:
+  name: ` + globalConfigName + `
```
enj: @stlaz I am doubting the usefulness of this.

stlaz: It prevents unnecessary errors if you're changing the name of that resource globally (should it ever change). Feel free to remove if you think that that should never be a problem.

enj: Changing it would break every operator (along with all other consumers).  I will leave it alone since it does let me trace usages (maybe that is useful).

PR #27, opened 2019-01-16
Add service account for deployment

enj:

@openshift/sig-auth


enj:
I included this in #31

PR #28, opened 2019-01-17
Merge top-level IdP config traversal functions, check for improper IdP config

stlaz:
This PR does a bunch refactorings:
* removes unnecessary creations of `idpDataSync` structures, uses only one such struct instead
* simplifies the way `sourceData` objects are stored in `idpDataSync` structure so that there's but a small amount of boilerplate
* merges the `configsync:convertToData()` and `idp:convertProviderCOnfigToOsinBytes()` functions to only traverse the IdP config once

It also adds checking for improperly created IdP configs so that we don't panic.

### DIFF
--- a/pkg/operator2/configsync.go
+++ b/pkg/operator2/configsync.go
@@ -123,91 +123,82 @@ func newSourceDataIDPConfigMap(index int, cmName, idpType string) (string, sourc
 	return dest, ret
 }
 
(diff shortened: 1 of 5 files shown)


stlaz:
/hold
--^

stlaz:
Rebased on current master.

enj:
@stlaz PTAL at #54 which includes this and other fixes.

PR #31, opened 2019-01-18
Add log level support

enj:
Add service account for deployment

---

Add log level support

This allows configuring the log level of the integrated OAuth server
via the operator config.

We now track the operator config's resource version in the
deployment hash.  This along with including the hash in the
deployment's metadata guarantees that we rollout correctly when the
operator's config changes.

We still do not rollout in all cases due to the resource syncer.
Once we correctly track those config maps and secrets, we should be
good.

We may end up rolling out "too often," but we can target the hash
more closely in the future if that happens.

---

@openshift/sig-auth


enj:
#34

PR #33, opened 2019-01-21
Try to make rename work

enj:


hack/lib/constants.sh
```diff
@@ -25,7 +25,7 @@ readonly OS_GOFLAGS_TAGS="include_gcs include_oss containers_image_openpgp"
 readonly OS_IMAGE_COMPILE_BINARIES=( )
 
 readonly OS_CROSS_COMPILE_TARGETS=(
-  cmd/osin-operator
+  cmd/authentication-operator
```
stlaz: Need to `mv cmd/{osin,authentication}-operator` too

stlaz:
Pushed a patch to fix the CI build error.

stlaz:
Getting the good old `RetryWatcher` and timeout errors now, feels like progress...
/retest

stlaz:
Needs the suggested changes, otherwise LGTM. Keeping `origin-` in the earlier commits was puzzling to me, but this was luckily fixed in later commits.

It seems that CI still feels sad :disappointed: 
/retest

PR #34, opened 2019-01-21
Consolidate open changes

enj:
All of these changes are required for beta2.

PR #36, opened 2019-01-22
Add status updates

stlaz:
Adds a `library-go` controller for status changes and also
adds status failure reporting in auth operator Sync() if an error
occured in the lower level machinery.

Based on current `library-go` and `kube-apiserver-operator` code.

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -16,8 +16,6 @@ import (
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	configinformer "github.com/openshift/client-go/config/informers/externalversions"
-	authopclient "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
-	authopinformer "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 	routeclient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	routeinformer "github.com/openshift/client-go/route/informers/externalversions/route/v1"
 	"github.com/openshift/cluster-authentication-operator/pkg/boilerplate/controller"
@@ -28,8 +26,9 @@ import (
 )
 
 const (
-	targetName       = "openshift-authentication"
-	globalConfigName = "cluster"
+	targetName         = "openshift-authentication"
+	targetNameOperator = "openshift-authentication-operator"
+	globalConfigName   = "cluster"
 
 	machineConfigNamespace = "openshift-config-managed"
 	userConfigNamespace    = "openshift-config"
@@ -75,7 +74,7 @@ const (
 )
 
 type authOperator struct {
-	authOperatorConfig authopclient.AuthenticationInterface
+	authOperatorConfigClient OperatorClient
 
 	recorder events.Recorder
 
@@ -94,8 +93,7 @@ type authOperator struct {
 }
 
 func NewAuthenticationOperator(
-	authOpConfigInformer authopinformer.AuthenticationInformer,
-	authOpConfigClient authopclient.AuthenticationsGetter,
+	authOpConfigClient OperatorClient,
 	kubeInformersNamespaced informers.SharedInformerFactory,
 	kubeClient kubernetes.Interface,
 	routeInformer routeinformer.RouteInformer,
@@ -106,7 +104,7 @@ func NewAuthenticationOperator(
 	resourceSyncer resourcesynccontroller.ResourceSyncer,
 ) operator.Runner {
 	c := &authOperator{
-		authOperatorConfig: authOpConfigClient.Authentications(),
+		authOperatorConfigClient: authOpConfigClient,
 
 		recorder: recorder,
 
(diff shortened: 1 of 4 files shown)


stlaz:
The weird order of commits is caused by Github ordering them based on their date rather than the `git` commit tree order.

stlaz:
Rebased

stlaz:
/hold cancel
#34 was merged long ago

stlaz:
For some reason I'm seeing
```
E0213 15:37:30.023897       1 operator.go:173] error updating operator status: Operation cannot be fulfilled on authentications.operator.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again
E0213 15:37:30.027876       1 operator.go:176] error updating operator status: Operation cannot be fulfilled on authentications.operator.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again
```

ericavonb:
@stlaz you should be able to use the library-go status helpers with the latest version like such:

https://github.com/openshift/cluster-kube-controller-manager-operator/blob/b3231ef5ee24f9f5b922da19728cf9ab4b1ad6ad/pkg/operator/starter.go#L116-L129

Otherwise, the client should have a `UpdateStatus` method you can use instead of updating the whole object.

stlaz:
The helpers from library-go and client's `UpdateStatus()` are already being used, the whole thing started working so we should be fine.
However, I just realized that this PR will be blocked by https://github.com/openshift/cluster-authentication-operator/pull/59 because we are creating the `ClusterOperator` resource and the operator won't move into the available state unless it's unpaused. This will cause the installer to wait indefinetely.

stlaz:
Added changes from two other PRs to see what the CI does.

stlaz:
Nvm, the bootstrap process still hangs, it was a long shot anyway. Let's have just the operator1 deleting changes.

pkg/operator2/starter.go
```diff
@@ -132,6 +137,20 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		resourceSyncer,
 	)
 
+	clusterOperatorStatus := status.NewClusterOperatorStatusController(
+		"openshift-authentication",
+		[]configv1.ObjectReference{
```
enj: This is incomplete, needs to have:

The 3 top level configs we look at
The 4 namespaces we look at
The operator config

pkg/operator2/status.go
```diff
@@ -0,0 +1,50 @@
...
+		Type:   operatorv1.OperatorStatusTypeProgressing,
+		Status: operatorv1.ConditionFalse,
+	})
+
+	v1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1.OperatorCondition{
+		Type:   operatorv1.OperatorStatusTypeFailing,
+		Status: operatorv1.ConditionFalse,
+	})
+
+	_, updateErr := c.authOperatorConfig.UpdateStatus(operatorConfig)
```
enj: Only issue updates on a change.

pkg/operator2/operator.go
```diff
@@ -158,10 +158,16 @@ func (c *authOperator) Sync(obj metav1.Object) error {
...
 	if err := c.handleSync(operatorConfig); err != nil {
+		if statusErr := c.setFailingStatus(operatorConfig, "OperatorSyncLoopError", err.Error()); statusErr != nil {
+			glog.Errorf("error updating operator status: %s", statusErr)
+		}
+
 		return err
 	}
 
-	// TODO update states and handle ClusterOperator spec/status
+	if statusErr := c.setAvailableStatus(operatorConfig); statusErr != nil {
```
enj: This is wholly insufficient - ex: if the deployment fails to rollout we would be stating everything is fine.

pkg/operator2/status.go
```diff
@@ -0,0 +1,50 @@
+package operator2
+
+import (
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+func (c *authOperator) setFailingStatus(operatorConfig *operatorv1.Authentication, reason, message string) error {
```
enj: Deep copy `operatorConfig` if you are going to mutate it.

enj:
I just did an initial pass review.  This is not ready.

pkg/operator2/status.go
enj: I believe `v1helpers.UpdateStatus` helps with this.

stlaz:
Sorry about the sorry state of the PR, I think I originally set it up so that we can build up on it, but even back then it was not very complete.

Most comments are fixed, yet the most important still stands and that's the one about when we should report as available. I'll need to see about that.

pkg/operator2/status.go
stlaz: `v1helpers.UpdateStatus()` seems to be doing that for us now.

pkg/operator2/starter.go
```diff
@@ -135,6 +139,23 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		resourceSyncer,
 	)
 
+	clusterOperatorStatus := status.NewClusterOperatorStatusController(
+		targetName,
+		[]configv1.ObjectReference{
+			{Group: operatorv1.GroupName, Resource: "authentications", Name: globalConfigName},
```
enj: Also add the infrastructure top level config per #79

PR #39, opened 2019-01-24
Changes consolidation PR #2

stlaz:
Let's just create a whole new PR since the other PRs seem to be failing differently just to see what happens. Nothing to lose at this point.


stlaz:
Merged already with #34, closing

PR #43, opened 2019-01-24
Fix loopback code exchange for /oauth/token/request

enj:
This change updates MasterCA and MasterURL to use the SSCS CA and
osin service URL.  These values are only used during loopback server
to server calls for the /oauth/token/request endpoint.

### DIFF
--- a/pkg/operator2/ca.go
+++ b/pkg/operator2/ca.go
@@ -1,0 +1,55 @@
+package operator2
+
+import (
+	"fmt"
+
+	"github.com/golang/glog"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	injectCABundleAnnotationName  = "service.alpha.openshift.io/inject-cabundle"
+	injectCABundleAnnotationValue = "true"
+)
+
+func (c *authOperator) handleCA() (*corev1.ConfigMap, error) {
+	cm := c.configMaps.ConfigMaps(targetName)
+	ca, err := cm.Get(servingCAName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		ca, err = cm.Create(defaultCA())
+	}
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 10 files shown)


pkg/operator2/ca.go
```diff
@@ -0,0 +1,55 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	injectCABundleAnnotationName  = "service.alpha.openshift.io/inject-cabundle"
+	injectCABundleAnnotationValue = "true"
+)
+
+func (c *authOperator) handleCA() (*corev1.ConfigMap, error) {
```
mrogers950: Please be specific about these functions handling the Service CA

pkg/operator2/route.go
```diff
@@ -25,7 +27,7 @@ func (c *authOperator) handleRoute() (*routev1.Route, error) {
 
 	if err := isValidRoute(route); err != nil {
 		// delete the route so that it is replaced with the proper one in next reconcile loop
-		c.route.Delete(route.Name, &metav1.DeleteOptions{})
+		glog.Infof("deleting invalid route %#v, deleteErr=%v", route, c.route.Delete(route.Name, &metav1.DeleteOptions{}))
```
mrogers950: Possibly logging deleteErr=nil seems weird, let's log that only if it happens

pkg/operator2/oauth.go
```diff
@@ -96,7 +96,7 @@ func (c *authOperator) handleOAuthConfig(operatorConfig *authv1alpha1.Authentica
 				ServingInfo: configv1.ServingInfo{
 					BindAddress: fmt.Sprintf("0.0.0.0:%d", containerPort),
 					BindNetwork: "tcp4",
-					// we have valid certs provided by alfred so that we can use reencrypt routes
+					// we have valid serving certs provided by alfred so that we can use reencrypt routes
```
mrogers950: Let's call alfred by his real name just so we don't confuse anyone not hip to the slang

enj:
/retest

The `kube:admin` test is passing so I may force merge if this takes much longer.

PR #45, opened 2019-01-24
Restore run level 9

enj:

@openshift/sig-auth I expect this to fail CI and I expect #43 to pass CI now that I have rebased it on master.  Lets see what happens.


stlaz:
The tests passing should be a good enough sign for this to get merged.
/lgtm

enj:
/hold

I do not think there is need for the bot to keep rerunning this.  Run level seems to be linked to the CI failures we are seeing.

I will manually rerun it once to show that the failure is consistent:

/retest

enj:
And #43 passed without issue!

stlaz:
Let me steal this PR for a bit to experiment with what's actually breaking CI. Chipping away deployments.

stlaz:
So it fails even with deployments delayed.

PR #48, opened 2019-01-29
handle kubeapiserver api movement

deads2k:
the kubeapiserver API is moving and the name is changing. This updates to handle both changes.  Not super gracefully, but the whole thing is temporary.

/assign @enj


enj:
/lgtm

If `kube:admin` test passes, we know this is ok.

enj:
/lgtm

Lets hope this works now :)

PR #52, opened 2019-01-30
Kick CI

enj:


enj:
I do not know if I actually need this to merge to make sure images are built...

PR #54, opened 2019-01-30
Support optional resources

enj:
Builds on #28 and fixes various bugs.

/assign @stlaz 

@stlaz review the new commits (your commits are unchanged).  I have manually tested this.


enj:
/retest

Failure is unrelated.

pkg/operator2/configsync.go
```diff
@@ -132,22 +133,30 @@ func newIDPSyncData() idpSyncData {
...
+// Returns the path for the Secret
+func (sd *idpSyncData) AddSecret(index int, secretName configv1.SecretNameReference, key string) string {
+	dest, data := newSourceDataIDPSecret(index, secretName, key)
 	sd.secrets[dest] = data
 
-	return dest
+	return data.path
+}
+
+func (sd *idpSyncData) AddSecretStringSource(index int, secretName configv1.SecretNameReference, key string) configv1.StringSource {
```
stlaz: I'd rather if we got rid of the `helpers.go` file by including the original function in `idp.go` instead of making this to be a method of this object. Having `AddSecretStringSource()` method does not make much sense because you don't have special storage for `SecretStringSource` in `idpSyncData`, there are only `Secrets` and `ConfigMaps`.

pkg/operator2/configsync.go
```diff
@@ -134,25 +134,33 @@ func newIDPSyncData() idpSyncData {
 // AddSecret initializes a sourceData object with proper data for a Secret
 // and adds it among the other secrets stored here
 // Returns the path for the Secret
-func (sd *idpSyncData) AddSecret(index int, secretName configv1.SecretNameReference, key string) string {
+func (sd *idpSyncData) AddSecret(index int, secretName configv1.SecretNameReference, key string, optional bool) string {
+	if optional && len(secretName.Name) == 0 {
```
stlaz: Validation should precede these additions, here I'd personally rather only do
```
if len(secretName.Name) == 0 {
    return ""
}
```
It's for a simple reason - you'll need to return errors, too, which is something you're not doing here. Otherwise the problem we have now when optional field is unset would now still be present for required fields with unset values.

Proposal:
Use just the condition mentioned above. That will effectively make all such fields optional. I prepared some preliminary validation code for our CRDs for origin, I need to re-check it and then post it, then I'll get to validation on the operator side.

stlaz:
I'll create a new PR with the suggested changes, those are non-breaking issues anyway.

PR #58, opened 2019-02-02
migrate operator config CRD to openshift/api/operator/v1

sallyom:
Migrate operator config CRD
see:
*  https://github.com/openshift/api/pull/193
* https://github.com/openshift/client-go/pull/84

@enj 
The above merged, this PR has 3 commits:
1) pulls updates deps to pull in changes in client-go  (hack/update-deps.sh)
2) updates config CRD to new gvr 
3) updates import paths
4) removes codegen logic

/hold


enj:
After client-go generation, do a bump(*), drop old type and generated code, drop old CRD, update old CR to new GVR.

enj:
@sallyom 

1. drop old type and generated code -> means the `pkg/apis` and `pkg/generated` folder needs to be deleted and import paths fixed - we can deal with the `Makefile` and shell script changes later
1. update old CR to new GVR -> update `defaultOperatorConfig` and its GVR (along with the `updates config CRD to new gvr - commit 1` that you have already done
1. As an aside, the `bump(*)` should be the first commit

pkg/operator2/deployment.go
```diff
@@ -8,15 +8,14 @@ import (
...
-	authv1alpha1 "github.com/openshift/cluster-authentication-operator/pkg/apis/authentication/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-const hashAnnotation = authv1alpha1.GroupName + "/rvs-hash"
+var hashAnnotation = operatorv1.GroupName + "/rvs-hash"
```
enj: @sallyom as a followup, make `operatorv1.GroupName` a `const` in `openshift/api` and then undo this change.

enj:
/lgtm

Tested with htpasswd

PR #60, opened 2019-02-05
Add CORS support for console public hostname

enj:
This change allows the console to use Javascript to interact with
the OAuth server.  This is required for JS to easily perform certain
actions such as logout.  Note that the console was always in the
allowed CORS list for all 3.x releases.

/assign @stlaz


stlaz:
Looking good!
/lgtm

PR #61, opened 2019-02-06
remove code for handling outdated kubeapiserver options

deads2k:
kube-apiserver has switched over and the CRD is moving to cluster-config-operator.

/assign @enj


enj:
/close

Not needed as we will not wedge on this resource.  I will open the CRD removal PR shortly.

PR #65, opened 2019-02-09
Add support for default key in operator boilerplate

enj:
Add support for default key in operator boilerplate

This change allows us to support flows where the key of the operator
does not exist (i.e. in case it is deleted).

It also adds generic support for defaulting the key.

---

Add defaulting for operatorv1.Authentication

This change makes it so that we render an empty
operatorv1.Authentication on start.  This is then defaulted to the
managed state at runtime.  A missing Authentication resource is
treated as the same as an empty one.

The defaults are stored once in the defaultCopyAuthenticationFunc
function (they are not duplicated in any manifests).  They are also
applied regardless of if the resource exists in the API (any unset
field means "give me the default").

### DIFF
--- a/manifests/0000_09_cluster-authentication-operator_05_deploy.yaml
+++ b/manifests/0000_09_cluster-authentication-operator_05_deploy.yaml
@@ -24,7 +24,7 @@ spec:
         command: ["authentication-operator", "operator"]
         args:
         - "--config=/var/run/configmaps/config/operator-config.yaml"
-        - "-v=2"
+        - "-v=100"
         resources:
           requests:
             memory: 50Mi
--- a/pkg/boilerplate/controller/informer.go
+++ b/pkg/boilerplate/controller/informer.go
@@ -28,12 +28,12 @@ func withSync() InformerOption {
 }
 
 func informerOptionToOption(opt InformerOption, getter InformerGetter) Option {
-	switch opt() {
+	switch o := opt(); o {
 	case syncDefault:
 		return WithInformerSynced(getter) // safe default
 	case noSync:
 		return func(*controller) {} // do nothing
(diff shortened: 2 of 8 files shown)


stlaz:
We don't necessary need to remove `defaultOperatorConfig`, I'm fine with us creating it on start if it's not present.

The only trouble I'm seeing is that now we have the constant default defined in two different places in the code, but meh.

enj:
/hold

I do not think we need it for beta2 and I want to mull on the shape of `WithDefaultKey` for a bit longer.

enj:
@stlaz take another look.  This is the direction I am thinking for #86

enj:
Tested locally.  Handles missing key as expected.

stlaz:
The code looks good, yet the logs of console suggest once again that it can't find the oauth well-known endpoint:
```
2019/03/1 12:47:25 auth: error contacting auth provider (retrying in 2s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/03/1 12:47:27 auth: error contacting auth provider (retrying in 4s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/03/1 12:47:31 auth: error contacting auth provider (retrying in 8s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/03/1 12:47:39 auth: error contacting auth provider (retrying in 16s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
```

enj:
/hold

I see the issue.  Will address next week.

PR #66, opened 2019-02-11
Redeploy when a secret/configmap from top-level config changes

stlaz:
Redeploys the osin server if any of the configMaps and secrets in the top-level config changed.

Manually tested.

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -167,30 +167,38 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 }
 
 func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) error {
+	// we get resource versions so that if either changes, we redeploy our payload
+	resourceVersions := []string{operatorConfig.GetResourceVersion()}
+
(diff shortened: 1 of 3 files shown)


pkg/operator2/operator.go
```diff
@@ -200,35 +208,36 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
...
 	cliConfig, _, err := resourceapply.ApplyConfigMap(c.configMaps, c.recorder, expectedCLIconfig)
 	if err != nil {
 		return err
 	}
+	resourceVersions = append(resourceVersions, cliConfig.GetResourceVersion())
 
 	// deployment, have RV of all resources
 	// TODO use ExpectedDeploymentGeneration func
-	// TODO manually get RV of all the config maps and secrets in syncData
 	// TODO we also need the RV for the serving-cert secret (servingCertName)
```
stlaz: Isn't the CA bundle enough?

ericavonb: The serving-cert could be rotated independently of the CA.

PR #68, opened 2019-02-12
idp: fix LDAP identity provider

stlaz:
The conversion of the LDAP IdP config type to the internal one
was missing the "Attributes" attribute.

Manually tested.


stlaz:
/approve
The CI bot apparently missed the PR creation webhook, adding approve so that this gets `/retests` even when I'm not babysitting it.

PR #69, opened 2019-02-12
Run single replica of authentication operator

derekwaynecarr:
There is no need to run 3 replicas of the operator.


derekwaynecarr:
need to still test this as we seem to also create deployment in code.

/hold

ericavonb:
It creates it's own deployment? You sure it's not the oauth server deployment you're looking at?

derekwaynecarr:
/hold cancel

yeah, you were right.  still getting my sea legs in the repo.

ericavonb:
@derekwaynecarr 👍 it's rough water right now

stlaz:
Readding
/lgtm

PR #70, opened 2019-02-12
chore(pkg/operator2/deployment): schedule on master nodes

ericavonb:
Run the oauth-server on master nodes only. Also, only run 1 replica - masters don't have enough resources for more and almost all of the other critical components are set to 1.

https://jira.coreos.com/browse/AUTH-191


ericavonb:
@stlaz or @mrogers950 could you review this?

PR #59, opened 2019-02-04
Delete operator 1

enj:
Delete operator 1

This is the minimal change to remove operator 1.  It does not attempt to update operator 2 to remove the "2" or make other cosmetic changes.

Signed-off-by: Monis Khan <mkhan@redhat.com>

---

Set operator 2 to Managed

Since operator 1 is gone, operator 2 needs to run by default to allow kube:admin to work.

Signed-off-by: Monis Khan <mkhan@redhat.com>

---

/hold

We need the router fixes for this to work.  The `kube:admin` test should fail and prevent this from merging until then.

PR #72, opened 2019-02-14
operator: change source of apiURL to infrastructure.config.openshift.io

abhinavdahiya:
Infrastructre global config [1] contains the URL to the APIServer and therefore using the `cluster-config-v1` to create the apiURL is not required.

Also installer#1169 [2]  is changing the structure of the DNS names, and console is redirecting to old apiserver URL. 

installer#1169 has feature exception approval and we need to make that in for feature feature for bea2 as upgradabililty into new DNS name is difficult.

[1]: https://github.com/openshift/api/blob/ea5d05408a95a765d44b5a4b31561b530f0b1f4c/config/v1/types_infrastructure.go#L47
[2]: https://github.com/openshift/installer/pull/1169


stlaz:
Thank you for your changes. However, all of the changed code goes away in https://github.com/openshift/cluster-authentication-operator/pull/59 so unless the CI is broken, I would refrain from any changes to it.

abhinavdahiya:
As I mentioned in the previous overview, the installer teams wants to merge https://github.com/openshift/installer/pull/1169 today before feature freeze as that change is disruptive. The CI on 1169 is green but merging that without this PR means users will not be able to access console due to the wrong apiurl being generated by current auth operator.


So it would be nice if we can go ahead with this change for now at least unblocking us as https://github.com/openshift/cluster-authentication-operator/pull/59 seems to be blocked for now
https://github.com/openshift/cluster-authentication-operator/pull/59#issuecomment-463262273

stlaz:
If we need to have this done, let's merge https://github.com/openshift/installer/pull/1242 which already has LGTM, then we can have #59 and then this PR won't be necessary.

crawford:
The [last failure](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/72/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws/263) was due to rate limits:

```
level=error msg="1 error occurred:"
level=error msg="\t* module.vpc.aws_route_table_association.route_net[3]: 1 error occurred:"
level=error msg="\t* aws_route_table_association.route_net.3: timeout while waiting for state to become 'success' (timeout: 5m0s)"
```

crawford:
[Failed](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/72/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws/265) because of a known bug:

```
[Conformance][Area:Networking][Feature:Router] The HAProxy router should enable openshift-monitoring to pull metrics [Suite:openshift/conformance/parallel/minimal]
```

/retest

enj:
/hold

You have already missed feature freeze.  #59 is the correct path to accomplish what you need.

crawford:
#59 isn't merged yet (and doesn't appear to be ready) and this PR will unblock changes that are needed for the Beta 2 release. I was told that can be safely merged before #59. Given the critical nature of the changes that are gated on this PR, I'm going to override your `/hold`. This isn't a feature; it's a bug.

/hold cancel

smarterclayton:
I agree with @crawford's override

enj:
> #59 isn't merged yet (and doesn't appear to be ready) and this PR will unblock changes that are needed for the Beta 2 release. I was told that can be safely merged before #59. Given the critical nature of the changes that are gated on this PR, I'm going to override your `/hold`. This isn't a feature; it's a bug.
> 
> /hold cancel

All of the changes are required for beta2 either way.  I would have expected @openshift/sig-auth to workaround the cert issues by disabling the CA check in the bootstrap user test.  That would nullify the need for this change and allow #59 to merge.  It is a much smaller hack that gets us closer to what we need to ship in beta2.

> I agree with @crawford's override

No worries there.  I will sort it out next week.

PR #73, opened 2019-02-15
fix(operator2/console): update console config resource name

ericavonb:
Console publishes its config under the name 'console' now.
Ref: https://github.com/openshift/console-operator/pull/134

### DIFF
--- a/pkg/operator2/console.go
+++ b/pkg/operator2/console.go
@@ -11,14 +11,9 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 )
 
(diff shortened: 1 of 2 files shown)


enj:
/hold 

Top level configs are supposed to be called `cluster`.  We should also not fail on errors here.

benjaminapetersen:
Agree with @enj, this is probably mine to fix.

benjaminapetersen:
/close

[PR 143](https://github.com/openshift/console-operator/pull/143) fixes the mistake on our end. Thanks!

PR #74, opened 2019-02-16
Pull 59 rebase

ericavonb:


ericavonb:
@sallyom the kube:admin test is failing because it tries to login before the auth server is ready. (https://bugzilla.redhat.com/show_bug.cgi?id=1678026 reports that login eventually works, making me think it's a readiness issue). Reporting the `ClusterOperator` status would prevent the install from declaring ready too early, but we can't merge #36 until after this one. Could we amend the test temporarily to wait for the auth deployment or something so that we can get this PR in and then merge #36?

enj:
Rebased #59 to preserve history / git signatures.

/close

PR #75, opened 2019-02-19
Add redeploy on serving cert and operator pod template change

stlaz:
Adds redeployment of the osin pods when the pod template or SSCS serving cert/key change

### DIFF
--- a/pkg/operator2/ca.go
+++ b/pkg/operator2/ca.go
@@ -15,18 +15,19 @@ const (
 	injectCABundleAnnotationValue = "true"
 )
 
-func (c *authOperator) handleServiceCA() (*corev1.ConfigMap, error) {
+func (c *authOperator) handleServiceCA() (*corev1.ConfigMap, *corev1.Secret, error) {
 	cm := c.configMaps.ConfigMaps(targetName)
+	secret := c.secrets.Secrets(targetName)
 	serviceCA, err := cm.Get(serviceCAName, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
 		serviceCA, err = cm.Create(defaultServiceCA())
 	}
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	if len(serviceCA.Data[serviceCAKey]) == 0 {
-		return nil, fmt.Errorf("config map has no service ca data: %#v", serviceCA)
+		return nil, nil, fmt.Errorf("config map has no service ca data: %#v", serviceCA)
 	}
 
 	if err := isValidServiceCA(serviceCA); err != nil {
@@ -36,10 +37,15 @@ func (c *authOperator) handleServiceCA() (*corev1.ConfigMap, error) {
 		if err := cm.Delete(serviceCA.Name, opts); err != nil && !errors.IsNotFound(err) {
 			glog.Infof("failed to delete invalid service CA config map: %v", err)
 		}
-		return nil, err
+		return nil, nil, err
 	}
 
-	return serviceCA, nil
+	servingCert, err := secret.Get(servingCertName, metav1.GetOptions{})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return serviceCA, servingCert, nil
 }
 
 func isValidServiceCA(ca *corev1.ConfigMap) error {
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -180,11 +180,11 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 	}
 	resourceVersions = append(resourceVersions, route.GetResourceVersion())
 
-	serviceCA, err := c.handleServiceCA()
+	serviceCA, servingCert, err := c.handleServiceCA()
 	if err != nil {
 		return err
 	}
-	resourceVersions = append(resourceVersions, serviceCA.GetResourceVersion())
+	resourceVersions = append(resourceVersions, serviceCA.GetResourceVersion(), servingCert.GetResourceVersion())
 
 	metadata, _, err := resourceapply.ApplyConfigMap(c.configMaps, c.recorder, getMetadataConfigMap(route))
 	if err != nil {
@@ -235,9 +235,14 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 	}
 	resourceVersions = append(resourceVersions, cliConfig.GetResourceVersion())
 
+	operatorDeployment, err := c.deployments.Deployments(targetNameOperator).Get(targetNameOperator, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	resourceVersions = append(resourceVersions, operatorDeployment.GetResourceVersion())
+
 	// deployment, have RV of all resources
 	// TODO use ExpectedDeploymentGeneration func
-	// TODO we also need the RV for the serving-cert secret (servingCertName)
 	expectedDeployment := defaultDeployment(
 		operatorConfig,
 		syncData,
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -67,6 +67,8 @@ const (
 	cliConfigMount      = systemConfigPathConfigMaps + "/" + cliConfigNameAndKey
 	cliConfigPath       = cliConfigMount + "/" + cliConfigNameAndKey
 
+	oauthMetadataName = systemConfigPrefix + "metadata"
+
 	userConfigPath = "/var/config/user"
 
 	servicePort   = 443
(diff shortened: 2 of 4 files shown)


enj:
We also need to track changes to any data we read/write to openshift-config-managed.

pkg/operator2/ca.go
```diff
@@ -15,18 +15,19 @@ const (
 	injectCABundleAnnotationValue = "true"
 )
 
-func (c *authOperator) handleServiceCA() (*corev1.ConfigMap, error) {
+func (c *authOperator) handleServiceCA() (*corev1.ConfigMap, *corev1.Secret, error) {
 	cm := c.configMaps.ConfigMaps(targetName)
+	sscsSecrets := c.secrets.Secrets("openshift-service-cert-signer")
```
enj: We cannot depend on the namespace of another component.

pkg/operator2/operator.go
```diff
@@ -31,6 +33,9 @@ const (
 	targetName       = "openshift-authentication"
 	globalConfigName = "cluster"
 
+	operatorNamespace      = "openshift-authentication-operator"
```
enj: just one const, `targetNameOperator`

pkg/operator2/operator.go
```diff
@@ -176,11 +181,12 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 	}
 	resourceVersions = append(resourceVersions, route.GetResourceVersion())
 
-	serviceCA, err := c.handleServiceCA()
+	serviceCA, servingCert, err := c.handleServiceCA()
 	if err != nil {
 		return err
 	}
 	resourceVersions = append(resourceVersions, serviceCA.GetResourceVersion())
+	resourceVersions = append(resourceVersions, servingCert.GetResourceVersion())
```
enj: nit, just append in the same line

pkg/operator2/operator.go
```diff
@@ -231,9 +237,14 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 	}
 	resourceVersions = append(resourceVersions, cliConfig.GetResourceVersion())
 
+	operatorPodTemplateHash, err := c.getOperatorDeploymentPodTemplateHash()
+	if err != nil {
+		return err
+	}
+	resourceVersions = append(resourceVersions, operatorPodTemplateHash)
```
enj: technically not a RV now :)

pkg/operator2/operator.go
```diff
@@ -280,3 +291,14 @@ func getPrefixFilter() controller.Filter {
 		DeleteFunc: prefix,
 	}
 }
+
+func (c *authOperator) getOperatorDeploymentPodTemplateHash() (string, error) {
```
enj: I think this is saying, if the deployment of the operator changes at all, redeploy the OAuth server.  Can we not just use the RV of the deployment?  And I believe we could have that fed into us via an env var.

pkg/operator2/ca.go
```diff
@@ -36,10 +37,15 @@ func (c *authOperator) handleServiceCA() (*corev1.ConfigMap, error) {
 		if err := cm.Delete(serviceCA.Name, opts); err != nil && !errors.IsNotFound(err) {
 			glog.Infof("failed to delete invalid service CA config map: %v", err)
 		}
-		return nil, err
+		return nil, nil, err
 	}
 
-	return serviceCA, nil
+	servingCert, err := sscsSecrets.Get("service-serving-cert-signer-serving-cert", metav1.GetOptions{})
```
enj: Same name issue.

pkg/operator2/operator.go
enj: > And I believe we could have that fed into us via an env var.

hmm I do not think that would work because it comes from the pod and not the deployment

stlaz: Yep, I learn about `EnvVarSource` this way but we probably still won't be able to use it :slightly_frowning_face:

enj:
> We also need to track changes to any data we read/write to openshift-config-managed.

@stlaz you still need to address ^

/lgtm
/retest

/hold

If you decide to address above in a different PR, feel free to unhold.

stlaz:
@enj I missed to answer that one, since CI probably won't cooperate today, I can address it here. However, when I was going through the code originally, it seemed that the only thing we're pushing to `openshift-config-managed` is the metadata configmap which is already being tracked.

Did I miss any resource or would you like these resources, should we add more, be tracked in one place somehow?

stlaz:
Based on the discussion yesterday, the problem is a bit different than I thought, so let's merge this since it's been reviewed already, I'll create another PR for the other issue.

enj:
/hold

The console operator seems to be wedging consistently on this.  We may be deadlocking each other somehow.

spadgett:
Looks like console is unable to hit the well-known endpoint with these changes

```
2019/02/24 14:00:08 cmd/main: cookies are secure!
2019/02/24 14:00:08 auth: error contacting auth provider (retrying in 2s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/02/24 14:00:10 auth: error contacting auth provider (retrying in 4s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/02/24 14:00:14 auth: error contacting auth provider (retrying in 8s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/02/24 14:00:22 auth: error contacting auth provider (retrying in 16s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/02/24 14:00:38 auth: error contacting auth provider (retrying in 32s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/02/24 14:01:10 auth: error contacting auth provider (retrying in 1m4s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
2019/02/24 14:02:14 auth: error contacting auth provider (retrying in 2m8s): discovery through endpoint https://172.30.0.1:443/.well-known/oauth-authorization-server failed: 404 Not Found
```

enj:
I think we need to move the `getMetadataConfigMap(route))` up in the sync loop so metadata does not wait for serving certs.

enj:
No failure in #82 would indicate the problem is localized to this PR.  Let us move the metadata logic up and add a comment along the lines of "make sure kube API server always sees our metadata as soon as the route has a host."  This should become less of an issue once we have a CO because the CVO will wait for us to be available, meaning the console will not be started until after the metadata is already in place (granted we have no way to guarantee that the kube API server has observed our config).

stlaz:
I added the metadata handling here as well and moved it right behind the route. Let's hope it helps, this feels a lot like a race-condition, on the other hand it's weird that #82 would just pass.

stlaz:
I moved the actual metadata publishing before the service CA is handled, too (thanks @enj, not adding that was silly -.-)

Also changed the metadata sync logic to create its CM in the target NS instead and then use the resource syncer to sync it to `openshift-config-managed`. Did not manually test it yet.

pkg/operator2/configsync.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 )
 
+type SyncResourceType string
```
enj: Follow the pattern in #83 and set up a single static sync - there is no flexibility on the name required.

stlaz: Done, the new spot is much better indeed.

pkg/operator2/starter.go
```diff
@@ -127,6 +132,14 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
+	// add syncing for the OAuth metadata ConfigMap
+	if err := resourceSyncer.SyncSecret(
```
enj: Wrong function.

pkg/operator2/operator.go
```diff
@@ -284,3 +292,13 @@ func getPrefixFilter() controller.Filter {
 		DeleteFunc: prefix,
 	}
 }
+
+func (c *authOperator) getOperatorDeploymentResourceVersion() (string, error) {
```
enj: Nit: just return the deployment so you match the other functions

stlaz:
Fixed latest comments

pkg/operator2/operator.go
```diff
@@ -284,3 +293,13 @@ func getPrefixFilter() controller.Filter {
 		DeleteFunc: prefix,
 	}
 }
+
+func (c *authOperator) getOperatorDeployment() (*appsv1.Deployment, error) {
+	deployments := c.deployments.Deployments(targetNameOperator)
+	operator, err := deployments.Get(targetNameOperator, metav1.GetOptions{})
```
enj: return here

stlaz: Made it a one-line function, maybe still worth as a shortcut, although we're only using it in this one spot.

stlaz: Inlined

enj:
/lgtm
/hold cancel

This has passed CI once already so we should be good.

PR #76, opened 2019-02-19
Reduce log spam and turn off unneeded delegation

enj:
This change lowers the default log level of the operator to 2.  This prevents leader election spam (we still want leader election even with one replica to make sure we behave nicely during an upgrade).

It also disables delegated authentication and authorization as the operator does not expose any HTTP endpoints.

@openshift/sig-auth


ericavonb:
solid changes 👍 
/lgtm

PR #77, opened 2019-02-20
Improve status handling

stlaz:
This will block the installation until we report availability

### DIFF
--- a/manifests/cluster-authentication-operator_03_version-mapping.yaml
+++ b/manifests/cluster-authentication-operator_03_version-mapping.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-authentication-operator
+  name: version-mapping
+data:
+  "0.0.1-snapshot_openshift": "quay.io/openshift/origin-hypershift:v4.0"
+  "0.0.1-snapshot": "quay.io/openshift/origin-cluster-authentication-operator:v4.0"
--- a/manifests/cluster-authentication-operator_06_clusteroperator.yaml
+++ b/manifests/cluster-authentication-operator_06_clusteroperator.yaml
@@ -6,3 +6,5 @@ status:
   versions:
   - name: operator
     version: "0.0.1-snapshot"
+  - name: integrated-oauth-server
+    version: "0.0.1-snapshot_openshift"
--- a/pkg/operator2/configmap.go
+++ b/pkg/operator2/configmap.go
@@ -51,6 +51,8 @@ func getMetadataStruct(route *routev1.Route) map[string]interface{} {
 	return ret
 }
 
+// TODO: the code in this file does not reflect situations where the
+// OAuthMetadata field of the Authentication object is set
 func getMetadata(route *routev1.Route) string {
 	host := route.Spec.Host
 	return strings.TrimSpace(fmt.Sprintf(stubMetadata, host, host, host))
--- a/pkg/operator2/configmap.go
+++ b/pkg/operator2/configmap.go
@@ -1,6 +1,7 @@
 package operator2
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -36,6 +37,19 @@ const stubMetadata = `
   ]
 }
 `
+
+func getMetadataStruct(route *routev1.Route) map[string]interface{} {
+	var ret map[string]interface{}
+
+	metadataJSON := getMetadata(route)
+	err := json.Unmarshal([]byte(metadataJSON), &ret)
+	if err != nil {
+		// should never happen unless the static metadata is broken
+		panic(err)
+	}
+
+	return ret
+}
 
 func getMetadata(route *routev1.Route) string {
 	host := route.Spec.Host
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -12,7 +12,6 @@ import (
 	"github.com/golang/glog"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
@@ -37,6 +36,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/library-go/pkg/operator/status"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
 var deploymentVersionHashKey = operatorv1.GroupName + "/rvs-hash"
@@ -46,6 +46,7 @@ const (
 	targetName          = "openshift-authentication"
 	targetNameOperator  = "openshift-authentication-operator"
 	globalConfigName    = "cluster"
+	osinOperandName     = "integrated-oauth-server"
 
 	operatorVersionEnvName = "OPERATOR_IMAGE_VERSION"
 
@@ -212,15 +213,13 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 
 	syncErr := c.handleSync(operatorConfigCopy)
 	if syncErr != nil {
-		c.setFailingStatus(operatorConfig, "OperatorSyncLoopError", syncErr.Error())
+		c.setFailingStatus(operatorConfigCopy, "OperatorSyncLoopError", syncErr.Error())
 	}
 
-	// Update status if it changed
-	if !equality.Semantic.DeepEqual(operatorConfig, operatorConfigCopy) {
-		if _, err := c.authOperatorConfigClient.Client.Authentications().UpdateStatus(operatorConfigCopy); err != nil {
-			return err
-		}
-	}
+	v1helpers.UpdateStatus(c.authOperatorConfigClient, func(status *operatorv1.OperatorStatus) error {
+		operatorConfigCopy.Status.OperatorStatus.DeepCopyInto(status)
+		return nil
+	})
 
 	return syncErr
 }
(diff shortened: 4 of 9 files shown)


manifests/cluster-authentication-operator_06_clusteroperator.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: openshift-authentication
+spec: {}
```
enj: omit all together ?

stlaz: I shamelessly copied this from kube-apiserver-operator, let's see what happens when we remove it.

nee1esh:
Looks like there are some syntax/build errors in pkg/operator2/operatorclient.go

stlaz:
CI is probably unable to apply the changes due to the required rebase (the unexpected == and << come from merge conflicts). We need to get status reporting right first anyway before merging this PR.

enj:
The CO code we have is buggy and incomplete because #36 was merged before it was ready.  This needs to wait until those issues are resolved.

stlaz:
I pushed just to create backup, the order of commits is not ok and the operator will be in a beautiful update cycle as we're updating the operator's config by the status, which in turn tries to update the osin deployment, which leads to updating the status.

I'll try to sort the cyclic issue by using generations where possible tomorrow.

stlaz:
The cycles are resolved, it might be better to join generation + resource kind to make sure that the value is more unique, but it seems that the installer properly waits for us to be ready \o/

stlaz:
@openshift/sig-auth Probably good/suitable for review. I've focused mostly on reporting availability when I believe we're sure to be serving authn requests properly (deployment is ready, the route gives 200s and the well-known endpoint advertises expected values).

The Failing state is still quite underdeveloped.

Also, see if you're fine with how redeployments are done now in relation to objects that include generation. This needed to change as "resourceVersion" changed with any status object of such object, whereas "generation" does not do that.

Note that this also moves our operator back to the default namespace as required per discussion on the team list.

pkg/operator2/operator.go
```diff
@@ -311,3 +431,34 @@ func getPrefixFilter() controller.Filter {
 		DeleteFunc: prefix,
 	}
 }
+
+func newHTTPSClient() (*http.Client, error) {
```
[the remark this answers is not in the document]
enj: Actually you do need to keep rebuilding since this will talk to a route.  As we discussed on slack, the TransportFor rest config logic is probably what you want.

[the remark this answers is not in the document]
stlaz: I keep the Transport and rebuild the client, not exactly sure whether I need to rebuild the Transport too. It seems to do the trick as is, though.

stlaz:
Fixed the comments. I thought it might be good to see what happens during the progressing status so I added setting reason to it in the waiting stages.

pkg/operator2/operator.go
```diff
@@ -283,16 +317,186 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
...
-	deployment, _, err := resourceapply.ApplyDeployment(c.deployments, c.recorder, expectedDeployment, c.getGeneration(), false)
+	deployment, changed, err := resourceapply.ApplyDeployment(c.deployments, c.recorder, expectedDeployment, c.getGeneration(), false)
 	if err != nil {
 		return err
 	}
 
 	glog.V(4).Infof("current deployment: %#v", deployment)
 
+	if changed {
+		if err = c.setProgressingStatus(operatorConfig, "DeploymentChanged", ""); err != nil {
```
enj: We should hit this branch whenever `changed || replicas != expected`

pkg/operator2/status.go
```diff
@@ -33,6 +33,28 @@ func (c *authOperator) setFailingStatus(operatorConfig *operatorv1.Authenticatio
 	return err
 }
 
+func (c *authOperator) setProgressingStatus(operatorConfig *operatorv1.Authentication, reason, message string) error {
```
enj: I am wondering if we call this too much - i.e. if we should build up a series of conditions and apply once?

pkg/operator2/operator.go
```diff
@@ -283,16 +317,186 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
...
+		if err = c.WaitUntilReady(operatorConfig, route, deployment.Annotations[deploymentVersionHashKey]); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
+func (c *authOperator) WaitUntilReady(operatorConfig *operatorv1.Authentication, route *routev1.Route, deploymentVersionHash string) error {
+	return wait.Poll(1, 5*time.Minute, func() (bool, error) {
```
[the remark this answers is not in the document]
enj: Max time should be around 2 mins.

manifests/cluster-authentication-operator_06_clusteroperator.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: authentication
```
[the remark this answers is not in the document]
enj: Ex version: https://github.com/openshift/cluster-image-registry-operator/pull/225

pkg/operator2/status.go
enj: Actually it should not matter because the helper functions only update when needed.

pkg/operator2/operator.go
```diff
@@ -206,7 +239,7 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 	if err != nil {
 		return err
 	}
-	resourceVersions = append(resourceVersions, authConfig.GetResourceVersion())
+	generationSum += authConfig.Generation
```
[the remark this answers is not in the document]
stlaz: The thing is that a hash of the same number of a different resource will always be the same so it does not matter whether you sum it or keep them separate.

We could however do separate `generation + kind` strings to distinguish this properly.

manifests/cluster-authentication-operator_06_clusteroperator.yaml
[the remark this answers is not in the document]
stlaz: Currently I'm focusing only on getting the conditions right.

We haven't worked with generations very much up until now, and it's still far away from being optimal.

I'll see about getting version in here.

pkg/operator2/operator.go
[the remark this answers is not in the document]
stlaz: Ah, never mind, since we do `rvs := strings.Join(resourceVersions, ",")` it's already order-dependent. It would have been anyway.

stlaz:
Addressed comments, should still add version.

stlaz:
Did some substantial changes to the operator status updating logic. We now also track the generation of our deployment.

Still need to add version.

stlaz:
Added version, turned out to be very easy.

stlaz:
Based on [e2e-aws/results/clusteroperators.json/](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/77/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws/416/artifacts/e2e-aws/clusteroperators.json), the version seems to work.

stlaz:
Hm, I see that both the operator and its pods ran into cert errors. Weird.

manifests/cluster-authentication-operator_03_version-mapping.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-authentication-operator
+  name: version-mapping
```
enj: Do we need this?  I think it is only used in the `VersionForOperand` code.

pkg/operator2/operator.go
```diff
@@ -379,10 +378,16 @@ func (c *authOperator) CheckReady(
 		return deploymentReady, err
 	}
 	if !deploymentReady {
-		c.setProgressingStatus(operatorConfig, "DeployementNotReady", deploymentMsg)
+		c.setProgressingStatus(operatorConfig, "DeploymentNotReady", deploymentMsg)
```
enj: Might want to be specific on which deployment.

pkg/operator2/configmap.go
```diff
@@ -37,6 +38,20 @@ const stubMetadata = `
 }
 `
 
+func getMetadataStruct(route *routev1.Route) map[string]interface{} {
+	var ret map[string]interface{}
+
+	metadataJSON := getMetadata(route)
+	err := json.Unmarshal([]byte(metadataJSON), &ret)
+	if err != nil {
+		panic(err)
```
enj: comment stating this panic should never happen unless our static metadata is broken

pkg/operator2/operator.go
```diff
@@ -180,34 +212,30 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 		return nil // TODO do something better for all states
 	}
 
-	operatorConfigCopy := operatorConfig.DeepCopy()
-
-	syncErr := c.handleSync(operatorConfigCopy)
+	syncErr := c.handleSync(operatorConfig)
```
enj: Keep the copy.  We want to always know that this is safe to mutate.

pkg/operator2/operator.go
```diff
@@ -180,34 +212,30 @@ func (c *authOperator) Sync(obj metav1.Object) error {
...
+		c.setAvailableStatus(operatorConfig)
 	}
 
-	// Update status if it changed
-	if !equality.Semantic.DeepEqual(operatorConfig, operatorConfigCopy) {
-		if _, err := c.authOperatorConfigClient.Client.Authentications().UpdateStatus(operatorConfigCopy); err != nil {
-			return err
-		}
-	}
+	v1helpers.UpdateStatus(c.authOperatorConfigClient, func(status *operatorv1.OperatorStatus) error {
```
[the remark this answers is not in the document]
stlaz: There was a bug caused by rebase, I forgot I should be setting the available status somewhere else. This is for all the cases.

manifests/cluster-authentication-operator_03_version-mapping.yaml
stlaz: I couldn't find `version-mapping` code, was looking at bad places (CVO repo) :man_facepalming: I'll use this instead of the `IMAGE_VERSION` env var in the payload deployment.

pkg/operator2/operator.go
```diff
@@ -180,34 +212,30 @@ func (c *authOperator) Sync(obj metav1.Object) error {
...
+		status = operatorConfig.Status.OperatorStatus.DeepCopy()
```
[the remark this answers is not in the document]
stlaz: Right, this is but a beautiful NOP

stlaz:
Addressed all comments, submitting to e2e to check that the new version for operand through version-mapping works.

pkg/operator2/operator.go
```diff
@@ -9,10 +9,11 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
```
enj: move down

PR #78, opened 2019-02-21
Use consoleURL field for console top level config

enj:
bump(*)

---

no-op fixes required to compile

---

Use protobuf for non custom resources

---

Use consoleURL field for console top level config

The console has moved to a URL field in its status instead of a hostname as it is easier for other components to consume.

---

/assign @stlaz


enj:
cc @benjaminapetersen @spadgett

stlaz:
I see this is related to https://github.com/openshift/console-operator/pull/151.
/lgtm

PR #79, opened 2019-02-21
Move to OsinServerConfig

enj:
This change moves to the OsinServerConfig struct.  This removes our dependence on the KubeAPIServerConfig.

It also wires the apiServerURL from the top level infrastructure config.  This allows us to provide the correct login url for the token request endpoint.

@openshift/sig-auth 

/hold

Blocked on https://github.com/openshift/origin/pull/21987

### DIFF
--- a/pkg/operator2/infrastructure.go
+++ b/pkg/operator2/infrastructure.go
@@ -1,0 +1,19 @@
+package operator2
+
+import (
+	"github.com/golang/glog"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func (c *authOperator) handleInfrastructureConfig() *configv1.Infrastructure {
+	infrastructureConfig, err := c.infrastructure.Get(globalConfigName, metav1.GetOptions{})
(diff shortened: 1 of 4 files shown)


pkg/operator2/infrastructure.go
```diff
@@ -0,0 +1,19 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func (c *authOperator) handleInfrastructureConfig() *configv1.Infrastructure {
+	infrastructureConfig, err := c.infrastructure.Get(globalConfigName, metav1.GetOptions{})
+	if err != nil {
+		glog.Infof("error getting infrastructure config: %v", err)
+		// have a placeholder that will at least look reasonable in the token request endpoint
+		return &configv1.Infrastructure{Status: configv1.InfrastructureStatus{APIServerURL: "<api_server_url>"}}
```
stlaz: Won't the invalid URL just break osin later on?

enj: Osin just prints the value; it does not parse it.

stlaz: Ok, I guess that's fine, thanks.

PR #80, opened 2019-02-21
Bug 1677587: Add openAPIV3Schema validation to operator config

sallyom:
@enj  I used `library-go/cmd/crd-schema-gen` to generate the CRD validation

operator api has been updated recently, so had to bump deps to get the fancy crd-validation-gen.  however, that bump caused issues in auth-operator that I did not want to fix in this PR, so I  deleted the bump commits, will address those issues here: https://github.com/openshift/cluster-authentication-operator/pull/97

### DIFF
--- a/manifests/cluster-authentication-operator_02_config.crd.yaml
+++ b/manifests/cluster-authentication-operator_02_config.crd.yaml
@@ -12,3 +12,149 @@ spec:
     singular: authentication
   subresources:
     status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: +required
+          properties:
+            logLevel:
+              description: logLevel is an intent based logging for an overall component.  It
+                does not give fine grained control, but it is a simple way to manage
+                coarse grained logging choices that operators have to interpret for
+                their operands.
+              type: string
+            managementState:
+              description: managementState indicates whether and how the operator
+                should manage the component
+              pattern: ^Managed|Unmanaged|Force|Removed$
+              type: string
+            observedConfig:
+              description: observedConfig holds a sparse config that controller has
+                observed from the cluster state.  It exists in spec because it is
+                an input to the level for the operator +nullable
+              type: object
+            operandSpecs:
+              description: operandSpecs provide customization for functional units
+                within the component
+              items:
+                properties:
+                  name:
+                    description: name is the name of this unit.  The operator must
+                      be aware of it.
+                    type: string
+                  operandContainerSpecs:
+                    description: operandContainerSpecs are per-container options
+                    items:
+                      properties:
+                        name:
+                          description: name is the name of the container to modify
+                          type: string
+                        resources:
+                          description: resources are the requests and limits to place
+                            in the container.  Nil means to accept the defaults.
+                          type: object
+                      type: object
+                    type: array
+                  unsupportedResourcePatches:
+                    description: 'unsupportedResourcePatches are applied to the workload
+                      resource for this unit. This is an unsupported workaround if
+                      anything needs to be modified on the workload that is not otherwise
+                      configurable. TODO Decide: alternatively, we could simply include
+                      a RawExtension which is used in place of the "normal" default
+                      manifest'
+                    items:
+                      properties:
+                        patch:
+                          description: patch the patch itself
+                          type: string
+                        type:
+                          description: 'type is the type of patch to apply: jsonmerge,
+                            strategicmerge'
+                          type: string
+                      type: object
+                    type: array
+                type: object
+              type: array
+            unsupportedConfigOverrides:
+              description: 'unsupportedConfigOverrides holds a sparse config that
+                will override any previously set options.  It only needs to be the
+                fields to override it will end up overlaying in the following order:
+                1. hardcoded defaults 2. observedConfig 3. unsupportedConfigOverrides
+                +nullable'
+              type: object
+          type: object
+        status:
+          properties:
+            conditions:
+              description: conditions is a list of conditions and their status
+              items:
+                properties:
+                  lastTransitionTime:
+                    format: date-time
+                    type: string
+                  message:
+                    type: string
+                  reason:
+                    type: string
+                  status:
+                    type: string
+                  type:
+                    type: string
+                type: object
+              type: array
+            generations:
+              description: generations are used to determine when an item needs to
+                be reconciled or has changed in a way that needs a reaction.
+              items:
+                properties:
+                  group:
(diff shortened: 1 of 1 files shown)


manifests/cluster-authentication-operator_02_config.crd.yaml
```diff
@@ -5,10 +5,49 @@ metadata:
...
+                required: ["name", "operandContainerSpecs", "unsupportedResourcePatches"]
+                properties:
+                  name:
+                    type: string
+                  operandContainerSpecs:
+                    type: object
+                    required: ["name"]
+                    properties:
+                      name:
+                        type: string
```
sallyom: @enj in operandContainerSpecs, there's also 
`Resources *corev1.ResourceRequirements `json:"resources,omitempty"``  do I need to add that?

enj:
> @enj not sure about this, i followed api/operator/v1/types.go

You seem to be implying that I know anything about `openAPIV3Schema`.  I have never used it for anything before :smile:

manifests/cluster-authentication-operator_02_config.crd.yaml
```diff
@@ -5,10 +5,68 @@ metadata:
...
+    # Ensure we will be able to deserialize the object into the golang type
+    openAPIV3Schema:
+      properties:
+        spec:
+          type: object
+          required: ["managementState"]
+          properties:
+            managementState:
+              type: string
+              enum:
```
enj: We need to tolerate "unset" (i.e. use the default) for every field.

sallyom: looking, ok

manifests/cluster-authentication-operator_02_config.crd.yaml
```diff
@@ -5,10 +5,71 @@ metadata:
...
+                - "Managed"
+                - "Unmanaged"
+                - "Removed"
+              default: "Managed"
+            logLevel:
+              enum:
+                - "Normal"
+                - "Dubug"
+                - "Trace"
+                - "TraceAll"
```
sallyom: No default here, bc it's not always set

manifests/cluster-authentication-operator_02_config.crd.yaml
```diff
@@ -5,10 +5,71 @@ metadata:
...
+            operandSpecs:
+              type: array
```
sallyom: operandSpecs not required, but _if_ it's provided validate each?

manifests/cluster-authentication-operator_02_config.crd.yaml
```diff
@@ -5,10 +5,71 @@ metadata:
...
+                      required: ["type", "patch"]
+                      properties:
+                        type:
+                          type: string
+                        patch:
+                          type: string
+            unsupportedConfigOverrides:
+              type: object
+            observedConfig:
+              type: object
```
sallyom: If we leave out unsupportedConfigOverrides and observedConfig, that means nobody can set them, right?  So, we need to list all possible properties?

manifests/cluster-authentication-operator_02_config.crd.yaml
sallyom: uh, nevermind, looks like setting `default` is not allowed: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation

manifests/cluster-authentication-operator_02_config.crd.yaml
sallyom: ^^ seems I'm wrong about that: https://github.com/kubernetes/kubernetes/issues/59151  I should proably _not_ include these, since we're not actually able to validate past 'object'

manifests/cluster-authentication-operator_02_config.crd.yaml
sallyom: it's safe to set "managementState" required, right? It will default to "Managed" based on `defaultOperatorConfig` [here](https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/starter.go#L37).

enj: > it's safe to set "managementState" required, right? It will default to "Managed" based on `defaultOperatorConfig` [here](https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/starter.go#L37).

No.  Any field can be left unset to indicate "give me the default."  Empty `spec` is valid:

```yaml
apiVersion: operator.openshift.io/v1
kind: Authentication
metadata:
  name: cluster
```

The existing code needs to be updated to handle this (i.e. it needs to do defaulting on inputs).

sallyom: ok

sallyom:
@sttts @damemi take a look, I removed the status fields, left only `type: object`, but other than that this CRD was generated using mike's PR

sallyom:
I _was_ going to add `enum` for managementState and logLevel like so:
```yaml
(managementState)
              enum:
                - 'Force'
                - 'Managed'
                - 'Unmanaged'
                - 'Removed'
(logLevel)
              enum:
                - 'Normal'
                - 'Debug'
                - 'Trace'
                - 'TraceAll'
```
but then realized just because these are only options listed in api/operator/v1/types.go, there's nothing enforcing that these are the only possible options, and in fact, auth-operator has used 'Paused' in managementState recently.

enj:
> I _was_ going to add `enum` for managementState and logLevel like so:
> 
> ```yaml
> (managementState)
>               enum:
>                 - 'Force'
>                 - 'Managed'
>                 - 'Unmanaged'
>                 - 'Removed'
> (logLevel)
>               enum:
>                 - 'Normal'
>                 - 'Debug'
>                 - 'Trace'
>                 - 'TraceAll'
> ```
> 
> but then realized just because these are only options listed in api/operator/v1/types.go, there's nothing enforcing that these are the only possible options, and in fact, auth-operator has used 'Paused' in managementState recently.

Isn't the point of validation to constrain the inputs to what is valid?

enj:
/hold

I believe we should have the strictest validation we can have while allowing fields to be optional.  I would say that includes proper enums for all relevant fields.

sallyom:
@enj I added the enums back, you can remove the hold if all looks ok

manifests/cluster-authentication-operator_02_config.crd.yaml
```diff
@@ -3,12 +3,109 @@ kind: CustomResourceDefinition
...
+              description: 'unsupportedConfigOverrides holds a sparse config that
+                will override any previously set options.  It only needs to be the
+                fields to override it will end up overlaying in the following order:
+                1. hardcoded defaults 2. observedConfig 3. unsupportedConfigOverrides'
+              type: object
+          type: object
+        status:
+          type: object
+  version: v1
+status:
```
enj: Should be unset

enj:
```
fail [github.com/openshift/origin/test/extended/operators/operators.go:161]: Mar  4 22:31:09.936: Some cluster operators never became available <nil>/openshift-authentication
```

manifests/cluster-authentication-operator_02_config.crd.yaml
sallyom: fixed, and now the only additions are the validation section.

manifests/cluster-authentication-operator_02_config.crd.yaml
```diff
@@ -12,3 +12,96 @@ spec:
     singular: authentication
   subresources:
     status: {}
```
sallyom: I'm not sure about this subresources field.  Looks like it's omitted from cluster-network-operator crd, we might want to omit that here?

sallyom: NOPE, that belongs, nevermind

PR #83, opened 2019-02-25
Move to a passthrough route to support mTLS

enj:
This change moves us to a passthrough route that can support the mTLS requirements of the request header identity provider.

1. Sync router certs from openshift-config-managed to openshift-authentication namespace
2. Mount (copied) router certs into deployment
3. SNI: wire each key, value pair in router certs secret to a domain and the mount path for the combined TLS cert and key data.

@openshift/sig-auth @Miciah @ironcladlou @knobunc

### DIFF
--- a/pkg/operator2/deployment.go
+++ b/pkg/operator2/deployment.go
@@ -28,6 +28,7 @@ func (c *authOperator) getGeneration() int64 {
 func defaultDeployment(
 	operatorConfig *operatorv1.Authentication,
 	syncData *configSyncData,
+	routerSecret *corev1.Secret,
 	resourceVersions ...string,
 ) *appsv1.Deployment {
 	replicas := int32(1) // TODO configurable?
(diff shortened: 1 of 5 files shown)


enj:
/hold

Needs to wait for #84 and https://github.com/openshift/cluster-ingress-operator/pull/139

@stlaz please review.  I have tested this locally and it certainly appears to work :smile:

stlaz:
/lgtm
We still need to wait for the ingress operator changes, though, as otherwise this would just be failing to get the secret from `openshift-config-managed`.

stlaz:
/lgtm cancel
I forgot that the bot may start retrying to retest the PR, there's no reason for that now as we know it would fail.

enj:
Rebased on master for minor git conflict.  Still waiting on https://github.com/openshift/cluster-ingress-operator/pull/139

enj:
https://github.com/openshift/cluster-ingress-operator/pull/139 is merged

/hold cancel
/retest

PR #86, opened 2019-02-28
Don't error out if top level configs do not exist

stlaz:
When either or both the OAuth and Authentication objects is not found,
don't fail and assume the defaults.

-----

The operator's config itself is handled in https://github.com/openshift/cluster-authentication-operator/pull/65

### DIFF
--- a/pkg/operator2/auth.go
+++ b/pkg/operator2/auth.go
@@ -1,6 +1,7 @@
 package operator2
 
 import (
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -8,8 +9,13 @@ import (
 
 func (c *authOperator) handleAuthConfig() (*configv1.Authentication, error) {
 	auth, err := c.authentication.Get(globalConfigName, metav1.GetOptions{})
+
 	if err != nil {
-		return nil, err
+		if !errors.IsNotFound(err) {
+			return nil, err
+		}
+		// did not find the object, use default
+		auth = defaultAuthenticationConfig()
 	}
 
 	expectedReference := configv1.ConfigMapNameReference{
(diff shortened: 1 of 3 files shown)


enj:
/hold

Let me update #65 so the approach I want to take here will become more clear.

pkg/operator2/auth.go
```diff
@@ -1,15 +1,21 @@
...
 func (c *authOperator) handleAuthConfig() (*configv1.Authentication, error) {
 	auth, err := c.authentication.Get(globalConfigName, metav1.GetOptions{})
+
 	if err != nil {
-		return nil, err
+		if !errors.IsNotFound(err) {
+			return nil, err
+		}
+		// did not find the object, use default
+		auth = defaultAuthenticationConfig()
```
ericavonb: I think you need to create the default object (before attempting to update it's status 🙂 )

pkg/operator2/oauth.go
```diff
@@ -44,7 +45,11 @@ func (c *authOperator) handleOAuthConfig(
 ) {
 	oauthConfig, err := c.oauth.Get(globalConfigName, metav1.GetOptions{})
 	if err != nil {
-		return nil, nil, nil, err
+		if !errors.IsNotFound(err) {
+			return nil, nil, nil, err
+		}
+		// did not find the object, use default
+		oauthConfig = defaultOAuthConfig()
```
ericavonb: same comment.

pkg/operator2/oauth.go
```diff
@@ -164,6 +169,19 @@ func (c *authOperator) handleOAuthConfig(
...
 }
 
+func defaultOAuthConfig() *configv1.OAuth {
+	return &configv1.OAuth{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: globalConfigName,
+		},
+		Spec: configv1.OAuthSpec{
+			TokenConfig: configv1.TokenConfig{
+				AccessTokenMaxAgeSeconds: 86400,
```
ericavonb: nit: make a const, add comment about how long this is.

enj:
#108

PR #90, opened 2019-03-04
Update operator and operand manifests for priority and imagePolicy

derekwaynecarr:
Operators deployed via the CVO should have system-cluster-critical priority.
Operators that use shas for images should use IfNotPresent pull policy.


derekwaynecarr:
smoke testing that will verify.

related: https://github.com/openshift/origin/pull/22217

PR #92, opened 2019-03-08
feat(manifests): configure cluster-montoring on auth resources

ericavonb:
Bug 1684407 - Missing auth Prometheus endpoint coverage


pkg/operator2/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+	existing, err := c.servicemonitors(svc.GetNamespace()).Get(svc.Name())
+	if err == nil || !apierrors.IsNotFound(err) {
+		// Existing instance found, or unknown error
+		return existing, err
+	}
+
+	// No existing instance found; attempt to create default
+	glog.V(5).Infof("creating ServiceMonitor '%s/%s'", svc.GetNamespace(), svc, GetName())
+	created, err := c.servicemonitors(svc.GetNamespace()).Create(serviceMonitorForService(svc))
+	if err == nil || !apierrors.IsAlreadyExists(err) {
```
stlaz: Why refetch it? We failed creating it for whatever reason, try in next sync loop iteration?

stlaz: Also, should we maybe stomp on any changes to this monitor?

manifests/metrics_role.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
stlaz: All other manifests filenames take the form `cluster-authentication-operator_{NUM}_{RESOURCE_DESC}.yaml`, we should keep that form in order to be able to identify our manifest files in the CVO payload I believe.

ericavonb: I figured I'd need to adjust these some and like figure out what number they should be at. The names are temporary while I get the rest sorted out. 
But also, we need to rename all the manifests in here to get rid of the `cluster-authentication-operator_` prefix. They show up as `0000_70_cluster-authentication-operator_cluster-authentication-operator_0...` in the release 😳

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -0,0 +1,49 @@
...
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    metricRelabelings:
+    - action: drop
+      regex: etcd_(debugging|disk|request|server).*
+      sourceLabels:
+      - __name__
+    port: https
```
stlaz: This should probably have the `/metrics` endpoint path specified?

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -0,0 +1,49 @@
+# Configure cluster-monitoring for cluster-authentication-operator
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: openshift-authentication-operator
+  namespace: openshift-authenticaion-operator
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
stlaz: Does the operator need to do anything to expose metrics somehow?

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: openshift-authentication-operator
-  namespace: openshift-authenticaion-operator
+  namespace: openshift-authentication-operator
```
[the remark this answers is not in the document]
ericavonb: done

manifests/06_deployment.yaml
```diff
@@ -49,6 +51,10 @@ spec:
         configMap:
           defaultMode: 440
           name: openshift-authentication-operator-config
+      - name: serving-cert
+        secret:
+          secretName: openshift-authentication-operator-serving-cert
+          optional: true
```
enj: Does this get reloaded if we start before we have it?  We probably want to wait for it otherwise.

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
ericavonb: I think `/metrics` is the default, but doesn't hurt to be explicit 👍

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
ericavonb: use the `WithServer` option from library-go: https://github.com/openshift/library-go/blob/dab26bb3a8dc7fccde7227194af755bbff30ce5d/pkg/controller/controllercmd/builder.go#L128

manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
```diff
@@ -0,0 +1,65 @@
+# Role for accessing metrics exposed by the cluster-authentication-operator
```
enj: We should consider making this into a cluster role if it is getting duplicated across all operators (jira for 4.2 seems reasonable).

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - action: drop
+      regex: etcd_(debugging|disk|request|server).*
+      sourceLabels:
+      - __name__
+    path: /metrics
+    port: https
+    scheme: https
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      serverName: metrics.openshift-authentication-operator.svc
```
enj: Wrong name?

manifests/02_service.yaml
```diff
@@ -0,0 +1,20 @@
+# Expose operator metrics
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: openshift-authentication-operator-serving-cert
+  labels:
+    app: openshift-authentication-operator
+  name: metrics
```
enj: Prefer `openshift-authentication-operator` for consistency.

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -0,0 +1,51 @@
...
```
enj: See other comment.

manifests/02_service.yaml
ericavonb: I used `metrics` to be consistent with the other operators (`openshift-apiserver-operator`, `openshift-kube-scheduler-operator`, `openshift-controller-manager-operator`, etc). It's also preferred only namespaces have the `openshift-` prefix with names kept simple; I figure we can update the other stuff in a separate pr.

manifests/06_deployment.yaml
ericavonb: I think the server dynamically reloads it (for library-go metrics servers)

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
enj: /hold

This one you have to address.

manifests/02_service.yaml
enj: OK.  I suggest changing the name of all resources in the OAuth server namespace from `openshift-authentication` to something like `oauth` or `integrated-oauth-server`.  Calling the service of the OAuth server `metrics` is incorrect, but you could create a new service with that name if you so desired (and maybe skip renaming the rest all together).

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
ericavonb: 9bd2d87bf4f0323f8b943e5dd3c657a1R47

PR #93, opened 2019-03-11
Temp increase log level

enj:

/hold

Just running CI to see how long it takes us to hit steady state.


enj:
```
Second		Event

+  0		Start
+  9		OAuth metadata ready
+ 82		OAuth deployment ready
+103		well-known observed by console
+ 81		Console can reach route without x509 errors
+  6		Console is available
+  4		Console is no longer progressing
```

Gathered the above metrics.

PR #96, opened 2019-03-20
use route's status to get info about host

stlaz:


ericavonb:
@stlaz what's the status on this PR? are you waiting on review from someone?

stlaz:
@ericavonb this was originally started as simple patch but the problematic is a bit deeper so I moved back to things with higher priority.
/hold not ready

enj:
#122

PR #97, opened 2019-03-20
Bug 1694819: bump deps and add nullable fields to crd validation

sallyom:
* bump to kubernetes-1.13.4 
* vendor openshift/kubernetes-apiextensions-apiserver to bring in PR that adds ability for nullable fields to CRDs, until that change merges upstream
* update authentication crd with proper validation
* bump deps, but keep patch on vendor until openshift/api types_oauth.go change is complete, so beta3-beta3 upgrades don't break. 
* fix build with bump-update clusterOperatorStatusController
* switch glog to klog for kubernetes-1.13.4


sallyom:
@enj only have to pin library-go, and it's passing now : )  should be able to release the hold, AFAICT this release from try.openshift.com has the updates we were waiting for: https://github.com/openshift/installer/releases/tag/v0.16.1 , payload version 4.0.0-0.9

pkg/crd-schema-gen.go
```diff
@@ -0,0 +1,3 @@
+package pkg
+
+import _ "github.com/openshift/library-go/cmd/crd-schema-gen/generator"
```
stlaz: I don't understand this change, you're running the command fro inside the library-go vendor in later commit, what do we need this for?

glide.yaml
```diff
@@ -11,7 +11,10 @@ import:
 - package: k8s.io/kube-aggregator
   version: kubernetes-1.12.5
 - package: k8s.io/apiextensions-apiserver
-  version: kubernetes-1.12.5
+  repo: https://github.com/openshift/kubernetes-apiextensions-apiserver
+  version: origin-4.0-kubernetes-1.12.4 # for nullable PR, instead of kubernetes-1.12.5
```
stlaz: I thought the change that allowed generations to be `null` was in library-go? I think I heard Michal Fojtik once say that we should avoid depending on our own kube forks. Could this be solved by kube 1.13 rebase?

glide.yaml
```diff
@@ -11,7 +11,10 @@ import:
...
+- package: sigs.k8s.io/controller-tools
+  repo: https://github.com/openshift/kubernetes-sigs-controller-tools
```
stlaz: What do we need this for? If it's for the CRD-generator, shouldn't it be a part of library-go deps if it provides the command?

Makefile
```diff
@@ -132,3 +132,12 @@ build-rpms:
 build-images:
 	hack/build-images.sh
 .PHONY: build-images
+
+# Generate CRD OpenAPIv3 validation 
+update-codegen-crds:
+	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --apis-dir vendor/github.com/openshift/api
+
+verify-codegen-crds:
```
stlaz: I think we should add this to the tests run by Travis.

glide.yaml
sallyom: checking on it.. will report back

sallyom: but many other operators are depending on that fork atm :woman_shrugging:

glide.yaml
damemi: library-go just has a script which calls the necessary components of controller-tools to generate openapi validations. glide strips transitive dependencies so we need to import the fork to be able to run the script here

stlaz: Ah, I see, thanks!

glide.yaml
sttts: > I think I heard Michal Fojtik once say that we should avoid depending on our own kube forks.

this is true until we have backports we depend on :-/

pkg/crd-schema-gen.go
damemi: this is to get glide to pick up the transitive dependencies needed for that script

glide.yaml
damemi: the nullable change is actually in kubernetes-apiextensions-apiserver, this fork is okay for now because we don't yet have the functionality in 1.13

Makefile
sallyom: I'm leaning toward moving this (at least) to a separate PR,we don't need this atm

pkg/crd-schema-gen.go
sallyom: thanks @damemi

sallyom:
@stlaz for this bump, I am pulling in the fork of kubernetes-apiextensions-apiserver that other operators are using, to pull in the ability for nullable fields.  Also pinning to library-go temporary-1.12  as kube-apiserver is also doing.  I've noted that those are temporary pins.

sallyom:
@enj it's green :eyes:

PR #99, opened 2019-03-24
Tighten APIs

enj:
Make default grant method impossible to use

Set the default grant method to deny so that all OAuth clients are
forced to set it explicitly.  The validation in origin will be
updated to mark grant method as required.

---

Infer challenge and login based on IDP

These fields offer a level of flexibility that is not particularly
useful to admins.  In reality all IDPs have a correct configuration
for these fields based on the feature set that the IDP supports.  We
should avoid asking users questions that we already know the answer
to.

---

Add issuer to OpenIDIdentityProvider

This change adds an issuer field which can be used to perform
discovery via the /.well-known/openid-configuration endpoint.  This
makes the OpenIDURLs struct obsolete.  It will be removed when this
change is made to the openshift/api repo.

---

OpenID: add discovery via /.well-known/openid-configuration

This change adds support for discovering OpenID configuration via
the issuer's /.well-known/openid-configuration endpoint.  This
reduces the config that is required (and the associated chance of
configuration errors) while preserving the same feature set.

Temporary fallback code was added to respect the old urls struct.
This allows an admin to migrate configuration and prevents upgrades
from breaking.

---

/assign @stlaz

### DIFF
--- a/pkg/operator2/idp.go
+++ b/pkg/operator2/idp.go
@@ -49,10 +49,16 @@ func init() {
 	utilruntime.Must(osinv1.Install(scheme))
 }
 
-func convertProviderConfigToOsinBytes(providerConfig *configv1.IdentityProviderConfig, syncData *configSyncData, i int) ([]byte, error) {
+type idpData struct {
+	provider  runtime.Object
+	challenge bool
+	login     bool
+}
+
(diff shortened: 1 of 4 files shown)


enj:
/retest

Terraform

enj:
/retest

Install

PR #101, opened 2019-03-25
Improve error messages

stlaz:
Adds more verbose error messages so that we know where the errors
we encounter come from


stlaz:
@enj moved all `%s` to `%v`

enj:
/retest

Terraform

PR #102, opened 2019-03-26
Bug 1691488: Fix route healthz to tolerate custom wildcard certs

enj:
This change builds a transport using the router-certs secret
provided by the ingress operator (this secret is expected to
contain both the CA and cert data).  If no CA data is found, it is
assumed that the admin has replaced the default wildcard certificate
with a globally valid cert.  To handle this, we fallback to the
default transport.

Remove use of rest.Config in favor of building custom transports.
This prevents us from emitting our service account token
unnecessarily.

Properly check well-known (we incorrectly skipped it).

Generally improve our readiness error reporting.

Bug 1691488

@openshift/sig-auth

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -18,7 +18,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/rest"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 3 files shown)


enj:
/hold

Not done testing yet.

enj:
/hold cancel

Tested successfully.

enj:
Tagging, beta blocker, passed local testing and CI OAuth test.

/retest

stlaz:
In the last run, the deployment was waiting 22 minutes for `v4-0-config-system-router-certs` from the start of the operator, not sure if that's reasonable.

The auth operator eventually turned Available, although then it lost `https://172.30.0.1/.well-known/oauth-authorization-server` again (404).

Anyway, let's
/retest

stlaz:
/retest
everything went 'nanners

PR #105, opened 2019-04-09
Bug 1694173: Improve status reporting

sallyom:
Currently we always set Available=False when we set Progressing=True and Available=True/Progressing=False.  This PR fine-tunes those to set independently when appropriate.

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -223,9 +223,13 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 
 	operatorConfigCopy := operatorConfig.DeepCopy()
 
+	// ClearFailingStatus
+	if v1helpers.IsOperatorConditionTrue(operatorConfigCopy.Status.Conditions, operatorv1.OperatorStatusTypeFailing) {
+		setFailingFalse(operatorConfigCopy)
+	}
 	syncErr := c.handleSync(operatorConfigCopy)
 	if syncErr != nil {
-		c.setFailingStatus(operatorConfigCopy, "OperatorSyncLoopError", syncErr.Error())
+		setFailingTrue(operatorConfigCopy, "OperatorSyncLoopError", syncErr.Error())
 	}
 
 	if _, _, err := v1helpers.UpdateStatus(c.authOperatorConfigClient, func(status *operatorv1.OperatorStatus) error {
@@ -376,7 +380,8 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 		if c.versionGetter.GetVersions()[operatorSelfName] != operatorVersion {
 			c.versionGetter.SetVersion(operatorSelfName, operatorVersion)
 		}
-		c.setAvailableStatus(operatorConfig)
+		setAvailableTrue(operatorConfig, "AsExpected")
+		setProgressingFalse(operatorConfig)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/operator2/operator.go
```diff
@@ -407,160 +412,175 @@ func (c *authOperator) checkReady(
...
 
-	wellknownReady, wellknownMsg, err := c.checkWellknownEndpointReady(authConfig, route)
+	wellknownReady, err := c.checkWellknownEndpointReady(authConfig, route, operatorConfig)
 	if err != nil {
 		return wellknownReady, fmt.Errorf("unable to check the .well-known endpoint: %v", err)
 	}
 	if !wellknownReady {
-		c.setProgressingStatus(operatorConfig, "WellknownNotReady", wellknownMsg)
+		setProgressingTrueAndAvailableFalse(operatorConfig, "WellKnownNotReady", err.Error())
 		return wellknownReady, nil
```
sallyom: need to fix that ^^

pkg/operator2/operator.go
```diff
@@ -407,160 +412,175 @@ func (c *authOperator) checkReady(
...
 	}
 
-	routeReady, routeMsg, err := c.checkRouteHealthy(route, routerSecret)
+	routeReady, err := c.checkRouteHealthy(route, routerSecret, operatorConfig)
 	if err != nil {
 		return routeReady, fmt.Errorf("unable to check route health: %v", err)
 	}
 	if !routeReady {
-		c.setProgressingStatus(operatorConfig, "RouteNotReady", routeMsg)
+		setProgressingTrueAndAvailableFalse(operatorConfig, "RouteNotReady", err.Error())
```
sallyom: ^^ need to fix that

pkg/operator2/operator.go
```diff
@@ -376,7 +380,8 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 		if c.versionGetter.GetVersions()[operatorSelfName] != operatorVersion {
 			c.versionGetter.SetVersion(operatorSelfName, operatorVersion)
 		}
-		c.setAvailableStatus(operatorConfig)
+		setAvailableTrue(operatorConfig, "ManagedResourcesSuccessfullyApplied")
```
sallyom: not sure about that msg

pkg/operator2/operator.go
```diff
@@ -407,160 +412,175 @@ func (c *authOperator) checkReady(
...
 	}
 
-	oauthClientsReady, oauthClientsMsg, err := c.oauthClientsReady(route)
+	oauthClientsReady, err := c.oauthClientsReady(route, operatorConfig)
 	if err != nil {
 		return oauthClientsReady, fmt.Errorf("unable to check OAuth clients' readiness: %v", err)
 	}
 	if !oauthClientsReady {
-		c.setProgressingStatus(operatorConfig, "OAuthClientsNotReady", oauthClientsMsg)
+		setProgressingTrueAndAvailableFalse(operatorConfig, "OAuthClientNotReady", err.Error())
```
sallyom: fix that, too

sallyom:
@enj I tested an upgrade to this PR payload and guess what? authentication CO was _not_ the very last CO to report new version & Available=True, first time I've seen that.  Upgrade was successful.

pkg/operator2/operator.go
```diff
@@ -430,40 +435,46 @@ func (c *authOperator) checkReady(
...
 
 	if osinDeployment.DeletionTimestamp != nil {
-		return false, "", fmt.Errorf("the deployment is being deleted")
+		setProgressingTrueAndAvailableFalse(operatorConfig, reason, "deployment is being deleted")
+		return false, nil
 	}
 
 	if osinDeployment.Status.UpdatedReplicas != osinDeployment.Status.Replicas || osinDeployment.Status.UnavailableReplicas > 0 {
-		return false, "not all deployment replicas are ready", nil
+		setProgressingTrue(operatorConfig, reason, "not all deployment replicas are ready")
```
enj: We are available here as long as we have one ready replica.

pkg/operator2/operator.go
enj: `AsExpected`?

sallyom: that works

pkg/operator2/operator.go
sallyom: right, updated, missed that!

PR #107, opened 2019-04-12
Check osin deployment availability after other checks

enj:
See comment in handleVersion.

@openshift/sig-auth 

/assign @stlaz @sallyom


pkg/operator2/status.go
```diff
@@ -48,15 +48,10 @@ func setProgressingFalse(operatorConfig *operatorv1.Authentication) {
 }
 
 func setProgressingTrueAndAvailableFalse(operatorConfig *operatorv1.Authentication, reason, message string) {
+	setProgressingTrue(operatorConfig, reason, message)
+
```
sallyom: that was not there?? :| wth,Sally!

wking:
Green :)  (although I don't understand this well enough to be able to review myself ;)

PR #108, opened 2019-04-12
Handle defaulting of top level config

enj:
This allows us to treat the empty value as a valid state.


enj:
/hold

Need minor change.

PR #111, opened 2019-04-18
bug 1700903: standardize resource names

ericavonb:
Remove "openshift-" prefix from the resource names for auth components

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -41,13 +41,15 @@ import (
 var deploymentVersionHashKey = operatorv1.GroupName + "/rvs-hash"
 
 const (
-	clusterOperatorName = "authentication"
-	targetName          = "openshift-authentication"
-	targetNameOperator  = "openshift-authentication-operator"
-	globalConfigName    = "cluster"
+	clusterOperatorName     = "authentication"
+	targetName              = "integrated-oauth-server"
+	targetNamespace         = "openshift-authentication"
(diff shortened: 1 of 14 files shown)


pkg/operator2/operator.go
```diff
@@ -41,13 +41,15 @@ import (
 var deploymentVersionHashKey = operatorv1.GroupName + "/rvs-hash"
 
 const (
-	clusterOperatorName = "authentication"
-	targetName          = "openshift-authentication"
-	targetNameOperator  = "openshift-authentication-operator"
-	globalConfigName    = "cluster"
+	clusterOperatorName     = "authentication"
+	targetName              = "oauth-server"
```
enj: Why `oauth-server` vs `integrated-oauth-server` to match existing APIs like authentication `spec.type`?

enj:
/hold

Need some reasoning on the name choice.

pkg/operator2/operator.go
```diff
@@ -41,13 +41,15 @@ import (
...
+	targetNamespace         = "openshift-authentication"
+	targetNameOperator      = "authentication-operator"
+	targetNamespaceOperator = "authentication-operator"
```
enj: For any const that you change, create a new variable name to make it clear what is getting updated.

Also, this const value is wrong.

ericavonb:
/retest
gotta love getting our own flake

PR #113, opened 2019-04-23
README: add steps for simple test IdP config

stlaz:
Probably looks a bit better than just being empty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,41 @@
 # cluster-authentication-operator
 
-soon?!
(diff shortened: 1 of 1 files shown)


stlaz:
Added `-B` to `htpasswd` commands so that this is now "very secure" (man pages, htpasswd, 23.04.2019).

enj:
/hold

err one sec

enj:
/hold cancel

Just was confirming the config.

PR #114, opened 2019-04-23
Bug 1694818: remove emoji from event logs

ericavonb:
Cause: operator used an emoji as the operator informer key

Consequence: error logs contain emoji character, confusing users

Fix: change key to more boring but informative value

Result: errors from operator controller shows "authentication-operator"
instead of emoji

### DIFF
--- a/pkg/boilerplate/operator/option.go
+++ b/pkg/boilerplate/operator/option.go
@@ -6,27 +6,19 @@ import (
 	"github.com/openshift/cluster-authentication-operator/pkg/boilerplate/controller"
 )
 
-// key is the singleton key shared by all events
-// the value is irrelevant, but pandas are awesome
-const key = "🐼"
-
 type Option func(*operator)
 
 func WithInformer(getter controller.InformerGetter, filter controller.Filter, opts ...controller.InformerOption) Option {
-	return toAppendOpt(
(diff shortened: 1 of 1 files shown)


pkg/boilerplate/operator/option.go
```diff
@@ -6,27 +6,19 @@ import (
...
 )
 
-// key is the singleton key shared by all events
-// the value is irrelevant, but pandas are awesome
-const key = "🐼"
-
 type Option func(*operator)
 
 func WithInformer(getter controller.InformerGetter, filter controller.Filter, opts ...controller.InformerOption) Option {
-	return toAppendOpt(
```
enj: /hold

Restore this function.  This repo has the canonical copy.

ericavonb: We can't use `o.name` with this function (service-ca-operator didn't use this function so the change worked). I'd rather remove the unnecessary indirection anyways.

enj: blah, sigh.  Shows we are leaking between the two use cases.  oh well.  I had added new options in some unmerged PRs but I guess I will have to deal with the extra boilerplate (or conditionally use `toAppendOpt`).

/hold cancel
/lgtm

PR #115, opened 2019-04-24
Pull the Failing -> Degraded rename + OpenID IdP API change

stlaz:
Keeping up with API changes


pkg/operator2/idp.go
```diff
@@ -233,16 +233,7 @@ func (c *authOperator) convertProviderConfigToIDPData(providerConfig *configv1.I
...
-func (c *authOperator) discoverOpenIDURLs(issuer, key string, ca configv1.ConfigMapNameReference, fallbackURLs configv1.OpenIDURLs) (*osinv1.OpenIDURLs, error) {
-	// TODO drop after 4.0beta3
-	if len(issuer) == 0 {
-		return &osinv1.OpenIDURLs{
-			Authorize: fallbackURLs.Authorize,
-			Token:     fallbackURLs.Token,
-			UserInfo:  fallbackURLs.UserInfo,
-		}, nil
-	}
-
```
sallyom: I _think_ this should be good now, since we're beta4

stlaz:
Unauthorized :octopus: 
/retest

pkg/operator2/idp.go
enj: I believe this is fine now.

PR #120, opened 2019-04-29
Bug 1692290: Sort RVs before hashing them

stlaz:
Bugzilla 1692290
cc @enj


enj:
RVs are unique and monotonically increasing so this should not cause any issues.

We get the RVs of synced config maps and secrets from a map meaning the iteration order is undefined.  Thus it makes sense if having a large number of IDPs would cause us to never stabilize.

/lgtm

PR #126, opened 2019-05-06
Bug 1706894: Make oauth-openshift.<ingress_domain> the permanent issuer of the OAuth server

enj:
We need a permanently stable issuer URL for the OAuth server - both
in terms of the OAuth spec and so that OAuth callbacks from OAuth
IDPs continue to work indefinitely (we have no way to reconcile the
external IDP on a change).

oauth-openshift is the shortest namespaced value we can use (we do
not want to use a value that a customer may want to use).  This is a
best effort attempt to make the external API "nice" to use.

To keep things consistent, everything that was previously exposed as
"integrated-oauth-server" is now "oauth-openshift" (this does make
us slightly inconsistent with the top level Authentication config
and the binary of the OAuth server - we can at least update the
latter in the future for consistency).

Bug 1706894

### DIFF
--- a/manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
+++ b/manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
@@ -28,7 +28,7 @@ spec:
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: integrated-oauth-server
+  name: oauth-openshift
   namespace: openshift-authentication
 spec:
   endpoints:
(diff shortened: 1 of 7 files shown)


enj:
Upgrades failing because backwards incompatible naming changes to both deployment and route host.  e2e-aws is green.  Confirmed with team on slack.  Merging.

enj:
Crash loop is expected due to SA name change:

```
Command "openshift-osinserver" is deprecated, will be removed in 4.0
W0506 17:18:09.073126       1 authentication.go:272] Unable to get configmap/extension-apiserver-authentication in kube-system.  Usually fixed by 'kubectl create rolebinding -n kube-system ROLE_NAME --role=extension-apiserver-authentication-reader --serviceaccount=YOUR_NS:YOUR_SA'
F0506 17:18:09.073236       1 cmd.go:53] configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:openshift-authentication:integrated-oauth-server" cannot get resource "configmaps" in API group "" in the namespace "kube-system"
```

PR #129, opened 2019-05-08
Move boilerplate to shared location

enj:
/hold

4.2


benjaminapetersen:
Eventually going to make it to `library-go`? Or probably going to stay external?

enj:
> Eventually going to make it to `library-go`? Or probably going to stay external?

Not anytime soon.

pkg/operator2/operator.go
```diff
@@ -10,6 +10,9 @@ import (
 	"strconv"
 	"strings"
 
+	"monis.app/go/openshift/controller"
+	"monis.app/go/openshift/operator"
```
wking: Hmm, @enj do these need to get moved back into an OpenShift repo (library-go?) in the wake of #205?

enj: https://github.com/openshift/operator-boilerplate-legacy

PR #130, opened 2019-05-30
Remove use of legacyconfig in session

enj:
xref: https://github.com/openshift/api/pull/327


stlaz:
/lgtm
Do we want this minus the bump in 4.1?

enj:
> /lgtm
> Do we want this minus the bump in 4.1?

No because we do not have the osin v1 session type in openshift/api @ 4.1

PR #136, opened 2019-06-19
[WIP] Handle the challenging and browser oauth clients

stlaz:
This PR adds a controller sync loop that takes care of making sure the oauth clients are present when we are in `Managed` state and running the integrated OAuth server.


pkg/operator2/oauthclients.go
```diff
@@ -0,0 +1,78 @@
...
+	routev1 "github.com/openshift/api/route/v1"
+	oauthdiscovery "github.com/openshift/library-go/pkg/oauth/oauthdiscovery"
+)
+
+func defaultBrowserOAuthClient(route *routev1.Route) *oauthv1.OAuthClient {
+	return &oauthv1.OAuthClient{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: oauthBrowserClientName,
+		},
+		Secret:       randomString(sha256.BlockSize),
```
stlaz: I'm not entirely sure about the requirements for this secret so this might be wrong

pkg/operator2/starter.go
```diff
@@ -70,6 +71,8 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		routeinformer.WithTweakListOptions(singleNameListOptions(targetName)),
 	)
 
+	oauthClientInformers := oauthinformer.NewSharedInformerFactoryWithOptions(oauthClient, resync)
```
enj: You did not start this.

pkg/operator2/operator.go
```diff
@@ -390,6 +393,18 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 	// BLOCK 4: deployment
 	// ==================================
 
+	browserOAuthClient, err := c.handleBrowserOAuthClient(route)
```
enj: I would prefer this to be a distinct controller sync loop.

pkg/operator2/operator.go
```diff
@@ -440,7 +455,6 @@ func (c *authOperator) handleVersion(
 	// Checks readiness of all of:
 	//    - route
 	//    - well-known oauth endpoint
-	//    - oauth clients
```
enj: I would probably leave this bit as-is.

pkg/operator2/oauthclients.go
enj: you want 256 bits.

pkg/operator2/oauthclients.go
```diff
@@ -0,0 +1,78 @@
...
+	browserClient, err := c.oauthClientClient.Get(oauthBrowserClientName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		browserClient, err = c.oauthClientClient.Create(defaultBrowserOAuthClient(route))
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if browserClient != nil && !isValidBrowserOAuthClient(route, browserClient) {
+		c.oauthClientClient.Delete(oauthBrowserClientName, &metav1.DeleteOptions{})
```
enj: I dont think we want deletes because they can cause disruption.

pkg/oauthclientscontroller/oauthclients.go
```diff
@@ -0,0 +1,98 @@
...
+	return err
+}
+
+func (c *oauthClientsController) ensureChallengingOAuthClient(route *routev1.Route) error {
+	_, _, err := ApplyOAuthClient(c.oauthClientGetter, nil, defaultChallengingOAuthClient(route))
+	return err
+}
+
+// ApplyOAuthClient merges objectmeta. It returns the final Object, whether any change as made, and an error
+func ApplyOAuthClient(
```
stlaz: This is written so that it matches the library-go functions (that's why `recorder` is unused - the functions that would leverage it are not exported in the lib).

I guess I should move this to the appropriate lib if we decide to use it.

pkg/operator2/starter.go
stlaz: It is started now \o/

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
+package oauthclientscontroller
+
+import (
+	"monis.app/go/openshift/controller"
```
deads2k: wtf...

stlaz: yeah, I did not notice that move in our operators originally, either :man_shrugging:

pkg/util/constants.go
```diff
@@ -0,0 +1,10 @@
+package util
```
enj: No package named util.  Also this is basically api stuff so `pkg/api` might be OK place to start.

pkg/util/util.go
```diff
@@ -0,0 +1,24 @@
+package util
```
enj: `pkg/crypto` until you move it.

enj: Use `Random256BitsString`

```go
package crypto

import (
	"crypto/rand"
	"encoding/base64"
)

// RandomBits returns a random byte slice with at least the requested bits of entropy.
// Callers should avoid using a value less than 256 unless they have a very good reason.
func RandomBits(bits int) []byte {
	size := bits / 8
	if bits%8 != 0 {
		size++
	}
	b := make([]byte, size)
	if _, err := rand.Read(b); err != nil {
		panic(err) // rand should never fail
	}
	return b
}

// RandomBitsString returns a random string with at least the requested bits of entropy.
// It uses RawURLEncoding to ensure we do not get / characters or trailing ='s.
func RandomBitsString(bits int) string {
	return base64.RawURLEncoding.EncodeToString(RandomBits(bits))
}

// Random256BitsString is a convenience function for calling RandomBitsString(256).
// Callers that need a random string should use this function unless they have a
// very good reason to need a different amount of entropy.
func Random256BitsString() string {
	// 32 bytes (256 bits) = 43 base64-encoded characters
	return RandomBitsString(256)
}
```

pkg/oauthclientscontroller/oauthclients.go
```diff
@@ -0,0 +1,98 @@
...
+	if contentSame && !*modified {
+		return existingCopy, false, nil
+	}
+
+	objectMeta := existingCopy.ObjectMeta.DeepCopy()
+	existingCopy = required.DeepCopy()
+	existingCopy.ObjectMeta = *objectMeta
+
+	if klog.V(4) {
+		klog.Infof("OAuthClient %q changes: %v", required.Name, resourceapply.JSONPatch(existing, existingCopy))
```
enj: You are logging an object that contains secrets.

pkg/oauthclientscontroller/oauthclients.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	modified := resourcemerge.BoolPtr(false)
+	existingCopy := existing.DeepCopy()
+
+	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+	contentSame := equality.Semantic.DeepEqual(existingCopy, required)
+	if contentSame && !*modified {
```
enj: This will not reconcile changes to the other fields, which we care about.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+
+func isBootstrappedClient(metadata metav1.Object) bool {
+	if name := metadata.GetName(); name == authoputil.OAuthBrowserClientName ||
+		name == authoputil.OAuthChallengingClientName {
+		return true
+	}
+	return false
+}
+
+func isAuthTypeIntegratedOAuthServer(t configv1.AuthenticationType) bool {
```
enj: Use the defaulting logic we already have.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+		err = c.ensureBrowserOAuthClient(oauthClient, route)
+
+	case authoputil.OAuthChallengingClientName:
+		err = c.ensureChallengingOAuthClient(route)
+	}
+
+	return err
+}
+
+func isBootstrappedClient(metadata metav1.Object) bool {
```
enj: This is just a filter by name.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+		oauthClientGetter: oauthClientGetter,
+		routeClient:       routeClient,
+	}
+
+	targetNameFilter := controller.FilterByNames(nil, authoputil.TargetName)
+	configNameFilter := controller.FilterByNames(nil, authoputil.GlobalConfigName)
+
+	return controller.New("OAuthClientsController", c,
+		controller.WithInformer(oauthClientInformer, controller.FilterFuncs{
+			AddFunc: isBootstrappedClient,
```
enj: `FilterByNames`...

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+			return err
+		}
+
+		if c.ensureChallengingOAuthClient(route); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	oauthClient := obj.(*oauthv1.OAuthClient)
```
enj: I would like this set of clients to be `map` based so that the console operator can re-use this code.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+	c := &oauthClientsController{
+		authOperatorClient: authOpConfigClient,
+
+		authConfigClient: authConfigClient,
+
+		oauthClientGetter: oauthClientGetter,
+		routeClient:       routeClient,
+	}
+
+	targetNameFilter := controller.FilterByNames(nil, authoputil.TargetName)
```
enj: These do not make sense without parent funcs.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+	configNameFilter := controller.FilterByNames(nil, authoputil.GlobalConfigName)
```
enj: And I do not think this makes sense at all because you cannot make a parent func for it.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+	authConfig, err := c.authConfigClient.Authentications().Get(authoputil.GlobalConfigName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Don't do anything if the operator is not in Managed state or the integrated
+	// OAuth server is not supposed to run in the cluster
+	// Eventually we may want to remove the oauthclients in certain states
+	if authOperatorConfig.Spec.ManagementState != operatorv1.Managed ||
+		!isAuthTypeIntegratedOAuthServer(authConfig.Spec.Type) {
```
enj: The authentication type does not matter.  Keycloak will reuse the existing OAuth clients.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil
+	}
+
+	route, err := c.routeClient.Get(authoputil.TargetName, metav1.GetOptions{})
```
enj: route name needs to be pluggable for console-operator reuse.

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return err
+	}
+
+	if obj == nil { // the client was probably deleted -> ensure both clients exist
```
enj: This is the wrong construct.  You only reconcile what matches your key `namespace, name`

pkg/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -0,0 +1,147 @@
...
+		authConfigClient: authConfigClient,
+
+		oauthClientGetter: oauthClientGetter,
+		routeClient:       routeClient,
+	}
+
+	targetNameFilter := controller.FilterByNames(nil, authoputil.TargetName)
+	configNameFilter := controller.FilterByNames(nil, authoputil.GlobalConfigName)
+
+	return controller.New("OAuthClientsController", c,
```
enj: I do not think this handles the "there are no OAuth clients at all" case correctly (the weird `nil` construct you have does not count).

stlaz:
After thorough thought I'm not sure about how to handle this all correctly - we would like each sync of the controller to react to a specific OAuth client. That would be easy BUT we also need to react to route changes where the relation is `route : oauthclient -> 1 : N` which means a parent filter can't be created for situation as that alway only generated one couple of `namespace, name`.

We can't ignore a route change as logins would stop working from then on. For that sake I'm wondering:
1. should there be but only controller that handles routes and handles all field updates in oauthclients
2. or should we have a separate controller for routes that would simply care for the oauthclient urls and separate for oauthclients' other fields?

I guess 1. does not make sense as we would still have to react to both objects which we don't want to do with our controllers.

@enj Is the direction that there should be two separate controllers that would each look after certain fields correct? It might be a bit more work as originally anticipated as we would create another controller with the route code we already have, but it seems to be heading the way we intended for the operator anyway.

stlaz:
Reworked the whole thing, this is WIP and quite fragile

stlaz:
/close
this is only rotting here now, we can write it from scratch with the comments from this PR in mind

PR #137, opened 2019-06-19
Make hack scripts work with go 1.12

enj:


sallyom:
I've verified ./hack/verify-govet.sh runs smoothly with this PR : )

PR #138, opened 2019-06-24
Simple move of the OAuth clients bootstrap

stlaz:
Move oauthclients bootstrap from the integrated OAuth server

------
This is to unblock any work that depends on OAuth clients bootstrap being done separately from the integrated OAuth server running.

I had trouble creating a cluster today so it's not yet tested.

The code should eventually be replaced with the outcome of https://github.com/openshift/cluster-authentication-operator/pull/136

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-authentication-operator
 COPY . .
 RUN go build -tags="ocp" -o authentication-operator ./cmd/authentication-operator
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.2:base
 COPY --from=builder /go/src/github.com/openshift/cluster-authentication-operator/authentication-operator /usr/bin/
 COPY manifests /manifests
 ENTRYPOINT ["/usr/bin/authentication-operator"]
--- a/pkg/operator2/oauthclients.go
+++ b/pkg/operator2/oauthclients.go
@@ -1,0 +1,103 @@
+package operator2
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+
+	oauthv1 "github.com/openshift/api/oauth/v1"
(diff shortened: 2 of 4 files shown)


sttts:
"I had trouble creating a cluster today so it's not yet tested."

Should it before merging?

sttts:
> Move oauthclients bootstrap from the integrated OAuth server

Link?

Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
```
enj: Update `Dockerfile.rhel` also

stlaz: Done

stlaz:
@sttts It was a problem with the installer ex

> "I had trouble creating a cluster today so it's not yet tested."
> 
> Should it before merging?

It was trouble with the installer, it was timing out w/o this patch in, I believe I saw the issue reported on Friday already, after building the fresh master I was able to get a cluster eventually.

The link for the moved code is now to be found in the origin PR that I linked to this one

pkg/operator2/oauthclients.go
```diff
@@ -0,0 +1,103 @@
...
+	"k8s.io/client-go/util/retry"
+
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	oauthclient "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	"github.com/openshift/library-go/pkg/oauth/oauthdiscovery"
+)
+
+// EnsureBootstrapOAuthClients creates or updates the bootstrap oauth clients that openshift relies upon.
+func (c *authOperator) ensureBootstrappedOAuthClients(masterPublicURL string) error {
+	const browserSecretByteLen = 32
```
enj: dead code

pkg/operator2/oauthclients.go
```diff
@@ -0,0 +1,103 @@
...
+
+		existingCopy := existing.DeepCopy()
+
+		if len(existingCopy.Secret) == 0 {
+			existingCopy.Secret = client.Secret
+		}
+
+		existingCopy.RedirectURIs = client.RedirectURIs
+
+		// If the GrantMethod is present, keep it for compatibility
```
enj: we probably do not want this

pkg/operator2/oauthclients.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		existing, err := oauthClients.Get(client.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		existingCopy := existing.DeepCopy()
+
+		if len(existingCopy.Secret) == 0 {
```
enj: this actually need a bit more nuance but is good enough for now

enj:
/lgtm

All this code should get deleted this week anyway so let us not block others.

enj:
/lgtm cancel

Seems like this breaks kube:admin

enj:
#139 has only minor changes on top of this PR.

/close

PR #144, opened 2019-07-02
Adding License section in readme.md file.

kshithijiyer:


stlaz:
The LICENSE file is to be found in the root of the repo, I don't see how README would be any different from all the other files to explicitly mention it. Closing.

PR #147, opened 2019-07-02
Honor CORS set via APIServer config

enj:

@stlaz we need validation for the APIServer config to make sure each value is a valid regex.


enj:
@stlaz the CRD schemas in the config operator also need to be updated.

stlaz:
/retest
/lgtm
I'll handle the config-operator changes and will work on the validation, then

PR #150, opened 2019-07-10
Remove conditions that are no longer set

enj:
This change adds a generic control loop to remove conditions that we no longer set.  This prevents orphaned conditions from confusing the ClusterOperatorStatusController's union logic.

xref: #142 

@stlaz @crawford @smarterclayton

### DIFF
--- a/pkg/operator2/legacyconditions/legacyconditions.go
+++ b/pkg/operator2/legacyconditions/legacyconditions.go
@@ -1,0 +1,29 @@
+package legacyconditions
+
+import (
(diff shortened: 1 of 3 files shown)


stlaz:
/retest
Code looks good to me, shall we wait for what happens in https://github.com/openshift/cluster-authentication-operator/pull/142?

enj:
/hold

Need to port https://github.com/openshift/cluster-kube-scheduler-operator/pull/142 to lib-go and re-use.

enj:
#159

/close

PR #151, opened 2019-07-15
Bug 1729289:  Fix bug where favicon is incorrect (was Shadowman, now OpenShift)

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1729289

@spadgett, @enj not sure how to test locally.  The favicon for OKD/OSE is correct as is (is the OKD icon.


enj:
/cherrypick release-4.1

I will test locally and tag if it works.

stlaz:
/lgtm
Appears to work on my cluster

Issue #141, opened 2019-06-28
things david doesn't like


deads2k:
1. https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/configmap.go#L55-L59 .  route.spec isn't the same as route.status. 
 Your route hasn't been accepted yet, so the kube-apiserver could end up forwarding to an endpoint that isn't you!
2.  This configmap is an output, not an input.  Please make it a separate control loop to keep your main loop tight with logical dependencies.  Also, you don't want a rollout failure to disrupt keeping this outbound information correct.
https://github.com/openshift/cluster-authentication-operator/blob/04eedbb3751852155db618bfd6e9ffa27f836f74/pkg/operator2/operator.go#L321-L326
3. Use a different `FooDegraded` for each control loop and allow the status union to combine them.  It will make each condition write more obvious.
4. `handleAuthConfig` is outbound state.  Move into a different sync loop
5. your check functions are good, but they should each set different `FooDegraded` conditions.
6. `serviceCA, servingCert, err := c.handleServiceCA()` appears unnecessary. Directly depend on the key and the kubelet will properly put your pod into pending.
7. hard code the hardcoded values in your oauth config.  Service name for instance.  Flexibility that isn't flexible is really hard to read.
8. instead of passing a bunch of state through various handles, use your clients to lookup values.  You can make them caching if the load becomes too high.
9. `accessTokenInactivityTimeoutSeconds` appears to be a default.  why didn't you default it?
10. `handleOAuthConfig` appears to be an attempt at combining multiple different configobservers into a single loop.  You do logically own all these things, but configobservation (even a single value) distinct from the main loop will give you working generations and logicaly separation you're lacking here.
11. sync-ing of user config is driven outside the main loop of other operators to ensure that it always works regardless of the state of other rollouts.  Secrets must rotate without choice, the old values are invalid.
12. ensureBootstrappedOAuthClients looks completely distinct.
13. looks like availability may be worth separating like static pod operators that have complicated availability rules.

enj:
100% agree with all above except:

>     1. https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/configmap.go#L55-L59 .  route.spec isn't the same as route.status.
>        Your route hasn't been accepted yet, so the kube-apiserver could end up forwarding to an endpoint that isn't you!

[...]

PR #154, opened 2019-07-24
WIP: Honor the None authentication type and the WebhookTokenAuthenticators

JAORMX:
When the None authentication type is set, we shouldn't be adding the
reference to the IntegratedOAuthMetadata config map. This way the
kube-apiserver-operator will nt attempt to track the configmap.

### DIFF
--- a/pkg/operator2/auth.go
+++ b/pkg/operator2/auth.go
@@ -1,6 +1,8 @@
 package operator2
 
 import (
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


JAORMX:
@enj is this going on the right path?

enj:
> @enj is this going on the right path?

I suggest holding off for now.  A lot of this code needs to be moved around to address #141.  Plus we are in bugfix mode right now anyway.

PR #155, opened 2019-07-24
deployment: add env vars from the Proxy object

stlaz:
The Proxy object specifies certain http-proxy related environmnent
variables that we need setting for the operand's cotainer.

/cc @sttts @enj

### DIFF
--- a/pkg/operator2/deployment.go
+++ b/pkg/operator2/deployment.go
@@ -22,6 +22,7 @@ func defaultDeployment(
 	operatorConfig *operatorv1.Authentication,
 	syncData *configSyncData,
 	routerSecret *corev1.Secret,
+	proxyConfig *configv1.Proxy,
 	operatorDeployment *appsv1.Deployment,
 	resourceVersions ...string,
 ) *appsv1.Deployment {
@@ -121,6 +122,7 @@ func defaultDeployment(
 								},
 							},
 							VolumeMounts:             mounts,
+							Env:                      proxyConfigToEnvVars(proxyConfig),
 							ReadinessProbe:           defaultProbe(),
 							LivenessProbe:            livenessProbe(),
 							TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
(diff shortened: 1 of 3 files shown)


stlaz:
Let's deal with the CA bundle once the networking team actually provides us with one -> can be reviewed.

pkg/operator2/proxy.go
```diff
@@ -0,0 +1,18 @@
...
+	"k8s.io/klog"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func (c *authOperator) handleProxyConfig() *configv1.Proxy {
+	proxyConfig, err := c.proxy.Get(globalConfigName, metav1.GetOptions{})
+	if err != nil {
+		klog.Infof("error getting proxy config: %v", err)
+		// have a placeholder that will at least look reasonable in the token request endpoint
```
enj: remove incorrect comment

pkg/operator2/operator.go
```diff
@@ -400,6 +403,9 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 		return err
 	}
 
+	proxyConfig := c.handleProxyConfig()
+	resourceVersions = append(resourceVersions, proxyConfig.GetResourceVersion())
```
enj: It annoys me that this adds another thing that needs explicit RV tracking.  Oh well.

Use `proxyConfig.ResourceVersion` directly - the generic methods should only be used when dealing with Object Meta generically, not when you have a specific type.

pkg/operator2/deployment.go
```diff
@@ -236,6 +238,24 @@ func getImagePullPolicy(operatorDeployment *appsv1.Deployment) corev1.PullPolicy
 	return containers[0].ImagePullPolicy
 }
 
+func proxyConfigToEnvVars(proxy *configv1.Proxy) []corev1.EnvVar {
+	envVars := []corev1.EnvVar{}
+
+	if noProxy := proxy.Spec.NoProxy; len(noProxy) > 0 {
```
enj: nit: could make a helper function to handle the conditional append logic

pkg/operator2/deployment.go
```diff
@@ -121,6 +122,7 @@ func defaultDeployment(
 								},
 							},
 							VolumeMounts:             mounts,
+							Env:                      proxyConfigToEnvVars(proxyConfig),
```
enj: How is the stuff for the operator itself going?  I see this is meant for the OAuth server.

pkg/operator2/deployment.go
```diff
@@ -236,6 +238,24 @@ func getImagePullPolicy(operatorDeployment *appsv1.Deployment) corev1.PullPolicy
 	return containers[0].ImagePullPolicy
 }
 
+func proxyConfigToEnvVars(proxy *configv1.Proxy) []corev1.EnvVar {
```
enj: This assumes that the transports that the OAuth server uses honors these proxy headers.  I would like an e2e to prove that.

pkg/operator2/deployment.go
stlaz: That should be handled from CVO, given that the operator itself can be considered its operand:
https://github.com/openshift/cluster-version-operator/pull/224

pkg/operator2/operator.go
stlaz: I guess most of it could be replaced with the config-observation logic as done in {K,O}AS-O, we'd probably only be left with checking the resource versions of configmaps/secrets (for CA/cert+priv key updates)

pkg/operator2/deployment.go
sttts: Testing is huge topic that is still unaddressed throughout the whole proxy initiative. This will be follow-up.

PR #156, opened 2019-07-25
Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in
golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile;
It also ignores a false positive warning that was found.


tools/import-verifier/import-verifier.go
```diff
@@ -319,6 +319,7 @@ func loadImportRestrictions(configFile string) ([]ImportRestriction, error) {
 func resolvePackage(targetPackage string) ([]Package, error) {
 	cmd := "go"
 	args := []string{"list", "-json", targetPackage}
+	// #nosec
```
stlaz: Please add an explanation about why you are turning off the check here

JAORMX: Will do; should I do it as part of the PR description or a comment in the code?

JAORMX: done!

enj:
Not opposed to this but we want this kind of stuff in the library-go Makefile thing and we need to move this repo to using that.

PR #157, opened 2019-07-25
Remove degraded

deads2k:


pkg/operator2/starter.go
```diff
@@ -156,6 +157,16 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 
+	staleConditions := staleconditions.NewRemoveStaleConditions(
+		[]string{
+			// in 4.1.0 this was accidentally in the list.  This can be removed in 4.3.
+			"Degraded",
```
enj: Use the const.

pkg/operator2/starter.go
```diff
@@ -2,6 +2,7 @@ package operator2
 
 import (
 	"fmt"
+	"github.com/openshift/library-go/pkg/operator/staleconditions"
```
enj: move down.

pkg/operator2/starter.go
deads2k: > Use the const.

? the const should be deleted.  It's a const that shouldn't exist because any usage of it is wrong.

deads2k:
if https://github.com/openshift/cluster-kube-scheduler-operator/pull/159 works, then I think this is ready.

pkg/operator2/starter.go
enj: Please follow-up and actually delete the const.

enj:
#159

/close

PR #158, opened 2019-07-25
[WIP]  make a generic operator client

deads2k:
I want a generic operator client.  I had this repo open to prove it works. 

@damemi how do you like it?


enj:
I believe this is in library-go now.

PR #159, opened 2019-07-25
Remove degraded condition

enj:


deads2k:
tag if https://github.com/openshift/cluster-kube-scheduler-operator/pull/159 works

deads2k:
> tag if [openshift/cluster-kube-scheduler-operator#159](https://github.com/openshift/cluster-kube-scheduler-operator/pull/159) works

the e2e test we have testing conditional removal passes. tagging

/lgtm

PR #160, opened 2019-07-29
Bug 1733336: Correct race between config sync loop and resource syncer

enj:
This patch makes two primary changes:

1. Generically track resource versions:

We depend on the following state:
- The operator deployment itself
- The authentications.operator.openshift.io resource
- Config maps and secrets that are owned by the operator
- The current observed (CLI) config (stored in a config map)
- Config maps and secrets that are synced from openshift-config

Thus by tracking the:
- operator deployment resource version
- authentications.operator.openshift.io generation
- resource version of config maps and secrets with config prefix

we can guarantee that we redeploy when our state has changed.

2. Correct race between config sync loop and resource syncer

The code in master incorrectly assumed that the calls to List
contained all of the config maps and secrets that were being synced.
Since the resource syncer cannot instantaneously sync resources, the
List would race against it.  Thus the logic to determine what data
was in use and the corresponding resource versions would produce
incorrect results due to invalid input.  This change completely
removes the need for the config sync code to track any resource
versions.  It also moves the resource syncer calls up with
additional blocking code below it.  This blocking code will be
retried until the calls to List and resource syncer have caught up.
This guarantees that the logic to determine what data is in use does
not produce incorrect results from invalid input.

Adds unit tests for all of the above.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ verify-commits:
 #   make test-unit
 #   make test-unit WHAT=pkg/build TESTFLAGS=-v
 test-unit:
-	GOTEST_FLAGS="$(TESTFLAGS)" hack/test-go.sh $(WHAT) $(TESTS)
+	GOTEST_FLAGS="$(TESTFLAGS)" GODEBUG=tls13 KUBERNETES_SERVICE_PORT_HTTPS=443 hack/test-go.sh $(WHAT) $(TESTS)
 .PHONY: test-unit
 
 test-sec:
--- a/pkg/operator2/branding.go
+++ b/pkg/operator2/branding.go
@@ -5,7 +5,6 @@ import (
 
 	"gopkg.in/yaml.v2"
 
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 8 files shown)


enj:
The expected transient error messages can be seen in [e2e-aws-console-login](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/160/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-console-login/60/artifacts/e2e-aws-console-login/pods/openshift-authentication-operator_authentication-operator-58d746d66d-6vrtv_operator.log):

```
E0729 00:42:36.350024       1 controller.go:129] {AuthenticationOperator2 AuthenticationOperator2} failed with: failed syncing configuration objects: secrets [v4-0-config-user-idp-0-file-data] in openshift-authentication not synced
```

Thus we are properly gating deployment rollouts until all config has been synced.

stlaz:
/lgtm

Tested manually and it appears to work, the approach seems fine.

I was hoping I'd be able to get the code duplication reduced but failed horribly.

PR #162, opened 2019-07-29
Bug 1733978: Correct race between config sync loop and resource syncer [4.1]

stlaz:
This patch makes two primary changes:

1. Generically track resource versions:

We depend on the following state:
- The operator deployment itself
- The authentications.operator.openshift.io resource
- Config maps and secrets that are owned by the operator
- The current observed (CLI) config (stored in a config map)
- Config maps and secrets that are synced from openshift-config

Thus by tracking the:
- operator deployment resource version
- authentications.operator.openshift.io generation
- resource version of config maps and secrets with config prefix

we can guarantee that we redeploy when our state has changed.

2. Correct race between config sync loop and resource syncer

The code in master incorrectly assumed that the calls to List
contained all of the config maps and secrets that were being synced.
Since the resource syncer cannot instantaneously sync resources, the
List would race against it.  Thus the logic to determine what data
was in use and the corresponding resource versions would produce
incorrect results due to invalid input.  This change completely
removes the need for the config sync code to track any resource
versions.  It also moves the resource syncer calls up with
additional blocking code below it.  This blocking code will be
retried until the calls to List and resource syncer have caught up.
This guarantees that the logic to determine what data is in use does
not produce incorrect results from invalid input.

Adds unit tests for all of the above.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ verify-commits:
 #   make test-unit
 #   make test-unit WHAT=pkg/build TESTFLAGS=-v
 test-unit:
-	GOTEST_FLAGS="$(TESTFLAGS)" hack/test-go.sh $(WHAT) $(TESTS)
+	GOTEST_FLAGS="$(TESTFLAGS)" GODEBUG=tls13 KUBERNETES_SERVICE_PORT_HTTPS=443 hack/test-go.sh $(WHAT) $(TESTS)
 .PHONY: test-unit
 
 # Run e2e tests.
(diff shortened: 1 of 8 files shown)


stlaz:
keeping https://github.com/openshift/cluster-authentication-operator/pull/161 opened should the CI decide master ref is good ref again

enj:
My guess is that the console login test will work after https://github.com/openshift/release/pull/4559 merges.

enj:
Blocked by https://github.com/openshift/console/pull/2209

jwforres:
Hasn't had soak time yet since master just merged today. Will likely wait for next week.

derekwaynecarr:
Dependent BZ has been VERIFIED.
Approved for 4.1.z

PR #172, opened 2019-08-09
Inject payload's system store with proxy CA when specified

stlaz:
/cc @enj @sttts

### DIFF
--- a/manifests/03_configmap.yaml
+++ b/manifests/03_configmap.yaml
@@ -7,3 +7,13 @@ data:
   operator-config.yaml: |
     apiVersion: operator.openshift.io/v1alpha1
     kind: GenericOperatorConfig
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-authentication
+  name: v4-0-config-system-trusted-ca-bundle
+  annotations:
+    release.openshift.io/create-only: "true"
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
--- a/manifests/03_configmap.yaml
+++ b/manifests/03_configmap.yaml
@@ -11,6 +11,16 @@ data:
 apiVersion: v1
 kind: ConfigMap
 metadata:
+  namespace: openshift-authentication-operator
+  name: trusted-ca-bundle
+  annotations:
+    release.openshift.io/create-only: "true"
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
   namespace: openshift-authentication
   name: v4-0-config-system-trusted-ca-bundle
   annotations:
--- a/manifests/07_deployment.yaml
+++ b/manifests/07_deployment.yaml
@@ -36,6 +36,9 @@ spec:
           name: config
         - mountPath: /var/run/secrets/serving-cert
           name: serving-cert
+        - mountPath: /var/run/configmaps/trusted-ca-bundle
+          name: trusted-ca-bundle
+          readOnly: true
         env:
         - name: IMAGE
           value: quay.io/openshift/origin-oauth-server:v4.2
@@ -53,6 +56,10 @@ spec:
         configMap:
           defaultMode: 440
           name: authentication-operator-config
+      - name: trusted-ca-bundle
+        configMap:
+          name: trusted-ca-bundle
+          optional: true
       - name: serving-cert
         secret:
           secretName: serving-cert
--- a/pkg/operator2/deployment.go
+++ b/pkg/operator2/deployment.go
@@ -71,6 +71,14 @@ func defaultDeployment(
 			path:      ocpBrandingSecretMount,
 			keys:      []string{configv1.LoginTemplateKey, configv1.ProviderSelectionTemplateKey, configv1.ErrorsTemplateKey},
 		},
+		{
+			name:      trustedCABundleName,
+			configmap: true,
+			path:      trustedCABundleMountDir,
+			mappedKeys: map[string]string{
+				trustedCABundleKey: trustedCABundleMountFile,
+			},
+		},
 	} {
 		v, m := data.split()
 		volumes = append(volumes, v)
@@ -257,10 +265,11 @@ func appendEnvVar(envVars []corev1.EnvVar, envName, envVal string) []corev1.EnvV
 }
 
 type volume struct {
-	name      string
-	configmap bool
-	path      string
-	keys      []string
+	name       string
+	configmap  bool
+	path       string
+	keys       []string
+	mappedKeys map[string]string
 }
 
 func (v *volume) split() (corev1.Volume, corev1.VolumeMount) {
@@ -269,6 +278,14 @@ func (v *volume) split() (corev1.Volume, corev1.VolumeMount) {
 	}
 
 	var items []corev1.KeyToPath
+	// maps' keys are random,  we need to sort the output to prevent redeployment hotloops
+	for _, key := range sets.StringKeySet(v.mappedKeys).List() {
+		items = append(items, corev1.KeyToPath{
+			Key:  key,
+			Path: v.mappedKeys[key],
+		})
+	}
+
 	for _, key := range v.keys {
 		items = append(items, corev1.KeyToPath{
 			Key:  key,
--- a/pkg/operator2/idp.go
+++ b/pkg/operator2/idp.go
@@ -3,9 +3,12 @@ package operator2
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
+
+	"k8s.io/klog"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -293,7 +296,7 @@ func (c *authOperator) discoverOpenIDURLs(issuer, key string, ca configv1.Config
 
 func (c *authOperator) transportForCARef(ca configv1.ConfigMapNameReference, key string) (http.RoundTripper, error) {
 	if len(ca.Name) == 0 {
-		return transportFor("", nil, nil, nil)
+		return transportFor("", trustedCABytes(), nil, nil)
 	}
 	cm, err := c.configMaps.ConfigMaps(userConfigNamespace).Get(ca.Name, metav1.GetOptions{})
 	if err != nil {
@@ -342,4 +345,13 @@ func encodeOrDie(obj runtime.Object) []byte {
 		panic(err) // indicates static generated code is broken, unrecoverable
 	}
 	return bytes
+}
+
+func trustedCABytes() []byte {
+	caData, err := ioutil.ReadFile(operatorTrustedCAFile)
+	if err != nil {
+		klog.Infof("could not read %s, it won't be used in transport", operatorTrustedCAFile)
+		return nil
+	}
+	return caData
 }
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -64,6 +64,8 @@ const (
 
 	rootCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 
+	systemTrustStoreDirPath = "/etc/pki/ca-trust/extracted/pem"
+
 	systemConfigPath           = "/var/config/system"
 	systemConfigPathConfigMaps = systemConfigPath + "/configmaps"
 	systemConfigPathSecrets    = systemConfigPath + "/secrets"
@@ -108,6 +110,12 @@ const (
 	consoleConfigMapSharedName = "console-config"
 	consoleConfigMapLocalName  = systemConfigPrefix + consoleConfigMapSharedName
 	consoleConfigKey           = consoleConfigMapSharedName + ".yaml"
+
+	// trustedCABundleName part of manifests, if changing this, need to change that, too
+	trustedCABundleName      = systemConfigPrefix + "trusted-ca-bundle"
+	trustedCABundleKey       = "ca-bundle.crt"
+	trustedCABundleMountDir  = systemTrustStoreDirPath
+	trustedCABundleMountFile = "tls-ca-bundle.pem"
 
 	ocpBrandingSecretName   = systemConfigPrefix + "ocp-branding-template"
 	ocpBrandingSecretMount  = systemConfigPathSecrets + "/" + ocpBrandingSecretName
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -62,9 +62,8 @@ const (
 	kasServiceAndEndpointName = "kubernetes"
 	kasServiceFullName        = kasServiceAndEndpointName + "." + corev1.NamespaceDefault + ".svc"
 
-	rootCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-
-	systemTrustStoreDirPath = "/etc/pki/ca-trust/extracted/pem"
+	rootCAFile            = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	operatorTrustedCAFile = "/var/run/configmaps/trusted-ca-bundle/ca-bundle.crt"
 
 	systemConfigPath           = "/var/config/system"
 	systemConfigPathConfigMaps = systemConfigPath + "/configmaps"
@@ -114,7 +113,7 @@ const (
 	// trustedCABundleName part of manifests, if changing this, need to change that, too
 	trustedCABundleName      = systemConfigPrefix + "trusted-ca-bundle"
 	trustedCABundleKey       = "ca-bundle.crt"
-	trustedCABundleMountDir  = systemTrustStoreDirPath
+	trustedCABundleMountDir  = "/etc/pki/ca-trust/extracted/pem"
 	trustedCABundleMountFile = "tls-ca-bundle.pem"
 
 	ocpBrandingSecretName   = systemConfigPrefix + "ocp-branding-template"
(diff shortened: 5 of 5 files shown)


enj:
/hold

This looks weird to me.  Please link me to the requirements for this.

stlaz:
Adding WIP, this is needlessly complicated now, apparently the `trusted-ca-bundle` config map will always exist and should contain the CA bundle we need (if I get https://docs.google.com/document/d/1y0t0yEOSnKc4abxsjxEQjrFa1AP8iHcGyxlBpqGLO08/edit#heading=h.y6ieif41wmlc right), let's just use the sync logic we already have + getting the CM resourceVersion for deployment as added just lately and always mount the CM to the pod

stlaz:
This PR will now be failing until something starts creating the `trusted-ca-bundle` configmap in the openshift-config-managed namespace.

I suspect the main blocker is `https://github.com/openshift/cluster-network-operator/pull/274`, although the configmap creation may eventually be done elsewhere.

pkg/operator2/deployment.go
```diff
@@ -81,6 +81,24 @@ func defaultDeployment(
 	volumes, mounts = toVolumesAndMounts(syncData.idpSecrets, volumes, mounts)
 	volumes, mounts = toVolumesAndMounts(syncData.tplSecrets, volumes, mounts)
 
+	volumes = append(volumes, corev1.Volume{
```
stlaz: I should probably move the `key->path` mapping among the local `volume` struct methods to make this a bit less of a special snowflake, I'll keep the WIP status of the PR for that reason

stlaz:
Added the injection for the operator as well, similarly to https://github.com/openshift/cluster-image-registry-operator/pull/340 but without the `optional` mount. If the CM is supposed to be present, then why not fail if it isn't.

manifests/03_configmap.yaml
```diff
@@ -7,3 +7,14 @@ data:
   operator-config.yaml: |
     apiVersion: operator.openshift.io/v1alpha1
     kind: GenericOperatorConfig
+
```
enj: remove space

pkg/operator2/starter.go
```diff
@@ -122,6 +122,14 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		return err
 	}
 
+	// add syncing for the console-config ConfigMap (indirect watch for changes)
```
enj: Why is this not just another config map with `config.openshift.io/inject-trusted-cabundle` in `manifests`?

manifests/07_deployment.yaml
```diff
@@ -53,6 +56,12 @@ spec:
         configMap:
           defaultMode: 440
           name: authentication-operator-config
+      - name: trusted-ca-bundle
+        configMap:
+          name: trusted-ca
+          items:
+          - key: ca-bundle.crt
+            path: tls-ca-bundle.pem
```
enj: Why do we want a different path instead of using `ca-bundle.crt`?

manifests/07_deployment.yaml
```diff
@@ -36,6 +36,9 @@ spec:
           name: config
         - mountPath: /var/run/secrets/serving-cert
           name: serving-cert
+        - mountPath: /etc/pki/ca-trust/extracted/pem
```
enj: This will not get included in the root CAs that our go code is using.  So what is the point?

pkg/operator2/deployment.go
```diff
@@ -71,6 +71,14 @@ func defaultDeployment(
 			path:      ocpBrandingSecretMount,
 			keys:      []string{configv1.LoginTemplateKey, configv1.ProviderSelectionTemplateKey, configv1.ErrorsTemplateKey},
 		},
+		{
+			name:      trustedCABundleLocalName,
+			configmap: true,
+			path:      trustedCABundleMountDir,
+			mappedKeys: map[string]string{
```
enj: Why are we adding this complexity?  Keep `key` and `path` the same.

enj: Looking at the proxy doc, it seems to want this specific path for whatever reason - but I think the whole thing is pointless, see above.

pkg/operator2/deployment.go
```diff
@@ -276,6 +285,13 @@ func (v *volume) split() (corev1.Volume, corev1.VolumeMount) {
 		})
 	}
 
+	for key, val := range v.mappedKeys {
```
enj: Map iteration order is random.  You will cause hot loops when this flaps.

pkg/operator2/operator.go
```diff
@@ -62,7 +62,8 @@ const (
 	kasServiceAndEndpointName = "kubernetes"
 	kasServiceFullName        = kasServiceAndEndpointName + "." + corev1.NamespaceDefault + ".svc"
 
-	rootCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	systemTrustStoreDirPath = "/etc/pki/ca-trust/extracted/pem"
```
enj: No.  `rootCAFile` is for in-cluster communication.

And even if you are trying to say "use this single bundle for outgoing communication" you will quickly run into the issue that sometimes you need a real CA from the default root store and sometimes you need the `tls-ca-bundle.pem` thing.

pkg/operator2/starter.go
stlaz: I'll make it so

danehans:
@stlaz now that [1] and [2] have merged, operators/operands requiring proxy support should no longer be blocked from implementing this feature.
 
[1] https://github.com/openshift/cluster-network-operator/pull/271
[2] https://github.com/openshift/cluster-network-operator/pull/274

danehans:
> Added the injection for the operator as well, similarly to openshift/cluster-image-registry-operator#340 but without the optional mount. If the CM is supposed to be present, then why not fail if it isn't.

@stlaz @bparees has suggested to make the ca bundle mount optional.  He doesn't want an operator to fail to schedule, causing a domino effect of failures. Operator's can successfully start  without the CA trust bundle and then restart when the bundle is available.

manifests/07_deployment.yaml
danehans: @enj this was the default chosen for all operators as referenced in the [bootstrap workflow](https://docs.google.com/document/d/1y0t0yEOSnKc4abxsjxEQjrFa1AP8iHcGyxlBpqGLO08/edit?usp=sharing) doc.

pkg/operator2/operator.go
danehans: @enj see my [comment](https://github.com/openshift/cluster-authentication-operator/pull/172/files#r314894913), `tls-ca-bundle.pem` is a merged user-provided/system trust bundle.

stlaz: I'm wondering whether @enj's comment is maybe more about trust domain separation rather than about the bundle being a universal hammer for trust.

I'm not quite familiar with how CA bundles are loaded for client connections for in-cluster API communication, but I guess if this was an issue, it would be an issue for all the components that mount the bundle such as we do here.

manifests/07_deployment.yaml
stlaz: From what I gather this is one of the root cert stores that golang searches to get its system trust store.

We have pieces of code in oauth-server that seem to be using default HTTP transport if the user does not provide their own CA.

If they however do specify a CA in a proxied environment, without NO_PROXY adjustments, they are either not aware that proxying is in place, which might mean that someone else is trying to steal all their traffic, or they don't know what they are doing. I'll gladly have the code fail in both cases :man_shrugging:

stlaz: Pfft, the logic for getting CA data we're using seems a bit arbitrary (not saying wrong, though), I'll need to have a better look.

pkg/operator2/operator.go
stlaz: Nvm, this is indeed wrong, `rootCAFile` is only used for in-cluster communication and should not be changed.

manifests/07_deployment.yaml
stlaz: If I got it right, this would only be used in the case where `ca` field is not specified for an OpenID IdP, which is exactly the single one use where we want it. If `ca` is configured and `NO_PROXY` is not set, it should not be our problem.

stlaz:
@danehans I made the CM optional, then. I don't think we'll need to restart the pod, we're dynamically reloading our transports for given CAs, I just added the reload for the system-default trust store, too.

pkg/operator2/transport.go
```diff
@@ -22,21 +22,27 @@ func transportFor(serverName string, caData, certData, keyData []byte) (http.Rou
...
 	// copy default transport
 	transport := net.SetTransportDefaults(&http.Transport{
 		TLSClientConfig: &tls.Config{
 			ServerName: serverName,
 		},
 	})
 
+	if len(caData) == 0 && len(certData) == 0 && len(keyData) == 0 {
+		// reload system cert pool in case trusted-ca-bundle changed
+		systemCertPool, err := x509.SystemCertPool()
```
enj: `SystemCertPool` does not reload the system roots.  After the initial run the result is cached.

pkg/operator2/deployment.go
```diff
@@ -269,6 +278,20 @@ func (v *volume) split() (corev1.Volume, corev1.VolumeMount) {
 	}
 
 	var items []corev1.KeyToPath
+	for key, val := range v.mappedKeys {
+		items = append(items, corev1.KeyToPath{
+			Key:  key,
+			Path: val,
+		})
+	}
+	// maps' keys are random, need to sort the output to prevent redeployment hotloops
```
enj: Use `StringKeySet(v.mappedKey).List()` to save boilerplate

manifests/07_deployment.yaml
enj: Sure but there is no need to mount that config map at `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` in the authentication operator.  It is all custom Go code / the Go std lib does not look at that path for system roots.

pkg/operator2/deployment.go
enj: Though again, I see no point in mounting the proxy CA at that path.

manifests/07_deployment.yaml
enj: Your attempt to reload system roots will not work without a restart of the entire pod.

manifests/07_deployment.yaml
stlaz: I think Go stdlib actually might look at those files (see `crypto/x509/root_linux.go`) but loading from a specific file sounds like a better idea than relying on this specific defaulting logic, for now at least, since we can easily control all the connections we make outside. It's something to bear in mind, though.

pkg/operator2/transport.go
stlaz: removed that code, should have checked more thoroughly, thanks

pkg/operator2/idp.go
```diff
@@ -343,3 +344,12 @@ func encodeOrDie(obj runtime.Object) []byte {
 	}
 	return bytes
 }
+
+func transportForTrustedCA(serverName string, certData []byte, keyData []byte) (http.RoundTripper, error) {
+	caData, err := ioutil.ReadFile(operatorTrustedCAFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read %s: %v", operatorTrustedCAFile, err)
```
enj: Why are failing on an optional CA?

pkg/operator2/route.go
```diff
@@ -127,31 +126,6 @@ func routerSecretToSNI(routerSecret *corev1.Secret) []configv1.NamedCertificate
 	return out
 }
 
-func routerSecretToCA(route *routev1.Route, routerSecret *corev1.Secret, ingress *configv1.Ingress) []byte {
```
enj: Deleting this is wrong.  If you do not have a CA bundle in the optional CA, you still need this to trust the route.

pkg/operator2/operator.go
```diff
@@ -62,7 +62,10 @@ const (
 	kasServiceAndEndpointName = "kubernetes"
 	kasServiceFullName        = kasServiceAndEndpointName + "." + corev1.NamespaceDefault + ".svc"
 
-	rootCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	rootCAFile            = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+	operatorTrustedCAFile = "/var/run/configmaps/trusted-ca-bundle/ca-bundle.crt"
+
+	systemTrustStoreDirPath = "/etc/pki/ca-trust/extracted/pem"
```
enj: Remove this and use the same path as all of the system config.

pkg/operator2/operator.go
```diff
@@ -109,6 +112,12 @@ const (
 	consoleConfigMapLocalName  = systemConfigPrefix + consoleConfigMapSharedName
 	consoleConfigKey           = consoleConfigMapSharedName + ".yaml"
 
+	// trustedCABundleName part of manifests, if changing this, need to change that, too
+	trustedCABundleName      = systemConfigPrefix + "trusted-ca-bundle"
+	trustedCABundleKey       = "ca-bundle.crt"
+	trustedCABundleMountDir  = systemTrustStoreDirPath
```
enj: See other comment about making this match everything else.

pkg/operator2/operator.go
stlaz: I believe this is only for the operand where we do actually want to use this patn

pkg/operator2/route.go
stlaz: Would it make more sense not to make it optional, given it should always exist?

pkg/operator2/operator.go
enj: Hmm okay, add a comment about that.

pkg/operator2/operator.go
enj: Remove this const and use `systemTrustStoreDirPath` directly.

pkg/operator2/route.go
enj: Even if it exists, I do not think we can guarantee that it will work against the route, whereas the the current code will work as long as there is no TLS intercepting proxy.  Merging the optional CA bundle with output of `routerSecretToCA` is probably the correct approach.

pkg/operator2/idp.go
stlaz: Yeah, that's stupid, I've gone a bit different path in latest update

stlaz:
/test e2e-aws-upgrade
aws problem

enj:
Even with https://github.com/openshift/cluster-network-operator/pull/296 in mind, I think this is fine.

/hold cancel

manifests/03_configmap.yaml
```diff
@@ -7,3 +7,23 @@ data:
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-authentication-operator
+  name: trusted-ca-bundle
+  annotations:
+    release.openshift.io/create-only: "true"
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
```
sttts: won't CVO stomp this configmap after network operator added the CA ?

sttts: same below of course

pkg/operator2/idp.go
```diff
@@ -343,3 +346,12 @@ func encodeOrDie(obj runtime.Object) []byte {
 	}
 	return bytes
 }
+
+func trustedCABytes() []byte {
```
sttts: are these additional to the system trust store? The file only includes some CAs, not all. Especially not possibly required intermediate ones.

pkg/operator2/operator.go
```diff
@@ -109,6 +110,12 @@ const (
 	consoleConfigMapLocalName  = systemConfigPrefix + consoleConfigMapSharedName
 	consoleConfigKey           = consoleConfigMapSharedName + ".yaml"
 
+	// trustedCABundleName part of manifests, if changing this, need to change that, too
+	trustedCABundleName      = systemConfigPrefix + "trusted-ca-bundle"
+	trustedCABundleKey       = "ca-bundle.crt"
+	trustedCABundleMountDir  = "/etc/pki/ca-trust/extracted/pem"
+	trustedCABundleMountFile = "tls-ca-bundle.pem"
```
sttts: a plain yaml text file would me much more readable than 35 constansts.

pkg/operator2/operator.go
```diff
@@ -525,7 +532,8 @@ func (c *authOperator) checkDeploymentReady(deployment *appsv1.Deployment, opera
 func (c *authOperator) checkRouteHealthy(route *routev1.Route, routerSecret *corev1.Secret, ingress *configv1.Ingress) (ready bool, msg, reason string, err error) {
 	caData := routerSecretToCA(route, routerSecret, ingress)
 
-	rt, err := transportFor("", caData, nil, nil)
+	// merge trustedCA data with router cert in case TLS intercept proxy is in place
+	rt, err := transportFor("", append(caData, trustedCABytes()...), nil, nil)
```
sttts: this reads the file on every request, not good.

sttts: Are we sure these bytes always append cleanly, i.e. that we have a trailing newline in caData? Looks fragile.

manifests/03_configmap.yaml
sttts: @danehans ^^

pkg/operator2/operator.go
sttts: ic, we probably reread to get the latest version (which can change in the background).

manifests/03_configmap.yaml
squeed: presumably not, because of the `create-only` annotation...

sttts: nvmd. The create-only label will make sure it is not overwritten.

PR #173, opened 2019-08-12
Bug 1740121: Prevent global sync error degraded status flapping

enj:


enj:
I have confirmed that this stops the degraded status flapping by purposefully breaking service-ca.  The CO resource correctly went degraded and the `authentication.operator` resource went degraded once and stayed that way until service-ca was fixed (at which point it returned to normal).

sallyom:
/lgtm

took me a few minutes to follow what was happening in this PR : ), got it now

PR #176, opened 2019-08-16
Bug 1707905: Update route when ingress domain changes

mrogers950:
Pushing what I have initially while looking into adding tests.

### DIFF
--- a/pkg/operator2/route.go
+++ b/pkg/operator2/route.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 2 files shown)


pkg/operator2/route.go
```diff
@@ -15,7 +16,21 @@ import (
...
+			// we're only interested in the targetName. ingress
+			if !strings.HasPrefix(ig.Host, targetName) {
+				continue
+			}
+			if !strings.HasSuffix(ig.Host, "." + ingress.Spec.Domain) {
+				// the domain has changed, so update the route
+				updateRoute = true
+			}
+		}
+	}
```
sttts: this is dangerous style to have if clauses without else ligned up. If you fall through with a modified err you get some funny behaviour.

pkg/operator2/route.go
```diff
@@ -15,7 +16,21 @@ import (
 )
 
 func (c *authOperator) handleRoute(ingress *configv1.Ingress) (*routev1.Route, *corev1.Secret, error) {
+	var updateRoute bool
 	route, err := c.route.Get(targetName, metav1.GetOptions{})
+	if err == nil {
+		// route exists, check to see if the ingress domain has changed
+		for _, ig := range route.Status.Ingress {
+			// we're only interested in the targetName. ingress
+			if !strings.HasPrefix(ig.Host, targetName) {
```
sttts: should this be `targetName+ "."`?

pkg/operator2/route.go
```diff
@@ -15,7 +16,21 @@ import (
...
+				continue
+			}
+			if !strings.HasSuffix(ig.Host, "." + ingress.Spec.Domain) {
+				// the domain has changed, so update the route
```
sttts: why does the right pre- and postfix imply that the domain has changed?

sttts:
Giving some naive comments. Don't know the code around much.

mrogers950:
@enj I've updated with tests (there may be a few cases I should add). Please check and see if my changes are closer to what you described.

enj:
> @enj I've updated with tests (there may be a few cases I should add). Please check and see if my changes are closer to what you described.

Was easier for me to respond with PR: https://github.com/mrogers950/cluster-authentication-operator/pull/2

pkg/operator2/route.go
```diff
@@ -22,7 +22,14 @@ func (c *authOperator) handleRoute(ingress *configv1.Ingress) (*routev1.Route, *
 	if err != nil {
 		return nil, nil, "FailedCreate", err
 	}
+	defRoute := defaultRoute(ingress)
 
+	if route.Spec.Host != defRoute.Spec.Host {
+		route, err = c.route.Update(defaultRoute(ingress))
```
enj: Note that this approach to update will fail because resource version will be unset.

pkg/operator2/route.go
```diff
@@ -36,14 +43,13 @@ func (c *authOperator) handleRoute(ingress *configv1.Ingress) (*routev1.Route, *
...
 	if err := isValidRoute(route, ingress); err != nil {
-		// TODO remove this delete so that we do not lose the early creation timestamp of our route
-		// delete the route so that it is replaced with the proper one in next reconcile loop
-		klog.Infof("deleting invalid route: %#v", route)
-		opts := &metav1.DeleteOptions{Preconditions: &metav1.Preconditions{UID: &route.UID}}
-		if err := c.route.Delete(route.Name, opts); err != nil && !errors.IsNotFound(err) {
-			klog.Infof("failed to delete invalid route: %v", err)
+		klog.Infof("route invalid: %v, restoring", err)
+		routeUpdate := defaultRoute(ingress)
+		routeUpdate.ObjectMeta = route.ObjectMeta // Preserve metadata
```
enj: I went with partial preservel in https://github.com/mrogers950/cluster-authentication-operator/pull/2

PR #177, opened 2019-08-19
Bug 1740363: operator: improve the reason reporter for route

mfojtik:
There are 50+ degraded auth operators with reason being `RouteStatusDegradedError`. This PR provide more detailed reason string (as don't have access to the message yet). 

/cc @enj 
/cc @stlaz

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -311,8 +311,8 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 		return fmt.Errorf("failed getting the ingress config: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


stlaz:
/lgtm

I bet there would be a cleaner way to do this, but for now to get better data from support, this is ok.

enj:
/hold

You cannot just change reason - the string is an API.

enj:
> /hold
> 
> You cannot just change reason - the string is an API.

In the sense that the string change can cause issues across upgrades.  You have to manually remove the old value since it's written out to status.  Also this is just a silly workaround to not having proper logs from failing clusters.

enj:
#179

PR #179, opened 2019-08-19
Bug 1740363: Use specific reasons for all route degraded errors

enj:

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -311,8 +311,8 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 		return fmt.Errorf("failed getting the ingress config: %v", err)
 	}
 
-	route, routerSecret, err := c.handleRoute(ingress)
-	handleDegraded(operatorConfig, "RouteStatus", err)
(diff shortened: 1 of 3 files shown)


enj:
```
$ oc get co authentication -o json | jq .status.conditions[0]
```

```json
{
  "lastTransitionTime": "2019-08-19T19:53:57Z",
  "message": "RouteStatusDegraded: route is not available at canonical host oauth-openshift.apps.m094-1.devcluster.openshift.com: [{Host:oauth-openshift.apps.m094.devcluster.openshift.com RouterName:default Conditions:[{Type:Admitted Status:True Reason: Message: LastTransitionTime:2019-08-19 18:16:08 +0000 UTC}] WildcardPolicy:None RouterCanonicalHostname:apps.m094.devcluster.openshift.com}]",
  "reason": "RouteStatusDegradedFailedHost",
  "status": "True",
  "type": "Degraded"
}
```

Thus we can tell that the operator is degraded with`RouteStatusDegradedFailedHost` which tells us the [exact line](https://github.com/openshift/cluster-authentication-operator/pull/179/files#diff-0fe7789e2b4f1e11359fcc53e5beeba4R29) of the failure.

PR #182, opened 2019-08-21
route: use getCanonicalHost for route healthy check

mfojtik:
We should not use the `route.Spec` to get the hostname.


enj:
/hold

Read the code that gets the route.

enj:
Closing since #176 will clean up any issues this has.

PR #184, opened 2019-08-22
Bug 1740366: Add strict validation for router certs

enj:
The ingress operator performs no validation on the certs it is given
by the user.  Thus it can easily pass us nonsensical data.  We guard
against it and provide useful error messages.

This change also detects if the ingress operator ever wedges and
fails to keep the router-certs updated with the correct certs on an
ingress.config domain plus cert change.

### DIFF
--- a/pkg/operator2/route.go
+++ b/pkg/operator2/route.go
@@ -1,6 +1,7 @@
 package operator2
 
 import (
+	"bytes"
 	"crypto/x509"
 	"fmt"
 
@@ -12,6 +13,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
+	"github.com/openshift/library-go/pkg/crypto"
 )
 
 func (c *authOperator) handleRoute(ingress *configv1.Ingress) (*routev1.Route, *corev1.Secret, string, error) {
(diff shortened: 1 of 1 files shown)


pkg/operator2/route.go
```diff
@@ -178,3 +179,74 @@ func isIngressAdmitted(ingress routev1.RouteIngress) bool {
...
+		return "InvalidPEMRouterSecret", fmt.Errorf("router secret contains invalid PEM data: %v", err)
+	}
+
+	opts := x509.VerifyOptions{
+		DNSName:       ingressToHost(ingress),
+		Intermediates: x509.NewCertPool(),
+		Roots:         x509.NewCertPool(),
+	}
+
+	var hasCA, hasCert bool
```
stlaz: `hasRoot`, `hasServer` better describe what's going on

pkg/operator2/route.go
```diff
@@ -178,3 +179,74 @@ func isIngressAdmitted(ingress routev1.RouteIngress) bool {
...
+			klog.V(4).Infof("cert %s failed verification: %v", certificate.Subject.String(), err)
+			continue
+		}
+
+		hasCert = true
+		break
+	}
+
+	if !hasCA {
+		return "NoCARouterSecret", fmt.Errorf("router secret contains no CA")
```
stlaz: "NoRootCARouterSecret", "contains no root CA" -> it might be containing CAs, just not roots

Wouldn't intermediates be enough for our use-case?

pkg/operator2/route.go
```diff
@@ -178,3 +179,74 @@ func isIngressAdmitted(ingress routev1.RouteIngress) bool {
...
+	}
+
+	if !hasCert {
+		return "NoCertRouterSecret", fmt.Errorf("router secret has no cert for ingress domain %s", domain)
```
stlaz: "NoServerCertRouterSecret"

enj:
Comments addressed.  Extra logging adding.  Made slight changes to how root CAs are handled to make sure we work with Let's Encrypt.

PR #187, opened 2019-08-29
Wire trusted-ca-bundle via /etc/pki

sttts:
Follow-up of https://github.com/openshift/cluster-authentication-operator/pull/172 to make it more standard.


manifests/07_deployment.yaml
```diff
@@ -23,10 +23,14 @@ spec:
       - name: operator
         image: quay.io/openshift/origin-cluster-authentication-operator:v4.0
         imagePullPolicy: IfNotPresent
-        command: ["authentication-operator", "operator"]
+        command: ["/bin/bash", "ec"]
```
mfojtik: -ec ?

manifests/07_deployment.yaml
```diff
@@ -23,10 +23,14 @@ spec:
...
         args:
-        - "--config=/var/run/configmaps/config/operator-config.yaml"
-        - "-v=2"
+        - |
+          if [ -s /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem ]; then
```
mfojtik: I wish the operator does this during startup and we don't have to hack ways in bash.

sttts: it cannot. Then it is too late. And this does not belong into a binary. It is system plumbing. It is just right here.

mfojtik: if `tls-ca-bundle.pem` is empty, will that brick the operator?

sttts: it is non-existent if not part of the configmap.

sttts: The network operator validates it.

enj:
After fixing the comment thing and making sure the bump size is expected, feel free to have anyone tag.

pkg/operator2/operator.go
```diff
@@ -113,7 +114,7 @@ const (
 	// trustedCABundleName part of manifests, if changing this, need to change that, too
 	trustedCABundleName      = systemConfigPrefix + "trusted-ca-bundle"
 	trustedCABundleKey       = "ca-bundle.crt"
-	trustedCABundleMountDir  = "/etc/pki/ca-trust/extracted/pem"
+	trustedCABundleMountDir  = systemTrustStoreDirPath
```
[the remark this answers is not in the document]
sttts: this is a revert. Would avoid to change anything not neccesary.

sttts:
> nit. Why is the glide bump so large?

because glide.

manifests/07_deployment.yaml
sttts: @stlaz noticed that we must not copy empty files if I have understood him correctly.

pkg/operator2/operator.go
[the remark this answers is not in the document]
sttts: I reverted the changes in that file, yes.

enj:
> I reverted the changes in that file, yes.

Not sure I understand but do not feel strongly to block anything.

/lgtm

pkg/operator2/operator.go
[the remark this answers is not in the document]
sttts: Have removed the constant now.

PR #194, opened 2019-09-13
Bug 1751147: operator: restart the operator pod on trusted-ca-bundle change

stlaz:
Introduces the watchdog for operator deployment so that there is
a mechanism which guards changes to the trusted-ca-bundle CM and
these are refetched by having the operator pod restart.

cc @enj @sttts

### DIFF
--- a/manifests/07_deployment.yaml
+++ b/manifests/07_deployment.yaml
@@ -18,6 +18,7 @@ spec:
       labels:
         app: authentication-operator
     spec:
+      shareProcessNamespace: true
       serviceAccountName: authentication-operator
       containers:
       - name: operator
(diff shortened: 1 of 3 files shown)


stlaz:
/hold
untested

pkg/operator2/operator.go
```diff
@@ -530,7 +530,9 @@ func (c *authOperator) checkDeploymentReady(deployment *appsv1.Deployment, opera
 func (c *authOperator) checkRouteHealthy(route *routev1.Route, routerSecret *corev1.Secret, ingress *configv1.Ingress) (ready bool, msg, reason string, err error) {
 	caData := routerSecretToCA(route, routerSecret, ingress)
 
-	rt, err := transportFor("", caData, nil, nil)
+	// FIXME: this reads too often, either always merge system-trust store in transportForInner or keep this in memory
+	systemCaData, _ := ioutil.ReadFile("/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem")
```
enj: Read the file once in an `init` block and store in `var`.  You will get restarted anyway when it changes.

enj: Also this does assume a newline at the end of the first byte slice.

manifests/07_deployment.yaml
```diff
@@ -55,6 +65,39 @@ spec:
...
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+          - mountPath: /var/run/configmaps/trusted-ca-bundle/
+            name: trusted-ca-bundle
+          - mountPath: /var/run/watchdog
+            name: watchdog-dir
+        terminationMessagePolicy: FallbackToLogsOnError
+        resources:
+          requests:
+            memory: 50Mi
```
enj: Seems high

stlaz:
It seems that the watchdog is actually wired inside the code already which I did not know.

Superseeded by https://github.com/openshift/cluster-authentication-operator/pull/195

PR #196, opened 2019-09-17
Inline constants so David does not die

enj:

@mfojtik @sttts @deads2k


sttts:
I support that @deads2k does not die.

/lgtm
/hold

Waiting for @stlaz to take a look.

mfojtik:
/hold cancel

Good start, I also don't want any harm to David

PR #198, opened 2019-09-18
manifests: split configmaps

stlaz:
It seems that during upgrade, CVO might be applying whole files,
not just objects, which would cause it to remove the contents of
trusted-ca-bundle configmap in loop with CNO.


stlaz:
/test e2e-aws-operator
installer failure

PR #200, opened 2019-09-20
Bug 1753886: Don't stomp on our own route

stlaz:
It seems that new fields with non-nil default values were added to
routes, specifically .spec.to.weigh and .spec.wildcardPolicy.
Since we do not expect them in the route, it gets stomped on
all of the time.

-----

We may want to find a better way to ensure the substantial fields are configured the way we need them.

cc @mfojtik @enj

### DIFF
--- a/pkg/operator2/route.go
+++ b/pkg/operator2/route.go
@@ -58,14 +58,19 @@ func (c *authOperator) handleRoute(ingress *configv1.Ingress) (*routev1.Route, *
 }
 
 func defaultRoute(ingress *configv1.Ingress) *routev1.Route {
+	// emulates server-side defaulting as in https://github.com/openshift/openshift-apiserver/blob/master/pkg/route/apis/route/v1/defaults.go
+	// TODO: replace with server-side apply
+	var weightVal int32 = 100
+
 	return &routev1.Route{
 		ObjectMeta: defaultMeta(),
 		Spec: routev1.RouteSpec{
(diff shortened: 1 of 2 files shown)


pkg/operator2/route.go
```diff
@@ -58,14 +58,17 @@ func (c *authOperator) handleRoute(ingress *configv1.Ingress) (*routev1.Route, *
 }
 
 func defaultRoute(ingress *configv1.Ingress) *routev1.Route {
+	var weightVal int32 = 100
```
enj: Add a comment noting that this is emulating server side defaulting with a link to https://github.com/openshift/openshift-apiserver/blob/a48daecd8d8f22de33be3c7ea8127672f84d46e0/pkg/route/apis/route/v1/defaults.go

Add a TODO to replace this with server side apply.  We have the same problem with services IIRC.

pkg/operator2/route.go
```diff
@@ -74,6 +77,7 @@ func defaultRoute(ingress *configv1.Ingress) *routev1.Route {
 				Termination:                   routev1.TLSTerminationPassthrough,
 				InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
 			},
+			WildcardPolicy: routev1.WildcardPolicyNone,
```
enj: Add comment here too.

stlaz:
added comments, fixed the unit test

PR #201, opened 2019-09-25
Update deps to 1.16

marun:


glide.yaml
```diff
@@ -1,29 +1,29 @@
...
 - package: k8s.io/client-go
-  version: kubernetes-1.14.0
+  version: kubernetes-1.16.0
 - package: k8s.io/kube-aggregator
-  version: kubernetes-1.14.0
+  version: kubernetes-1.16.0
+- package: k8s.io/code-generator
+  version: kubernetes-1.16.0
 - package: k8s.io/apiextensions-apiserver
-  repo: https://github.com/openshift/kubernetes-apiextensions-apiserver
```
stlaz: Can we safely turn this into the regular apiextensions-apiserver now?

glide.yaml
```diff
@@ -1,29 +1,29 @@
...
+# Pin kube deps to the versions vendored by the target release
 - package: k8s.io/kube-openapi
-  version: b3a7cee44a305be0a69e1b9ac03018307287e1b0
+  version: 743ec37842bffe49dd4221d9026f30fb1d5adbc4
 - package: sigs.k8s.io/structured-merge-diff
-  version: e85c7b244fd2cc57bb829d73a061f93a441e63ce
-- package: k8s.io/code-generator
-  version: kubernetes-1.14.0
-
-  # pinned to a level that kubernetes uses.
```
stlaz: Can we unpin that, then?

glide.yaml
marun: My assumption was that the upstream carries for 1.14 are included in 1.16, and the 1.16 branch of `https://github.com/openshift/kubernetes-apiextensions-apiserver` appeared to  have no carries.

glide.yaml
marun: This change does not unpin, it changes the pin to the version used by 1.16.  The comment on line 20 applies to the packages that follow, including this one.

glide.yaml
stlaz: sounds good, then

glide.yaml
stlaz: Ah, I see that now, it's a tag, not a branch, thanks

stlaz:
/retest
the test clusters seem to be quite broken

PR #206, opened 2019-10-17
 grab minTLSVersion and cipherSuites from APIServer config 

stlaz:
/assign @sanchezl


stlaz:
This won't build until https://github.com/openshift/library-go/pull/559 merges

PR #207, opened 2019-10-20
Revert "remove degrade condition"

wking:
This reverts commit 17a0064a2c10bbf1e2b5926f21192ed5153cb5e2, #159.

4.2 has forked off, so the master branch is going to be 4.3, which is when the the comment said we could drop this.


marun:
@wking Are you interested in picking this PR back up?

stlaz:
/close
this has gotten stale (probably my fault, sorry)

There's a bunch of other conditions that appear in that controller now, let's remove them all at once at some point in the future

PR #208, opened 2019-10-20
pkg/operator2/operator_test: Add unit tests for handleVersion

wking:
I'm thinking about refactoring handleVersion to distinguish prereq stuff (the ingress route, etc.) from operand stuff (deployment), and unit tests will guard against accidental changes ;).

For folks like me who are unfamiliar with library-go's operator status handling, the prefixed conditions will eventually get collated into conditions with unprefixed names [here][1].

It's still not clear to me what, if anything, clears out stale conditions that don't get explicitly clobbered.  For example, say round 1 ended up like the test with:

    route: happy; OAuth: happy; deployment: 2 current, available replicas

which sets, among other conditions, `OAuthClientsDegraded=False`.  Then route health died and round 2 ended up like the test with:

    route: 500; OAuth: happy; deployment: 2 current, available replicas

which does nothing with `OAuthClientsDegraded`.  Would we still be setting `OAuthClientsDegraded=False` after round 2 despite calling [`setProgressingTrueAndAvailableFalse(operatorConfig, "RouteNotReady", routeMsg)` and bailing][2] before actually looking at the well-known endpoints in round 2?

[The `authConfig.Spec.Type != configv1.AuthenticationTypeIntegratedOAuth` condition][3] bypasses most of the `checkWellknownEndpointsReady` logic, but because that is attempting to load [a hard-coded `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`][4] from the local host, we'd have to make changes to get it reliably testing.

Run with:

```console
$ go test ./pkg/operator2/
```

Or with `make test-unit`.

[1]: https://github.com/openshift/library-go/blob/61ddf5cd7d9cba7ca5c548d5f1b9277f55f887db/pkg/operator/status/condition.go#L14-L78
[2]: https://github.com/openshift/cluster-authentication-operator/blob/3c3b1ed8b5e3de8efa516ab3a9368d75e87ec17c/pkg/operator2/operator.go#L381-L382
[3]: https://github.com/openshift/cluster-authentication-operator/blob/3c3b1ed8b5e3de8efa516ab3a9368d75e87ec17c/pkg/operator2/operator.go#L477-L480
[4]: https://github.com/openshift/cluster-authentication-operator/blob/3c3b1ed8b5e3de8efa516ab3a9368d75e87ec17c/pkg/operator2/operator.go#L483

### DIFF
--- a/pkg/operator2/deployment.go
+++ b/pkg/operator2/deployment.go
@@ -25,6 +25,7 @@ func defaultDeployment(
 	routerSecret *corev1.Secret,
 	proxyConfig *configv1.Proxy,
 	operatorDeployment *appsv1.Deployment,
+	image string,
 	resourceVersions ...string,
 ) *appsv1.Deployment {
 	replicas := int32(2) // TODO configurable?
@@ -115,7 +116,7 @@ func defaultDeployment(
 					ServiceAccountName: "oauth-openshift",
 					Containers: []corev1.Container{
 						{
-							Image:           oauthserverImage,
+							Image:           image,
 							ImagePullPolicy: getImagePullPolicy(operatorDeployment),
 							Name:            "oauth-openshift",
 							Command:         []string{"/bin/bash", "-ec"},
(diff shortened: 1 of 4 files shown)


stlaz:
/close
the generic workload controllers logic is now used to handle version bumps

PR #211, opened 2019-11-01
ingressstate: add ingress state controller

mfojtik:
The purpose of this controller is to perform health checks against endpoints and report the state of the pods serving those endpoints. This controller should report degraded conditions if:

1) Endpoint does not have any ready addresses
2) Endpoint has addresses that are not reachable (insecure for now)
3) Endpoint has addresses that target pods which are either terminated or have restartCount higher than 1

I think with this we can pretty much rule out oauth-server pods having problems and narrow the possible causes for RouteDegraded in telemetry to either ingress or network.

TODO: Add event checker for CNI problems.

/cc @deads2k 
/cc @stlaz

### DIFF
--- a/pkg/controller/ingressstate/ingress_state_controller.go
+++ b/pkg/controller/ingressstate/ingress_state_controller.go
@@ -1,0 +1,247 @@
+package ingressstate
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const ingressStateControllerWorkQueueKey = "key"
+
+type IngressStateController struct {
+	endpointsGetter corev1client.EndpointsGetter
+	podGetter       corev1client.PodsGetter
+	queue           workqueue.RateLimitingInterface
+	cachesToSync    []cache.InformerSynced
+	targetNamespace string
+	operatorClient  v1helpers.OperatorClient
+	eventRecorder   events.Recorder
+}
+
+func NewIngressStateController(kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	endpointsGetter corev1client.EndpointsGetter,
+	podsGetter corev1client.PodsGetter,
+	operatorClient v1helpers.OperatorClient,
+	targetNamespace string,
+	recorder events.Recorder,
+) *IngressStateController {
+	c := &IngressStateController{
+		endpointsGetter: endpointsGetter,
+		podGetter:       podsGetter,
+		targetNamespace: targetNamespace,
+		operatorClient:  operatorClient,
+		eventRecorder:   recorder.WithComponentSuffix("ingress-state-controller"),
+		queue:           workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "IngressStateController"),
+	}
+
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForTargetNamespace.Core().V1().Pods().Informer().HasSynced)
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForTargetNamespace.Core().V1().Endpoints().Informer().HasSynced)
+
+	kubeInformersForTargetNamespace.Core().V1().Pods().Informer().AddEventHandler(c.eventHandler())
+	kubeInformersForTargetNamespace.Core().V1().Endpoints().Informer().AddEventHandler(c.eventHandler())
+
+	return c
+}
+
+func (c *IngressStateController) eventHandler() cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.queue.Add(ingressStateControllerWorkQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(ingressStateControllerWorkQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(ingressStateControllerWorkQueueKey) },
+	}
+}
+
+// checkPodStatus will check the target pod container status and return list of possible problems.
+func (c *IngressStateController) checkPodStatus(reference *corev1.ObjectReference) []string {
+	pod, err := c.podGetter.Pods(reference.Namespace).Get(reference.Name, metav1.GetOptions{})
+	if err != nil {
+		return []string{fmt.Sprintf("error getting pod %q: %v", reference.Name, err)}
+	}
+	result := []string{}
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.Ready && containerStatus.State.Running != nil {
+			continue
+		}
+		if terminated := containerStatus.State.Terminated; terminated != nil {
+			result = append(result, fmt.Sprintf("pod %q container %q terminated with %q", pod.Name, containerStatus.Name, terminated.Message))
+		}
+		if containerStatus.RestartCount > 1 {
+			result = append(result, fmt.Sprintf("pod %q container %q restarted %d times", pod.Name, containerStatus.Name, containerStatus.RestartCount))
+		}
+		// TODO: Add more here, like we can detect "pending" pods
+	}
+	return result
+}
+
+// checkEndpointHealthz will check the health of given https://endpointIP:6443/healthz
+// TODO: this is insecure for now, but enough to tell us if we can get through
+func (c *IngressStateController) checkEndpointHealthz(endpointIP string) error {
+	// TODO: consider making this secure, however it will mean wiring the trust which must match the same trust as ingress controller
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	response, err := http.Get(fmt.Sprintf("https://%s:6443/healthz", endpointIP))
+	if err != nil {
+		reportedError := err
+		if response != nil {
+			reportedError = fmt.Errorf("status:%q, body: %q, error: %v", response.Status, response.Body, reportedError)
+		}
+		return reportedError
+	}
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("status:%q, body: %q", response.Status, response.Body)
+	}
+	return nil
+}
+
+// degradedConditionNames lists all conditions that this controller manage
+var degradedConditionNames = []string{
+	"IngressStateEndpointsDegraded",
+	"IngressStatePodsDegraded",
+}
+
+func (c *IngressStateController) sync() error {
+	foundConditions := []operatorv1.OperatorCondition{}
+
+	endpoint, err := c.endpointsGetter.Endpoints(c.targetNamespace).Get("oauth-openshift", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if len(endpoint.Subsets) == 0 {
+		foundConditions = append(foundConditions, operatorv1.OperatorCondition{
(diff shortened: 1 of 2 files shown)


pkg/controller/routestate/route_state_controller.go
```diff
@@ -0,0 +1,243 @@
...
+		// TODO: Add more here, like we can detect "pending" pods
+	}
+	return result
+}
+
+// checkEndpointHealthz will check the health of given https://endpointIP:6443/healthz
+// TODO: this is insecure for now, but enough to tell us if we can get through
+func (c *RouteStateController) checkEndpointHealthz(endpointIP string) (bool, error) {
+	// TODO: consider making this secure, however it will mean wiring the trust which must match the same trust as ingress controller
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
stlaz: If you grab a serving-CA cert from somewhere, you get additional value of checking that it is valid, although I'm not sure there's any component that tries to reach the OAuth server on the service plane

mfojtik: @stlaz i would need the same service-ca as ingress/router is using to be 100% sure, i thought it can be a nice follow up :-)

stlaz: not entirely true, we're using SNI, so the serving cert for route is different than on service plane

stlaz: but yeah, could work as a follow-up

pkg/controller/routestate/route_state_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	endpointsGetter corev1client.EndpointsGetter
+	podGetter       corev1client.PodsGetter
+	queue           workqueue.RateLimitingInterface
+	cachesToSync    []cache.InformerSynced
+	targetNamespace string
+	operatorClient  v1helpers.OperatorClient
+	eventRecorder   events.Recorder
+}
+
+func NewRouteStateController(kubeInformersForTargetNamespace informers.SharedInformerFactory,
```
stlaz: We're checking service endpoints here, isn't this really but a NetworkStateController?

pkg/controller/routestate/route_state_controller.go
```diff
@@ -0,0 +1,246 @@
...
+			result = append(result, fmt.Sprintf("pod %q container %q restarted %d times", pod.Name, containerStatus.Name, containerStatus.RestartCount))
+		}
+		// TODO: Add more here, like we can detect "pending" pods
+	}
+	return result
+}
+
+// checkEndpointHealthz will check the health of given https://endpointIP:6443/healthz
+// TODO: this is insecure for now, but enough to tell us if we can get through
+func (c *RouteStateController) checkEndpointHealthz(endpointIP string) (bool, error) {
```
stlaz: Only return error, you never return `false, nil` or `true, err` in this function

pkg/controller/routestate/route_state_controller.go
```diff
@@ -0,0 +1,246 @@
...
+			Reason:  "ZeroEndpointsSubsets",
+			Message: "No endpoints found for oauth-server",
+		})
+	}
+
+	for _, subset := range endpoint.Subsets {
+		// If endpoint does not have any ready addresses, only not-ready, report it as RouteStateEndpointsDegraded
+		if len(subset.Addresses) == 0 && len(subset.NotReadyAddresses) > 0 {
+			foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+				Type:    "RouteStateEndpointsDegraded",
```
stlaz: Won't this report `Degraded` during install?

pkg/controller/routestate/route_state_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		if containerStatus.Ready && containerStatus.State.Running != nil {
+			continue
+		}
+		if terminated := containerStatus.State.Terminated; terminated != nil {
+			result = append(result, fmt.Sprintf("pod %q container %q terminated with %q", pod.Name, containerStatus.Name, terminated.Message))
+		}
+		if containerStatus.RestartCount > 1 {
+			result = append(result, fmt.Sprintf("pod %q container %q restarted %d times", pod.Name, containerStatus.Name, containerStatus.RestartCount))
+		}
+		// TODO: Add more here, like we can detect "pending" pods
```
stlaz: about about `Error` and `CrashLoopBackOff` pods?

pkg/controller/routestate/route_state_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		return []string{fmt.Sprintf("error getting pod %q: %v", reference.Name, err)}
+	}
+	result := []string{}
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.Ready && containerStatus.State.Running != nil {
+			continue
+		}
+		if terminated := containerStatus.State.Terminated; terminated != nil {
+			result = append(result, fmt.Sprintf("pod %q container %q terminated with %q", pod.Name, containerStatus.Name, terminated.Message))
```
stlaz: Won't this report degraded on configuration change (=> new pods rollout)

pkg/controller/routestate/route_state_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	if response.StatusCode != http.StatusOK {
+		return false, fmt.Errorf("status:%q, body: %q", response.Status, response.Body)
+	}
+	return true, nil
+}
+
+// degradedConditionNames lists all conditions that this controller manage
+var degradedConditionNames = []string{
+	"RouteStateEndpointsDegraded",
```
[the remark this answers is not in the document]
mfojtik: this is re-using the mechanism ingress/router use, so if something is broken here the route won't work.

pkg/controller/routestate/route_state_controller.go
mfojtik: sure, but it should be just minor blip.

pkg/controller/routestate/route_state_controller.go
mfojtik: yeah, hence the TODO: add more as we need :-)

pkg/controller/routestate/route_state_controller.go
mfojtik: @stlaz maybe, we can check the e2e run

mfojtik: @stlaz i wonder if we can tweak this to "all oauth pods are broken" instead of "at least one pod is broken"

pkg/controller/routestate/route_state_controller.go
mfojtik: no, but I think IngressStateController might be more accurate, renamed.

pkg/controller/routestate/route_state_controller.go
mfojtik: ok, I changed the logic and will require at least 2 failed pods to set degraded.

mfojtik:
/hold

I want to spend some time testing this

pkg/controller/routestate/route_state_controller.go
stlaz: I think there used to be a test that was checking the presence of degraded during installation. Would you be fine moving this to Progressing instead?

pkg/controller/routestate/route_state_controller.go
stlaz: :+1:

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,247 @@
...
+		if containerStatus.Ready && containerStatus.State.Running != nil {
+			continue
+		}
+		if terminated := containerStatus.State.Terminated; terminated != nil {
+			result = append(result, fmt.Sprintf("pod %q container %q terminated with %q", pod.Name, containerStatus.Name, terminated.Message))
+		}
+		if containerStatus.RestartCount > 1 {
+			result = append(result, fmt.Sprintf("pod %q container %q restarted %d times", pod.Name, containerStatus.Name, containerStatus.RestartCount))
+		}
+		// TODO: Add more here, like we can detect "pending" pods
```
mfojtik: there is code in library-go that handle detection and reporting of "pending" pods... we use that for static pod installer.

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,247 @@
...
+var degradedConditionNames = []string{
+	"IngressStateEndpointsDegraded",
+	"IngressStatePodsDegraded",
+}
+
+func (c *IngressStateController) sync() error {
+	foundConditions := []operatorv1.OperatorCondition{}
+
+	endpoint, err := c.endpointsGetter.Endpoints(c.targetNamespace).Get("oauth-openshift", metav1.GetOptions{})
+	if err != nil {
```
mfojtik: can we check for 404 and report it as "no endpoints" ?

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	for _, subset := range endpoint.Subsets {
+		// If endpoint does not have any ready addresses, only not-ready, report it as IngressStateEndpointsDegraded
+		if len(subset.Addresses) == 0 && len(subset.NotReadyAddresses) > 0 {
+			foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+				Type:    "IngressStateEndpointsDegraded",
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "NonReadyEndpoints",
+				Message: "No ready endpoints found for oauth-server",
```
mfojtik: can you include `subset.NotReadyAddresses` in message?

pkg/controller/ingressstate/ingress_state_controller.go
mfojtik: you don't have to use it here, eventually I need to make that library-go code generic so we can use it here... follow up.

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		// TODO: Add more here, like we can detect "pending" pods
+	}
+	return result
+}
+
+// checkEndpointHealthz will check the health of given https://endpointIP:6443/healthz
+// TODO: this is insecure for now, but enough to tell us if we can get through
+func (c *IngressStateController) checkEndpointHealthz(endpointIP string) error {
+	// TODO: consider making this secure, however it will mean wiring the trust which must match the same trust as ingress controller
```
mfojtik: @stlaz might have opinions on this

stlaz: Like I said before, the endpoint should be served using the service-ca-signed server cert, you could probably even use the CA from `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`

marun:
I will be forking and resubmitting this change in a new PR.

pkg/controller/ingressstate/ingress_state_controller.go
marun: I copied the approach in library-go, WDYT?

https://github.com/openshift/cluster-authentication-operator/pull/237/files#diff-ba3fd38aabbab497940396c7f19e82f9R191

pkg/controller/ingressstate/ingress_state_controller.go
marun: afaik the service ca needs to be used since the endpoints use a serving cert.

pkg/controller/ingressstate/ingress_state_controller.go
marun: Done

pkg/controller/ingressstate/ingress_state_controller.go
marun: Done

pkg/controller/ingressstate/ingress_state_controller.go
stlaz: service-ca should be a part of the above bundle AFAIK, either should work, but I am also inclined to using the service-ca

marun: What mechanism is ensuring that the service ca bundle is included in the bundle provided with service accounts? It doesn't appear that the kao is requesting injection of the service ca bundle, so I'm not sure how the composition would work.

Also, if the service account bundle could be used to verify serving certs, why would the service ca operator need to support configmap bundle injection? Any pod that wanted to verify serving cert connections could just use the ca bundle supplied with its service account. If this is not supported today, maybe it should be?

marun: @stlaz You were on the right track - the service ca bundle is provided verbatim with every service account token as `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`. I'm not sure why this shouldn't be used instead of configmap injection.

marun:
This PR should be closed as it was replaced by #237

stlaz:
/close
#237 merged

PR #213, opened 2019-11-01
Add test to check reporting not enough replicas

stlaz:
Currently fails because it can't find `authentications.operator.openshift.io "cluster" not found` even though it exists?

### DIFF
--- a/test/e2e/conditions_test.go
+++ b/test/e2e/conditions_test.go
@@ -1,0 +1,104 @@
+package e2e
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


test/e2e/conditions_test.go
```diff
@@ -0,0 +1,99 @@
...
+		return compareConditions(expectedCondition, actualCondition), nil
+	})
+}
+
+func compareConditions(c1, c2 *operatorv1.OperatorCondition) bool {
+	if c1 == c2 { // both are nil or the same pointer
+		return true
+	}
+
+	return c1.Type == c2.Type && c1.Status == c2.Status &&
```
vareti: do we also need to consider the case where only one of `c1` or `c2` is nil?

stlaz: yes, should be fixed now

marun:
@stlaz Is an e2e necessary, or would a unit test of `checkDeploymentReady` suffice?

stlaz:
I think testing conditions makes more sense in what we could call the real world

marun:
> I think testing conditions makes more sense in what we could call the real world

The only thing this e2e is validating that a unit test wouldn't is retrieval of a deployment. I'm not exactly sure why that is something valuable to test given that failure to retrieve the deployment will result in a degraded condition that would be caught without this e2e.

stlaz:
/close
this got stale, if we need it, we could perhaps do it better as Maru suggested

PR #218, opened 2019-11-19
Bug 1671604: redeploy oauth server deployment if bootstrap user is disabled

vareti:
As the title mentions, this PR aims to resolve 1671604.

When the bootstrap user is deleted, cluster authentication operator detects the change and redeploys the `oauth-server`. `oauth-server` will not include bootstrap identity provider in the slice of authorization handlers during start up, there by disabling password based logins using `oc` when no other password authenticators are configured. These changes are divided in PRs across multiple repositories.

https://github.com/openshift/oauth-server/pull/23
https://github.com/openshift/library-go/pull/594


stlaz:
there's no commit message or bug associated to this PR - why would you want to redeploy on the kubeadmin secret removal?

stlaz:
I see there's multiple PRs opened related to this but there's no visibility between them

pkg/operator2/operator.go
```diff
@@ -330,12 +333,19 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 		operatorDeployment,
 		resourceVersions...,
 	)
+
+	// redeploy on operatorConfig.spec changes or when bootstrap user is deleted
+	forceRollOut := (operatorConfig.Generation != operatorConfig.Status.ObservedGeneration)
+	if !c.bootstrapUserExists() && c.bootstrapUserChangeRollOut {
+		forceRollOut = true
+		c.bootstrapUserChangeRollOut = !c.bootstrapUserChangeRollOut
```
stlaz: this will cause an unnecessary rollout every time the operator is started anew after the bootstrap user has been removed

pkg/operator2/secret.go
```diff
@@ -6,7 +6,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-
```
stlaz: keep it

vareti: restored it.

pkg/operator2/operator.go
vareti: That is a good point which I failed to think when implementing this.

Now I thought about it a little bit and this is what I could come up with. Instead of blindly setting it to true, `bootstrapUserChangeRollOut` is set conditionally, thereby avoiding the forced roll outs.

pkg/operator2/secret.go
```diff
@@ -27,6 +28,17 @@ func (c *authOperator) expectedSessionSecret() (*corev1.Secret, error) {
 	return secret, nil
 }
 
+func (c *authOperator) bootstrapUserExists() bool {
+	b := bootstrapauthenticator.NewBootstrapUserDataGetter(c.secrets, c.namespaces)
+
+	enabled, err := b.IsEnabled()
```
vareti: As this is a library call, I took the liberty of re-using it, rather can writing the same logic here.

pkg/operator2/secret.go
```diff
@@ -27,6 +27,14 @@ func (c *authOperator) expectedSessionSecret() (*corev1.Secret, error) {
 	return secret, nil
 }
 
+func (c *authOperator) bootstrapUserExists() (bool, error) {
```
stlaz: The contents of this function now directly map to the conditions above, so everything is done twice - you can just remove this

pkg/operator2/operator.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/library-go/pkg/authentication/bootstrapauthenticator"
```
stlaz: move among other openshift imports

pkg/operator2/operator.go
```diff
@@ -330,12 +344,23 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
...
 	)
+
+	// redeploy on operatorConfig.spec changes or when bootstrap user is deleted
+	forceRollOut := operatorConfig.Generation != operatorConfig.Status.ObservedGeneration
+	if c.bootstrapUserChangeRollOut {
+		if userExists, err := c.bootstrapUserExists(); err != nil {
+			klog.Warningf("Unable to determine the state of bootstrap user: %v", err)
+		} else if !userExists {
+			forceRollOut = true
+			c.bootstrapUserChangeRollOut = !c.bootstrapUserChangeRollOut
```
stlaz: This can never change from `false` to `true` (given the wrapping condition), so you can safely do `c.bootstrapUserChangeRollOut = false`

pkg/operator2/secret.go
vareti: Yes. I removed this..

pkg/operator2/operator.go
vareti: Done!

pkg/operator2/operator.go
vareti: Not sure why `go fmt` did not work previously. Fixed it now.

pkg/operator2/operator.go
```diff
@@ -73,11 +74,13 @@ type authOperator struct {
...
-	services    corev1client.ServicesGetter
-	endpoints   corev1client.EndpointsGetter
-	secrets     corev1client.SecretsGetter
-	configMaps  corev1client.ConfigMapsGetter
-	deployments appsv1client.DeploymentsGetter
+	services                corev1client.ServicesGetter
+	endpoints               corev1client.EndpointsGetter
+	secrets                 corev1client.SecretsGetter
+	configMaps              corev1client.ConfigMapsGetter
+	namespaces              corev1client.NamespacesGetter
```
stlaz: you don't need `namespaces` to be a part of this structure

nee1esh:
pkg/operator2/operator.go:147:49: c.bootstrapUserDataGetter.IsEnabled undefined (type bootstrapauthenticator.BootstrapUserDataGetter has no field or method IsEnabled)

vareti:
> pkg/operator2/operator.go:147:49: c.bootstrapUserDataGetter.IsEnabled undefined (type bootstrapauthenticator.BootstrapUserDataGetter has no field or method IsEnabled)

it depends on https://github.com/openshift/library-go/pull/594. Once library-go PR is merged, changes can be imported to fix this error

pkg/operator2/starter.go
```diff
@@ -35,31 +36,31 @@ const (
 	resync = 20 * time.Minute
 )
 
-func RunOperator(ctx *controllercmd.ControllerContext) error {
+func RunOperator(controllerCtx *controllercmd.ControllerContext) error {
```
stlaz: why the name change?

pkg/operator2/routercerts/controller.go
```diff
@@ -197,20 +198,20 @@ func verifyWithAnyCertificate(serverCerts []*x509.Certificate, options x509.Veri
 	return err
 }
 
-func (c *RouterCertsDomainValidationController) Run(workers int, stopCh <-chan struct{}) {
+func (c *RouterCertsDomainValidationController) Run(ctx context.Context, workers int) {
```
stlaz: Please move the non-automatic changes outside the bump commit

pkg/operator2/starter.go
```diff
@@ -204,7 +206,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	}
 
 	for _, controller := range []interface {
-		Run(workers int, stopCh <-chan struct{})
+		Run(ctx context.Context, workers int)
```
stlaz: if the name change was because of this, wouldn't it make more sense to change it here instead?

pkg/operator2/starter.go
```diff
@@ -186,10 +187,11 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	)
 
 	// TODO remove this controller once we support Removed
-	managementStateController := management.NewOperatorManagementStateController("authentication", operatorClient, ctx.EventRecorder)
+	managementStateController := management.NewOperatorManagementStateController("authentication", operatorClient, controllerCtx.EventRecorder)
 	management.SetOperatorNotRemovable()
 	// TODO move to config observers
 	// configobserver.NewConfigObserver(...)
+	ctx := controllerCtx.Ctx
```
stlaz: why not just pass it directly?

pkg/operator2/operator.go
```diff
@@ -330,12 +336,19 @@ func (c *authOperator) handleSync(operatorConfig *operatorv1.Authentication) err
 		operatorDeployment,
 		resourceVersions...,
 	)
+
+	// redeploy on operatorConfig.spec changes or when bootstrap user is deleted
+	forceRollOut := (operatorConfig.Generation != operatorConfig.Status.ObservedGeneration)
+	if !c.bootstrapUserExists() && c.bootstrapUserChangeRollOut {
```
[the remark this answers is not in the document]
vareti: Done!

pkg/operator2/routercerts/controller.go
vareti: Created a new commit for these changes

pkg/operator2/starter.go
vareti: I only wanted to make it more readable. `ctx.Ctx` does not seem to look good for reading. Since it is used in many places, created a new variable instead of passing directly.

stlaz:
/lgtm
CI appears to be on fire

PR #220, opened 2019-11-25
Add manifest annotations for hosted deployment exclusions

csrwng:
Enables the CVO to exclude manifests in an externally hosted control plane deployment
See openshift/cluster-version-operator#252

### DIFF
--- a/manifests/07_deployment.yaml
+++ b/manifests/07_deployment.yaml
@@ -7,6 +7,7 @@ metadata:
     app: authentication-operator
   annotations:
     config.openshift.io/inject-proxy: operator
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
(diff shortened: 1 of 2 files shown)


stlaz:
Where's the enhancement for this?

stlaz:
/hold
This shan't merge without an enhancement

csrwng:
Because of timing, this change is meant to be part of an internal API for now. It will be eventually replaced by an official enhancement in 4.4 (CVO profiles).

stlaz:
/hold stop
/lgtm

I see. Who would I be to stop a well-meant internal API change. Please note that I will consider reverting if the enhancement does not arrive in 4.4 enhancement freeze time.

stlaz:
/hold cancel
because /hold stop will not work obviously

PR #221, opened 2019-11-26
Bug 1776421: Add manifest annotations for hosted deployment exclusions (4.3)

csrwng:
Enables the CVO to exclude manifests in an externally hosted control plane deployment
See openshift/cluster-version-operator#252


stlaz:
this is a backport of https://github.com/openshift/cluster-authentication-operator/pull/220

PR #222, opened 2019-11-28
[WIP] [ugly commits] add config observation

stlaz:
currently probably quite fragile, many FIXMEs and TODOs

The previous resource sync for the resources from the oauth config was replaces by syncing it from the configobserver and it seems to work, just as well as does the slightly reworked logic about how mounts are placed in the pods

- [ ] add observers for the fields of the authentication object
- [ ] add observation for whatever we need from the consoles
- [ ] add observation for whatever we need from the infrastructures object
- [ ] make the deployment use the observed config
- [ ] ?
- [ ] profit


stlaz:
did some very wild rebasing to make this reviewable. Given that github does not respect logical ordering of commits but does it chronologically instead, any reviewer might want to pull the changes and review locally.

pkg/operator2/configobservation/oauth/observe_idps_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
+)
+
+type mockResourceSyncer struct {
+	t      *testing.T
+	synced map[string]string
+}
+
+func (rs *mockResourceSyncer) SyncConfigMap(destination, source resourcesynccontroller.ResourceLocation) error {
```
marun: (No action required) Would it be make sense to avoid mocking by separating api calls from logic, e.g.

```golang

type idpResult {
  currentSyncData  *datasync.ConfigSyncData
  observedSyncData *datasync.ConfigSyncData
  observedConfig   map[string]interface{}
  eventMessage     string
}

// With dependencies on the API, this function is ideally validated in e2e
func ObserveIdentityProviders(genericlisters configobserver.Listers, recorder events.Recorder, previouslyObservedConfig map[string]interface{}) (map[string]interface{}, []error) {
    listers := genericlisters.(configobservation.Listers)
    resourceSyncer := genericlisters.ResourceSyncer()

    oauthConfig, _ := listers.OAuthLister.Get("cluster")
    ... 

    result, _ := computeIdpResult(oathConfig, previouslyObservedConfig)
    ...

    datasync.HandleIdPConfigSync(resourceSyncer, result.currentSyncData, result.observedSyncData)
    
    if len(result.eventMessage) {
        recorder.Eventf("ObserveIdentityProviders", result.eventMessage)
    }

    return result.observedConfig, nil
}

// With no dependencies on the API, the logic of this function is ideally validated with unit testing
func computeIdpResult(oathConfig *v1.OAuth, previousConfig map[string]interface{}) (*idpResult, error) {
  ...
}
```

pkg/operator2/configobservation/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,89 @@
...
+	currentTokenConfig, _, err := unstructured.NestedMap(previouslyObservedConfig, tokenConfigPath...)
+	if err != nil {
+		return previouslyObservedConfig, append(errs, err)
+	}
+
+	currentAccessTokenMaxAgeSecondsFloat, _, err := unstructured.NestedFloat64(currentTokenConfig, "accessTokenMaxAgeSeconds")
+	if err != nil {
+		errs = append(errs, err)
+	}
+	currentAccessTokenMaxAgeSeconds := float64(currentAccessTokenMaxAgeSecondsFloat)
```
marun: Is this conversion necessary given that `NestedFloat64` returns a float64?

pkg/operator2/configobservation/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/operator2/configobservation"
+)
+
+const defaultAccessTokenMaxAgeSeconds = float64(86400) // a day
```
marun: Given that the golant struct fields are int32 and the value in seconds is fixed point, is it desirable to use a floating point type?

pkg/operator2/configobservation/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,89 @@
...
+
+	observedAccessTokenMaxAgeSeconds := float64(oauthConfig.Spec.TokenConfig.AccessTokenMaxAgeSeconds)
+	if observedAccessTokenMaxAgeSeconds == 0 {
+		observedAccessTokenMaxAgeSeconds = defaultAccessTokenMaxAgeSeconds
+	}
+
+	observedTokenConfigFieldMap := map[string]interface{}{}
+	unstructured.SetNestedField(observedTokenConfigFieldMap, observedAccessTokenMaxAgeSeconds, "accessTokenMaxAgeSeconds")
+
+	observedAccessTokenInactivityTimeoutSeconds := float64(oauthConfig.Spec.TokenConfig.AccessTokenInactivityTimeoutSeconds)
```
marun: It appears that this switch statement is just ensuring a value less than 0 is reset to zero, so maybe simplify as follows?

```golang
if val < 0 {
  val = 0 
}
fieldMap["val"] = val
```

pkg/operator2/configobservation/oauth/template_conversions.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, nil, err
+	}
+
+	switch brand {
+	case okdBrand:
+		// do nothing, OAuth server has right branding for this
+
+	case ocpBrand, "dedicated", "online", "azure":
+		// all of these are equivalent to ocp
+		templates = osinv1.OAuthTemplates{
```
marun: Maybe define this outside of the switch statement and use it here and for default?

pkg/operator2/configobservation/routersecret/observe_router_secret.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/openshift/cluster-authentication-operator/pkg/operator2/configobservation"
+)
+
+func ObserveRouterSecret(genericlisters configobserver.Listers, recorder events.Recorder, previouslyObservedConfig map[string]interface{}) (map[string]interface{}, []error) {
+	namedCertificatesPath := []string{"servingInfo", "namedCertificates"}
+	listers := genericlisters.(configobservation.Listers)
+	errs := []error{}
+
+	currentNamedCertificates, _, err := unstructured.NestedFieldCopy(previouslyObservedConfig, namedCertificatesPath...)
```
marun: Maybe use NestedSlice?

pkg/operator2/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,60 @@
...
+
+var clusterDefaultCORSALlowedOrigins = []string{
+	`//127\.0\.0\.1(:|$)`,
+	`//localhost(:|$)`,
+}
+
+func ObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	const corsAllowedOriginsPath = "corsAllowedOrigins"
+
+	listers := genericListers.(configobservation.Listers)
```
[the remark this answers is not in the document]
stlaz: The cast should be safe (those listers are constant), if it fails, the programmer did not do his job well (happened to me several times while developing this).

pkg/operator2/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,60 @@
...
+	currentCORSSet.Insert(clusterDefaultCORSALlowedOrigins...)
+
+	observedConfig := map[string]interface{}{}
+	apiServer, err := listers.APIServerLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("apiserver.config.openshift.io/cluster: not found")
+		return defaultConfig, errs
+	}
+	if err != nil {
+		return existingConfig, errs
```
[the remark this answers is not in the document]
stlaz: The error should probably be appended, correct.

`existingConfig` is returned because returning `defaultConfig` might rewrite whatever we had even if `err` is just a transient network error, and we would have another rewrite once that is over. I'll add a comment.

pkg/operator2/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,60 @@
...
+	`//127\.0\.0\.1(:|$)`,
+	`//localhost(:|$)`,
+}
+
+func ObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	const corsAllowedOriginsPath = "corsAllowedOrigins"
+
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	defaultConfig := map[string]interface{}{}
```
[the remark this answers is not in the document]
stlaz: `SetNestedStringSlice()` does a small little bit more than just that, I'd rather stick with the current code.

pkg/operator2/configobservation/oauth/idp_conversions.go
```diff
@@ -43,7 +46,61 @@ type idpData struct {
 	login     bool
 }
 
-func (c *authOperator) convertProviderConfigToIDPData(providerConfig *configv1.IdentityProviderConfig, syncData *configSyncData, i int) (*idpData, error) {
+// FIXME: will have to return sync data as well, also maybe warnings?
```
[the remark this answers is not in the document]
stlaz: historical artifacts from WIP, will be removed in next revision, thx

pkg/operator2/deployment.go
```diff
@@ -35,61 +41,84 @@ func defaultDeployment(
...
-		{
-			name:      "v4-0-config-system-session",
-			configmap: false,
-			path:      "/var/config/system/secrets/v4-0-config-system-session",
-			keys:      []string{"v4-0-config-system-session"},
+	// FIXME: ignore errors for now, this should probably be done earlier when working with
+	//        the authentication.operator resource anyway
+	idpSyncData, _ := getSyncDataFromOperatorConfig(&operatorConfig.Spec.ObservedConfig)
+
+	for _, data := range append(
```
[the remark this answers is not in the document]
stlaz: legacy choice, this code is now gone thanks to @sanchezl :champagne: \o/

pkg/operator2/configobservation/oauth/observe_idps_test.go
stlaz: Both seem to bear more or less similar results I guess, I find the mocking to be quite elegant way to test this, but I can see how your approach separates the logical unit outside so that it is easier to raise expectations over it.

pkg/operator2/configobservation/oauth/observe_tokenconfig.go
stlaz: I need to re-check this module, I think I left it at a WIP state...

pkg/operator2/configobservation/oauth/observe_tokenconfig.go
stlaz: glad you asked, this was such a PITA to get right.

The config is really a json that gets deserialized into golang nested interfaces.

When we make this int32 and store as such, the information about it being int is obviously stored.

Eventually, the `configObserver`, which lives in library-go, compares the config it got originally with the one that it got as observed config.

This is where all the fun happens. because the json parser actually likes to think of numbers as of floats prior to them being integers. That means that if we make this integer, same jsons will actually be considered different in their go interpretation even if our comparison here tells you the numbers are equal. That results in hotloop trying to update the `observedConfig` field on same values.

I tried many things, played with the json parser flags, use `json.Number` as the default representation for numbers found in the jsons we have here but nothing except this really worked.

I did a terrible job with code comments about this, I see.

pkg/operator2/configobservation/apiserver/observe_cors.go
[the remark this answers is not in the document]
stlaz: on a second thought, I might try to do that

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+
+	switch s.Resource {
+	case ConfigMapType:
+		vol.ConfigMap = &corev1.ConfigMapVolumeSource{
+			LocalObjectReference: corev1.LocalObjectReference{
+				Name: volName,
+			},
+			Items: items,
+		}
+	default:
```
marun, line 197: (No action required) Maybe specify SecretType as an explicit case and panic in default?

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+	// idps that are synced have this prefix
+	return fmt.Sprintf("v4-0-config-user-idp-%d-%s", i, field)
+}
+
+func getIDPPath(i int, resource, dest string) string {
+	// root path for IDP data
+	return fmt.Sprintf("/var/config/user/idp/%d/%s/%s", i, resource, dest)
+}
+
+func syncOrDie(syncFunc func(dest, src resourcesynccontroller.ResourceLocation) error, dest, src string) {
```
marun, line 221: nit: Maybe factor out the common functionality between this and `syncSecretOrDie`?

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+
+type ResourceType string
+
+const (
+	ConfigMapType ResourceType = "configMap"
+	SecretType    ResourceType = "secret"
+)
+
+type sourceData struct {
+	Src      string       `json:"src"`  // name of the source in openshift-config namespace
```
marun, line 31: nit: why isn't the name of this field just `Name` ? As per further comments on fields in this struct, a good name can largely obviate the need for explanatory comments.

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+	Path     string       `json:"path"` // the mount path that this source is mapped to
+	Key      string       `json:"key"`
+	Resource ResourceType `json:"resource"`
```
marun, line 34: nit: why isn't the name of this field `ResourceType` or `Type` ?

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+type ResourceType string
+
+const (
+	ConfigMapType ResourceType = "configMap"
+	SecretType    ResourceType = "secret"
+)
+
+type sourceData struct {
+	Src      string       `json:"src"`  // name of the source in openshift-config namespace
+	Path     string       `json:"path"` // the mount path that this source is mapped to
```
marun, line 32: nit: why isn't the name of this field `MountPath` ?

pkg/operator2/datasync/validation.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return []error{err}
+	}
+
+	data, exists := s.Data[src.Key]
+	if !exists {
+		return []error{fmt.Errorf("missing required key: %q", src.Key)}
+	}
+
+	return validators[src.Key](data)
```
marun, line 39: Would it make sense to check that the key was found and return an error if not?

pkg/operator2/datasync/validation.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return []error{err}
+	}
+
+	data, exists := cm.Data[src.Key]
+	if !exists {
+		return []error{fmt.Errorf("missing required key: %q", src.Key)}
+	}
+
+	return validators[src.Key]([]byte(data))
```
marun, line 53: ditto

pkg/operator2/datasync/validation.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return errs
+}
+
+func validatePrivateKey(pemKey []byte) []error {
+	if len(pemKey) == 0 {
+		return []error{fmt.Errorf("required private key is empty")}
+	}
+	if _, err := keyutil.ParsePrivateKeyPEM(pemKey); err != nil {
+		return []error{fmt.Errorf("failed to parse the private key")}
```
marun, line 86: Would it make sense to wrap the original error to aid in troubleshooting?

pkg/operator2/datasync/validation.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	errs := []error{}
+	now := time.Now()
+
+	cert, rest := pem.Decode(pemCerts)
+	for ; cert != nil; cert, rest = pem.Decode(rest) {
+		parsed, err := x509.ParseCertificate(cert.Bytes)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to parse a certificate in the chain"))
```
marun, line 118: Would it make sense to include the original error?

pkg/operator2/datasync/validation_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+func Test_validateSecret(t *testing.T) {
```
marun, line 15: (No action required) Consider validating the field-agnostic wrapper separately from the actual validation functions to simplify writing the tests (i.e. no indexer/lister required) and make life easier for future maintainers when they attempt to determine coverage.

pkg/operator2/datasync/validation_test.go
```diff
@@ -0,0 +1,351 @@
...
+		name    string
+		src     sourceData
+		secrets []*corev1.Secret
+		want    []error
+	}{
+		{
+			name: "secret missing",
+			src:  sourceData{Src: "somesecret", Key: corev1.TLSPrivateKeyKey},
+			want: []error{
+				fmt.Errorf("secret \"somesecret\" not found"),
```
marun, line 26: (No action required) Consider avoiding a dependency on a specific error message as this style of testing is brittle and will have to be maintained in lock-step with the code.

pkg/operator2/datasync/validation.go
stlaz: When it comes to private keys, I'd rather avoid printing errors that I don't directly control.

pkg/controllers/configobservation/oauth/idp_conversions.go
```diff
@@ -50,10 +53,69 @@ type idpData struct {
...
+) ([]interface{}, *datasync.ConfigSyncData, []error) {
+
+	converted := []osinv1.IdentityProvider{}
+	syncData := datasync.NewConfigSyncData()
+	errs := []error{}
+
+	for i, idp := range defaultIDPMappingMethods(identityProviders) {
+		data, err := convertProviderConfigToIDPData(cmLister, secretsLister, &idp.IdentityProviderConfig, syncData, i)
+		if err != nil {
+			klog.Errorf("failed to honor IDP %#v: %v", idp, err)
```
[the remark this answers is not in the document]
stlaz: remnants of the ages past, removed

pkg/controllers/configobservation/oauth/idp_conversions.go
```diff
@@ -50,10 +53,69 @@ type idpData struct {
...
+				UseAsLogin:      data.login,
+				MappingMethod:   string(idp.MappingMethod),
+				Provider: runtime.RawExtension{
+					Raw: encodeOrDie(data.provider),
+				},
+			},
+		)
+	}
+
+	// TODO: the following is a bit perverted, I wonder if it could be done differently
```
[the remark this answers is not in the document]
stlaz: I think it's actually fine, I replaced the comment with something more useful.

pkg/controllers/configobservation/oauth/observe_idps.go
```diff
@@ -0,0 +1,99 @@
...
+	existingSyncData, err := GetIDPConfigSyncData(existingConfig)
+	if err != nil {
+		return existingConfig, append(errs, err)
+	}
+
+	oauthConfig, err := listers.OAuthLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		// revert to default state, meaning no IdPs
+		klog.Warning("oauth.config.openshift.io/cluster: not found")
+		return map[string]interface{}{}, nil
```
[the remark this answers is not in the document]
stlaz: added the errs there

pkg/controllers/configobservation/oauth/observe_idps.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/configobservation"
+	"github.com/openshift/cluster-authentication-operator/pkg/operator2/datasync"
+)
+
+var identityProvidersMounts = []string{"volumesToMount", "identityProviders"}
+
+func ObserveIdentityProviders(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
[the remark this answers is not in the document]
stlaz: The pruning should always happen, and I think it logically belongs within the observation function. We wouldn't have to deal with it at all in theory, but it's a safety precaution based on the observations of the past.

pkg/controllers/configobservation/oauth/observe_templates.go
```diff
@@ -0,0 +1,102 @@
...
+
+	var observedTemplates interface{}
+	if templates != nil {
+		convertedBytes, err := json.Marshal(templates)
+		if err != nil {
+			return existingConfig, append(errs, err)
+		}
+
+		if err := json.NewDecoder(bytes.NewBuffer(convertedBytes)).Decode(&observedTemplates); err != nil {
+			return nil, append(errs, fmt.Errorf("decode of observed config failed with error: %v", err))
```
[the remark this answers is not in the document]
stlaz: added the existing config instead

pkg/operator2/oauth.go
```diff
@@ -7,21 +7,31 @@ import (
...
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
-	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/configobservation"
+)
+
+var (
+	scheme  = runtime.NewScheme()
+	codecs  = serializer.NewCodecFactory(scheme)
+	encoder = codecs.LegacyCodec(osinv1.GroupVersion) // TODO I think there is a better way to do this
```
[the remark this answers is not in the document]
stlaz: It's the question old as the repo, but you're right, @sttts might be able to know the answer, I'll ask

pkg/controllers/configobservation/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,110 @@
...
+	oauthConfig, err := listers.OAuthLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warning("oauth.config.openshift.io/cluster: not found")
+		return map[string]interface{}{
+			"oauthConfig": map[string]interface{}{
+				"tokenConfig": map[string]interface{}{
+					"accessTokenMaxAgeSeconds": defaultAccessTokenMaxAgeSeconds,
+				},
+			},
+		}, nil
```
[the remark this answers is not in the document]
stlaz: added the errors

pkg/controllers/configobservation/oauth/template_conversions.go
```diff
@@ -0,0 +1,104 @@
...
+		templates = ocpDefaults
+
+	default:
+		if defaultBrand == ocpBrand {
+			// build-time ocp selection
+			templates = ocpDefaults
+		}
+	}
+
+	// user-configured overrides everything else, individually.
```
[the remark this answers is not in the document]
stlaz: The comment is now improved :)

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+			syncFunc = resourceSyncer.SyncConfigMap
+			newConfigMapNames.Insert(dest)
+		} else {
+			newSecretNames.Insert(dest)
+		}
+
+		syncOrDie(syncFunc, dest, newData.data[dest].Src)
+	}
+
+	for _, dst := range sets.StringKeySet(oldData.data).List() {
```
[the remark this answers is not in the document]
stlaz: this is the list of keys, which are in fact the destination, see the description of the `ConfigSyncData` at the beginning of the file

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+		if oldData.data[dst].Resource == ConfigMapType {
+			oldConfigMapNames.Insert(dst)
+		} else {
+			oldSecretNames.Insert(dst)
+		}
+	}
+
+	notInUseConfigMapNames := oldConfigMapNames.Difference(newConfigMapNames)
```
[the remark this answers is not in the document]
stlaz: lol, yeah

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	for dest := range notInUseSecretNames {
+		syncOrDie(resourceSyncer.SyncSecret, dest, "")
+	}
+}
+
+// newSourceDataIDP returns a name which is unique amongst the IdPs, and sourceData
+// which describes the volumes and mount volumes to mount the CM/Secret to
+func newSourceDataIDP(index int, resourceType ResourceType, resourceName, field, key string) (string, sourceData) {
+	dest := getIDPName(index, field)
```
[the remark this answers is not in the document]
stlaz: I'd still go with `dest`, it's the name of the destination resource

pkg/operator2/datasync/syncdata.go
```diff
@@ -0,0 +1,238 @@
...
+	dirPath := getIDPPath(index, string(resourceType), dest)
+
+	ret := sourceData{
```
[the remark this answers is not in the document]
stlaz: aye aye, returning `ret` directly, kept `dest` and `dirPath` variables for verbosity

pkg/operator2/datasync/syncdata.go
stlaz: addressed

pkg/operator2/datasync/syncdata.go
stlaz: done, exported the sync function and reused

pkg/operator2/datasync/syncdata.go
stlaz: You're right, I was blindly refactoring the old code without thinking about the names too much, but these name suggestion do make the code more readable, addressed :+1:

pkg/operator2/datasync/validation.go
stlaz: I don't mind the panic in case the programmer forgets to add a validator

pkg/operator2/datasync/validation.go
stlaz: added

pkg/controllers/configobservation/oauth/observe_tokenconfig_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"k8s.io/client-go/tools/cache"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/configobservation"
+)
+
+func TestObserveTokenConfig(t *testing.T) {
```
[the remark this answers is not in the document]
stlaz: added 3 tests, 2 of them are more or less "the universe is sane", but they do test the corner cases

stlaz:
addressed the comments, during which I noticed default authz token max age, which is not actually being dynamically set, was missing, fixed that

pkg/controllers/configobservation/oauth/observe_idps.go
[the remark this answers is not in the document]
marun: Why should the pruning happen for anything but the existing config? Isn't newly created config always just the field(s) required?

pkg/controllers/configobservation/oauth/observe_tokenconfig_test.go
```diff
@@ -34,6 +34,52 @@ func TestObserveTokenConfig(t *testing.T) {
...
+					TokenConfig: configv1.TokenConfig{
+						AccessTokenMaxAgeSeconds: 0,
+					},
+				},
+			},
+			previouslyObservedConfig: map[string]interface{}{},
+			expected: map[string]interface{}{
+				"oauthConfig": map[string]interface{}{
+					"tokenConfig": map[string]interface{}{
+						"accessTokenMaxAgeSeconds":    float64(86400),
```
marun, line 56: (No action required) Maybe this is just my ignorance of accepted practice for cluster operators, but I find it strange to see magic numbers rather than constants.

pkg/controllers/configobservation/oauth/observe_tokenconfig_test.go
```diff
@@ -34,6 +34,52 @@ func TestObserveTokenConfig(t *testing.T) {
...
+						"authorizeTokenMaxAgeSeconds": float64(300),
+					},
+				},
+			},
+			errors: []error{},
+		},
+		{
+			name: "inactivity 0 means disabled",
```
marun, line 64: I'm confused - how is this different from `"max age 0 still means default max age"`?

marun:
@stlaz Thank you for cleanup! Other than the seemingly duplicated test, lgtm. Please squash for merge.

pkg/controllers/configobservation/oauth/observe_idps.go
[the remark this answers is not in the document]
stlaz: that was in the case we got rid of the pruning, it needs to happen for any output and especially for the existing config, perhaps bad wording on my side

stlaz:
/retest
unrelated failures

PR #226, opened 2019-12-11
Use gomod

stlaz:


deads2k:
More deps that maybe I'd like, but we can start here.

/lgtm

stlaz:
replaced spaces with tabs in depmagnet, retagging

PR #229, opened 2019-12-11
Enable bindata

sanchezl:
Use bindata (instead of inline code) to create the oauth-openshift deployment.


bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - name: v4-0-config-system-cliconfig
+          configMap:
+            name: v4-0-config-system-cliconfig
+            defaultMode: 420
+            optional: true
+        - name: v4-0-config-system-serving-cert
+          secret:
+            secretName: v4-0-config-system-serving-cert
+            defaultMode: 420
+            optional: true
```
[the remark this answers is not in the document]
sanchezl: Done

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - name: v4-0-config-system-session
+          secret:
+            secretName: v4-0-config-system-session
+            defaultMode: 420
+            optional: true
+        - name: v4-0-config-system-cliconfig
+          configMap:
+            name: v4-0-config-system-cliconfig
+            defaultMode: 420
+            optional: true
```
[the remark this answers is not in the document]
sanchezl: Done

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
+      volumes:
+        - name: v4-0-config-system-session
+          secret:
+            secretName: v4-0-config-system-session
+            defaultMode: 420
+            optional: true
```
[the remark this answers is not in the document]
sanchezl: Done

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - name: v4-0-config-system-serving-cert
+          secret:
+            secretName: v4-0-config-system-serving-cert
+            defaultMode: 420
+            optional: true
+        - name: v4-0-config-system-service-ca
+          configMap:
+            name: v4-0-config-system-service-ca
+            defaultMode: 420
+            optional: true
```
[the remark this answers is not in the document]
sanchezl: Done

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - name: v4-0-config-system-router-certs
+          secret:
+            secretName: v4-0-config-system-router-certs
+            defaultMode: 420
+            optional: true
```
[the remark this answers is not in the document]
sanchezl: Done

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - name: v4-0-config-system-ocp-branding-template
+          secret:
+            secretName: v4-0-config-system-ocp-branding-template
+            defaultMode: 420
+            optional: true
```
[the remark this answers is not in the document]
sanchezl: Done

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,141 @@
...
+        - name: oauth-openshift
+          image: ${IMAGE}
+          command:
+            - /bin/bash
+            - '-ec'
+          args:
+            - >
+              if [ -s
+              /var/config/system/configmaps/v4-0-config-system-trusted-ca-bundle/ca-bundle.crt
+              ]; then
```
stlaz: This condition might be easier to read if it's just one line

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,141 @@
...
+                  echo "Copying system trust bundle"
+                  cp -f /var/config/system/configmaps/v4-0-config-system-trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+              fi
+              exec oauth-server osinserver
+              --config=/var/config/system/configmaps/v4-0-config-system-cliconfig/v4-0-config-system-cliconfig
+              --v=${LOG_LEVEL}
```
stlaz: These line get squashed in the real deployment so that the args look like this:
```
      if [ -s /var/config/system/configmaps/v4-0-config-system-trusted-ca-bundle/ca-bundle.crt ]; then
          echo "Copying system trust bundle"
          cp -f /var/config/system/configmaps/v4-0-config-system-trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
      fi exec oauth-server osinserver --config=/var/config/system/configmaps/v4-0-config-system-cliconfig/v4-0-config-system-cliconfig --v=2
```
As @soltysh pointed out earlier in his yaml-multiline findings, you may want to start with
```
args:
  - |
```
which is supposed to keep the newlines.

See https://yaml-multiline.info/ which is quite helpful

You'll have to add backslashes before each line ending in the `exec` part I think

sanchezl: @stlaz what do you think of running the "copy system bundle" part in an init container? Then the cmd/args for this container will be more maintainable?

stlaz: Wouldn't you need a volume for such a change to persist?

sanchezl: Yes, I think an EmptyDir volume would do as this is the only file expected in the target directory. But I'll leave that for another day since this is working as is.

bindata/oauth-openshift/deployment.yaml
```diff
@@ -0,0 +1,139 @@
...
+          terminationMessagePolicy: FallbackToLogsOnError
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
+      volumes:
+        - name: v4-0-config-system-session
+          secret:
+            secretName: v4-0-config-system-session
+            defaultMode: 420
```
deads2k: this looks suspicious/wrong.  Aren't these supposed to be octal?

deads2k:
I'm glad to see this kind of a change

/approve

bindata/oauth-openshift/deployment.yaml
sanchezl: 420 decimal == 0644 octal, which is the default anyway, so I have removed these lines.

PR #231, opened 2019-12-18
Bug 1788135: Fix IPv6 Support

russellb:
This PR includes the changes necessary to work in a cluster with IPv6 enabled (either IPv6-only or IPv4 and IPv6).

"bindNetwork: tcp" will always do the right thing, based on whether IPv4/IPv6/both are enabled.

One other fix was needed to more carefully join a host and port number together.

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"reflect"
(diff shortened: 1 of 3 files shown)


pkg/operator2/operator.go
```diff
@@ -557,7 +558,7 @@ func (c *authOperator) getAPIServerIPs() ([]string, error) {
 
 		ips := make([]string, 0, len(subset.Addresses))
 		for _, address := range subset.Addresses {
-			ips = append(ips, fmt.Sprintf("%s:%d", address.IP, targetPort))
```
stlaz: why the int -> int conversion?

stlaz:
/lgtm cancel
please fix the small issue or explain what I missed while looking at the code and we shall merge this

pkg/operator2/operator.go
russellb: You're right, not needed.

I either copied this or did it out of habit.  In lots of other places where this change is needed, the port is an `int64`, not `int`.

stlaz:
/lgtm
/approve
lets' see if building fails even after retested

russellb:
This is also needed in release-4.3

/cherry-pick release-4.3

PR #232, opened 2020-01-03
bump kube to 1.17 and pull latest library-go

stlaz:
Building the operator images fails because OSBS builders are disconnected from the internets but the builder image tries to use go mod and fails on dependency pulls.

Latest library-go forces using the vendor directory in builds.

This PR also bumps the builder images in Dockerfiles to golang-1.13

cc @tnozicka @sanchezl @vareti


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS builder
```
tnozicka: I'd drop this whole file as we have done in other repos. There is no OKD yet and registry.svc.ci.openshift.org/openshift/origin-v4.0:base is actually OCP anyways

go.mod
```diff
@@ -3,40 +3,28 @@ module github.com/openshift/cluster-authentication-operator
...
+	k8s.io/klog v1.0.0
 	monis.app/go v0.0.0-20190702030534-c65526068664
-	sigs.k8s.io/structured-merge-diff v1.0.0 // indirect
 )
 
 replace (
 	github.com/openshift/api => github.com/openshift/api v3.9.1-0.20191209132752-992bc3a41fe6+incompatible
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191205152420-9faca5198b4f
-	github.com/openshift/library-go => github.com/openshift/library-go v0.0.0-20191211124107-e0f1590a316e
+	github.com/openshift/library-go => github.com/openshift/library-go v0.0.0-20200103110434-cd8a585d2a77
```
tnozicka: I think we aim at using require here

stlaz:
@tnozicka I think I will be able to drop Dockerfile only after the dependent CI PR merged

Dockerfile
tnozicka: CI was using wrong dockerfile, pls drop it in a followup

go.mod
```diff
@@ -3,40 +3,21 @@ module github.com/openshift/cluster-authentication-operator
...
-	github.com/certifi/gocertifi v0.0.0-20180905225744-ee1a9a0726d2 // indirect
 	github.com/getsentry/raven-go v0.2.1-0.20190513200303-c977f96e1095 // indirect
 	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
-	github.com/openshift/api v3.9.1-0.20191107113335-6feaabc7037a+incompatible
-	github.com/openshift/client-go v0.0.0-20191022152013-2823239d2298
-	github.com/openshift/library-go v0.0.0-20191106124920-97392cdc9bd3
-	github.com/pkg/profile v1.3.0 // indirect
+	github.com/openshift/api v0.0.0-20200106203948-7ab22a2c8316
+	github.com/openshift/client-go v0.0.0-20191219165006-ac3b642258cc
+	github.com/openshift/library-go v0.0.0-20200103144857-38e0f6451b16
```
mfojtik: v0.0.0-20200106191802-9821002633e8

seems to be the latest, yours is outdated :-)

PR #235, opened 2020-01-07
Remove unused dockerfile and add gomod validation

stlaz:
/assign @marun


stlaz:
depends on https://github.com/openshift/release/pull/6549

PR #237, opened 2020-01-10
Enable health checking of oauth-server endpoints

marun:
Failed health checks will result in degraded operator conditions. This is intended to allow differentiation between endpoint health (a responsibility of this operator) and route health (a responsibility of the network operator).

Forked from #211

Co-authored-by: Maru Newby <marun@redhat.com>

### DIFF
--- a/manifests/03_openshift_service_ca.yaml
+++ b/manifests/03_openshift_service_ca.yaml
@@ -1,0 +1,8 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: service-ca-bundle
+  namespace: openshift-authentication-operator
+  annotations:
+    service.beta.openshift.io/inject-cabundle: "true"
+data: {}
--- a/pkg/controller/ingressstate/ingress_state_controller.go
+++ b/pkg/controller/ingressstate/ingress_state_controller.go
@@ -1,0 +1,344 @@
+package ingressstate
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const (
+	ingressStateControllerWorkQueueKey = "key"
+
+	maxToleratedPodPendingDuration = 5 * time.Minute
+)
+
+// degradedConditionTypes contains all conditions that this controller manages
+var degradedConditionTypes = sets.NewString(
+	"IngressStateEndpointsDegraded",
+	"IngressStatePodsDegraded",
+)
+
+type IngressStateController struct {
+	endpointsGetter corev1client.EndpointsGetter
+	podsGetter      corev1client.PodsGetter
+	queue           workqueue.RateLimitingInterface
+	cachesToSync    []cache.InformerSynced
+	targetNamespace string
+	operatorClient  v1helpers.OperatorClient
+	eventRecorder   events.Recorder
+}
+
+func NewIngressStateController(kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	endpointsGetter corev1client.EndpointsGetter,
+	podsGetter corev1client.PodsGetter,
+	operatorClient v1helpers.OperatorClient,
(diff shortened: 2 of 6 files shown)


manifests/03_openshift_service_ca.yaml
```diff
@@ -0,0 +1,8 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: service-ca-bundle
+  namespace: openshift-authentication
```
stlaz: wrong namespace?

stlaz:
as a result of the above, I can see authn-operator logs the foillowing failures:
```
E0110 12:18:26.052819       1 ingress_state_controller.go:168] key failed with : error reading service ca bundle: open /var/run/configmaps/service-ca-bundle/service-ca.crt: no such file or directory
```

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,328 @@
...
+	// in that case.
+	unhealthyEndpoints := []string{}
+	unhealthyPodCount := 0
+	unhealthyPodMessages := []string{}
+
+	for _, subset := range endpoints.Subsets {
+		// Report degraded if all addresses for the endpoint are not ready
+		if len(subset.Addresses) == 0 && len(subset.NotReadyAddresses) > 0 {
+			msg := fmt.Sprintf("%d endpoints for oauth-server are reporting 'not ready'", len(subset.NotReadyAddresses))
+			conditions = append(conditions, endpointsDegraded("NonReadyEndpoints", msg))
```
stlaz: this (and later uses of the `endpointsDegraded` helper) overrides the `IngressStateEndpointsDegraded` condition with the latest msg, is that intended?

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,328 @@
...
+)
+
+const (
+	ingressStateControllerWorkQueueKey = "key"
+
+	maxToleratedPodPendingDuration = 5 * time.Minute
+)
+
+// degradedConditionNames contains all conditions that this controller manages
+var degradedConditionNames = sets.NewString(
```
stlaz: I guess `degradedConditionTypes` would be more accurate to stick with the conditions lingo

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,328 @@
...
+			Message: fmt.Sprintf("Unhealthy pods found: %s", strings.Join(unhealthyPodMessages, ",")),
+		})
+	}
+
+	// Ensure condition names are found in degradedConditionNames
+	for _, condition := range conditions {
+		if !degradedConditionNames.Has(condition.Type) {
+			return nil, fmt.Errorf("Unsupported degraded condition name %q", condition.Type)
+		}
+	}
```
stlaz: Why, don't you trust yourself? Isn't this more of a unit-test material rather than production code?

manifests/03_openshift_service_ca.yaml
marun: Good catch!  Done.

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,328 @@
...
+// checkEndpointHealthz will check the health of given https://endpointIP:6443/healthz
+func checkEndpointHealthz(endpointIP string, rootCAs *x509.CertPool) error {
+	client := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: rootCAs,
+			},
+		},
+	}
+	url := fmt.Sprintf("https://%s:6443/healthz", endpointIP)
```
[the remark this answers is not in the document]
marun: TIL! Done.

pkg/controller/ingressstate/ingress_state_controller.go
marun: Hah! No. 

I find it easy to be ignorant when I inherit code in an unfamiliar domain (e.g. endpoints). I've updated the code with an assumption that I think makes the code simpler to reason about - that there will only ever be a single subset. This is true so long as the service exposes only a single port.

pkg/controller/ingressstate/ingress_state_controller.go
marun: Done

pkg/controller/ingressstate/ingress_state_controller.go
marun: Moved to unit test.

marun:
@stlaz Thank you for your review! Updated, PTAL.

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	podMessages := map[string][]string{}
+	unhealthyPodCount := 0
+	for _, address := range addresses {
+		if err := checkEndpoints(address.IP); err != nil {
+			unhealthyAddresses = append(unhealthyAddresses, fmt.Sprintf("%s:%v", address.IP, err))
+		}
+		if address.TargetRef != nil && address.TargetRef.Kind == "Pod" {
+			podName := address.TargetRef.Name
+			if _, alreadyChecked := podMessages[podName]; alreadyChecked {
+				// A pod only needs to be checked once per sync for any address
```
stlaz: looks like `continue`'s gone missing here

stlaz:
small nit, otherwise lgtm

pkg/controller/ingressstate/ingress_state_controller.go
marun: D'oh again!  Done.

marun:
@stlaz Added continue, squashed and rebased.  PTAL!

stlaz:
/test all
tests had a bad day yesterday apparently

PR #238, opened 2020-01-10
WIP: Add oauth-apiserver manifests

sttts:
bindata manifests. No wiring yet.


stlaz:
looks fine to me, this will probably need bindata-update.

Depending on how soon you would like to merge this you could even try to add the oauth-apiserver to /manifests/image-references so that we know once it's available.

PR #248, opened 2020-02-03
Bug 1727983: OIDC IdPs: allow challenge flows if OIDC server advertises password grants

stlaz:
When it's possible to perform password grants, allow challenge flows.
This means that we can do for the OIDC IdPs what we are doing for
all the other IdPs - just forwarding user's login/password.

However, the above behavior is discouraged for the OAuth 2.0 protocol
(https://tools.ietf.org/id/draft-ietf-oauth-security-topics-13.html#rfc.section.3.4)
In the long run, we should remove it and implement the
Authorization Code granting instead.

----

I think implementing Authorization Code OAuth 2.0 flows would need at least:
- changes to `oc` so that you are now required to pick the identity provider
- making the user aware what URL are they authenticating against when entering credentials via `oc`

(The above takes into consideration the case that the oauth-server was compromised and, while we can act as the user against the cluster, we can still at least protect the users' credentials from being stolen by the server just sending `WWW-Authenticate`, waiting for the password to come from `oc`)

- implementing the authorization code flow in the oauth-server itself

----

/hold
I would like the OpenShift-Online folks to help me test whether this suites their needs expressed in the related bug

/cc @deads2k @sttts @marun


pkg/operator2/idp.go
```diff
@@ -216,33 +216,37 @@ func (c *authOperator) convertProviderConfigToIDPData(providerConfig *configv1.I
 	return data, nil
 }
 
-func (c *authOperator) discoverOpenIDURLs(issuer, key string, ca configv1.ConfigMapNameReference) (*osinv1.OpenIDURLs, error) {
+// discoverOpenIDConfig retrieves basic information about an OIDC server with hostname
+// given by the `issuer` argument. It returns the authentication, token and user
+// info urls in *osinv1.OpenIDUrls structure, and a boolean value telling whether
+// our challenge flows can be used against such a server.
+func (c *authOperator) discoverOpenIDConfig(issuer, key string, ca configv1.ConfigMapNameReference) (*osinv1.OpenIDURLs, bool, error) {
```
marun: nit: Maybe simplify both this change and future maintenance by returning a module-local struct with fields for url and whether password grants are permitted? That would also have the advantage of being self documenting vs requiring as much detail in the docstring.

pkg/operator2/idp.go
```diff
@@ -311,6 +316,15 @@ func isValidURL(rawurl string, optional bool) bool {
 	return u.Scheme == "https" && len(u.Host) > 0 && len(u.Fragment) == 0
 }
 
+func passwordGrantsSupported(grantTypes []string) bool {
+	for _, t := range grantTypes {
+		if t == "password" {
```
marun: Should this check be case-agnostic?

pkg/operator2/idp.go
```diff
@@ -168,7 +168,7 @@ func (c *authOperator) convertProviderConfigToIDPData(providerConfig *configv1.I
 			return nil, fmt.Errorf(missingProviderFmt, providerConfig.Type)
 		}
 
-		urls, err := c.discoverOpenIDURLs(openIDConfig.Issuer, corev1.ServiceAccountRootCAKey, openIDConfig.CA)
+		urls, challengesAllowed, err := c.discoverOpenIDConfig(openIDConfig.Issuer, corev1.ServiceAccountRootCAKey, openIDConfig.CA)
```
marun: nit: Maybe name the boolean more specifically (e.g. 'passwordChallengeSupported') so that it's clear why challenges would be allowed?

stlaz:
@marun in the state this is now it won't work since the oidc-configuration could be client-specific, I need to be a bit more clever about detecting the password grants in the OIDC.

stlaz:
Updated the code, according to my finds:
```
Google:
{
  "error": "unsupported_grant_type",
  "error_description": "Invalid grant_type: password"
}

GitLab (wrong password):
{"error":"invalid_grant","error_description":"The provided authorization grant is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client."} 

- success returns:
{"access_token":"<token>","token_type":"Bearer","refresh_token":"<different_token>"}

- "something" grant type:
{"error":"unsupported_grant_type","error_description":"The authorization grant type is not supported by the authorization server."} 

Azure:
- wrong password
{"error":"invalid_grant","error_description":"AADSTS50126: Error validating credentials due to invalid username or password.\r\nTrace ID: <trace_id>\r\nCorrelation ID: <correlation id>\r\nTimestamp: 2020-02-18 11:48:00Z","error_codes":[50126],"timestamp":"2020-02-18 11:48:00Z","trace_id":"<trace id>","correlation_id":"<correlation id>","error_uri":"https://login.microsoftonline.com/error?code=50126"}

- unknown user:
{"error":"invalid_grant","error_description":"AADSTS50034: The user account {EmailHidden} does not exist in the <ME_CENSORED_THIS> directory. To sign into this application, the account must be added to the directory.\r\nTrace ID: <trace id>\r\nCorrelation ID: <correlation id>\r\nTimestamp: 2020-02-18 12:06:35Z","error_codes":[50034],"timestamp":"2020-02-18 12:06:35Z","trace_id":"<trace id>","correlation_id":"<correlation id>","error_uri":"https://login.microsoftonline.com/error?code=50034"}

- correct password
{"token_type":"Bearer","scope":"openid profile email","expires_in":3599,"ext_expires_in":3599,"access_token":"<token>","id_token":"<different_token>"}

- password grant on common/
{"error":"invalid_request","error_description":"AADSTS9001023: The grant type is not supported over the /common or /consumers endpoints. Please use the /organizations or tenant-specific endpoint.\r\nTrace ID: <trace id>\r\nCorrelation ID: <correlation id>\r\nTimestamp: 2020-02-18 11:57:01Z","error_codes":[9001023],"timestamp":"2020-02-18 11:57:01Z","trace_id":"<trace id>","correlation_id":"<correlation id>"}
```
I did not find a way to disable password auth for GitLab OIDC (I tried it with an unknown grant_type instead), and Google does not seem to allow this at all. 
oauth.com has a button to get a testing oauth client but it does not seem to work.
I never got response from the OpenShift Online client for my request for oauth clients, but I will try to get to them to at least test this PR since they requested it

stlaz:
@marun as for automated testing, I am not sure how we could go around that (without setting up our own OIDC provider).

I can share with you the configs that I used for Google, GitLab and GitHub for manual testing, but those are otherwiser private and rely on additional configuration on the OIDC side

marun:
> @marun as for automated testing, I am not sure how we could go around that (without setting up our own OIDC provider).

I'm not suggesting that this PR tackle automating OIDC testing if the amount of work is too great, but I think it may be worth surfacing this gap as tech debt so effort can be prioritized going forward to resolve it.

pkg/operator2/idp.go
```diff
@@ -216,7 +233,11 @@ func (c *authOperator) convertProviderConfigToIDPData(providerConfig *configv1.I
 	return data, nil
 }
 
-func (c *authOperator) discoverOpenIDURLs(issuer, key string, ca configv1.ConfigMapNameReference) (*osinv1.OpenIDURLs, error) {
+// discoverOpenIDConfig retrieves basic information about an OIDC server with hostname
```
marun: nit: This comment does not appear to match the function signature, which lacks a boolean return value.

pkg/operator2/idp.go
```diff
@@ -292,6 +313,77 @@ func (c *authOperator) transportForCARef(ca configv1.ConfigMapNameReference, key
...
+	// prepare the request, require application/json for simple parsing the response
+	req, err := http.NewRequest("POST", tokenURL, body)
+	if err != nil {
+		return false, err
+	}
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Accept", "application/json")
+
+	reqBytez, err := httputil.DumpRequest(req, true)
+	klog.V(1).Infof("teh bytes be: %s", reqBytez)
```
marun: nit: s/teh/the

pkg/operator2/idp.go
```diff
@@ -292,6 +313,77 @@ func (c *authOperator) transportForCARef(ca configv1.ConfigMapNameReference, key
...
+	query := url.Values{}
+	query.Add("client_id", clientID)
+	query.Add("client_secret", string(clientSecret))
+	query.Add("grant_type", "password")
+	query.Add("scope", "openid")
+	query.Add("username", "test")
+	query.Add("password", "test")
+	body := strings.NewReader(query.Encode())
+
+	// prepare the request, require application/json for simple parsing the response
```
marun: (No action required) nit: Maybe localize the comment about the format to where the Accept headers is set?

pkg/operator2/idp.go
```diff
@@ -292,6 +313,77 @@ func (c *authOperator) transportForCARef(ca configv1.ConfigMapNameReference, key
...
+	defer resp.Body.Close()
+
+	respJSON := json.NewDecoder(resp.Body)
+	respMap := map[string]interface{}{}
+	if err = respJSON.Decode(&respMap); err != nil {
+		return false, fmt.Errorf("failed to decode response from the OIDC server: %v", err)
+	}
+
+	if errVal, ok := respMap["error"]; ok {
+		oidcPasswordChecks[secret.ResourceVersion] = errVal == "invalid_grant" // wrong password, but password grants allowed
```
marun: So this error is the best way to discover that oidc provider supports password auth?

pkg/operator2/idp.go
```diff
@@ -292,6 +313,77 @@ func (c *authOperator) transportForCARef(ca configv1.ConfigMapNameReference, key
...
+	} else {
+		_, ok = respMap["access_token"] // in case we managed to hit the correct user
```
marun: Hmm, in case l/p test/test works?

I can see how this would be a good manual test for whether a password check is supported. Is this best practice for detecting oidc capability though? My naive expectation would be that it wouldn't be necessary to authenticate to discover what types of authentication were supported, but I'll openly admit that I don't know auth conventions.

pkg/operator2/idp.go
stlaz: I forgot to remove this, it was for testing purposes. I tend to amuse myself by misspelling "the" this way

pkg/operator2/idp.go
stlaz: I was looking for something like "yes, I would allow this, but give me correct credentials", which is what this error type seems to be used for in the tested OIDC providers.

pkg/operator2/idp.go
stlaz: > Hmm, in case l/p test/test works

Yes

> I can see how this would be a good manual test for whether a password check is supported. Is this best practice for detecting oidc capability though?

I originally thought that we could simply depend on the discovery endpoint, specifically on the `grant_types_supported` field, but this only seems to be set by Google, neither GitLab nor Azure expose the field, GitHub does not have the endpoint at all. And even if this field was set in discovery, the password grant flow can still be disabled on some oauth clients.

I think that leaves us in the uncomfortable position for having to check for the flow ourselves :disappointed:

pkg/operator2/idp.go
stlaz: Btw - this would log the client secret so we definitely do not want that

stlaz:
Addressed comments

stlaz:
/retest
rate limit exceeded on AWS

stlaz:
that was a bit of a premature push... Anyway - added a first part of e2e test, probably not gonna work, I haven't even run it yet but it's enough code to worry about losing

Also fixed a bug that would appear if request header IdP was configured in challenge mode, although it's not all that nice at this moment.

stlaz:
The basic "set up my own instance of gitlab and try password grant check against it" now finally works. Made me discover a bug in the original code.

stlaz:
/test e2e-aws
storage exploded

stlaz:
/hold cancel
I would like this to still go to 4.5, so I propose that I'd open a 4.6 blocking BZ for an e2e test for this with Keycloak if the reviewer would be ok with this.

Also, I removed the challenge-redirect IDP check, during testing I noticed we don't allow OIDC and RequestHeader IDP both to be set, so that should be somewhat safe - https://github.com/openshift/origin/blob/afc841de5012eb1cb721be2ce6c70ea4c584ae9f/vendor/k8s.io/kubernetes/openshift-kube-apiserver/admission/customresourcevalidation/oauth/validate_idp.go#L66

test/library/waits.go
```diff
@@ -71,3 +78,49 @@ func conditionsStatusString(conditions []configv1.ClusterOperatorStatusCondition
...
+}
+
+func WaitForHTTPStatus(t *testing.T, waitDuration time.Duration, client *http.Client, targetURL string, expectedStatus int) error {
+	t.Logf("waiting for HEAD at %q to report %d", targetURL, expectedStatus)
+
+	loggingReductionCounter := 0
+	return wait.PollImmediate(time.Second, waitDuration, func() (bool, error) {
+		resp, err := client.Head(targetURL)
+		if err != nil {
+			return false, nil
```
marun, line 114: Are you intentionally not logging the error? In the event of misconfiguration, I worry that the lack of detail could complicate troubleshooting.

test/library/waits.go
```diff
@@ -71,3 +78,49 @@ func conditionsStatusString(conditions []configv1.ClusterOperatorStatusCondition
...
+		}
+		if resp.StatusCode == expectedStatus {
+			return true, nil
+		}
+
+		if (loggingReductionCounter % 10) == 0 { // only log failure once in 10 seconds
```
marun, line 120: Given that there could be a range of potential status other than that expected, is it safe assume this log message will be representative of the intervening failures?

test/library/operand.go
```diff
@@ -0,0 +1,44 @@
...
+	config, err := cmClient.ConfigMaps("openshift-authentication").Get(context.TODO(), "v4-0-config-system-cliconfig", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	rawConfig := config.Data["v4-0-config-system-cliconfig"]
+
+	configObj := &osinv1.OsinServerConfig{}
+	if err := json.NewDecoder(bytes.NewBuffer([]byte(rawConfig))).Decode(configObj); err != nil {
+		fmt.Printf("decoded object dump: %s", configObj)
```
marun, line 27: I'm assuming this line doesn't belong? In the event of an error, `configObj` is unlikely to be populated.

test/library/operand.go
```diff
@@ -0,0 +1,44 @@
...
+		return nil, fmt.Errorf("failed to decode object: %v", err)
+	}
+
+	return configObj, nil
+}
+
+// GetIDPByName returns a pointer to the identity provider of a name from the
+// config.
+// Do not modify the returned value, it would modify the config value as well
```
marun, line 36: Would it make sense to make a copy to avoid burdening the caller with this restriction?

test/library/idpdeployment.go
```diff
@@ -0,0 +1,209 @@
...
+				Name:   namespace,
+				Labels: CAOE2ETestLabels(), // add label to easily remove/get the NS by hand
+			},
+		},
+		metav1.CreateOptions{},
+	)
+	require.NoError(t, err)
+
+	cleanup = func() {
+		// remove the NS, it will take away all the resources create here along with it
```
marun, line 54: (No action required) Is this comment telling me anything I wouldn't know otherwise?

test/library/idpdeployment.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+}
+
+func CleanIDPConfigByName(t *testing.T, configClient configv1client.OAuthInterface, idpName string) {
+	config, err := configClient.Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("cleanup: failed to retrieve oauth/cluster: %v", err)
+	}
+
+	i := 0
```
marun, line 166: (No action required) Where using an index variable outside of a loop, consider avoiding single-character names and using something more descriptive for clarity (e.g. `indexOfTargetIDP`).

test/library/idpdeployment.go
```diff
@@ -0,0 +1,209 @@
...
+	for _, idp := range config.Spec.IdentityProviders {
+		if idp.Name == idpName {
+			break
+		}
+		i++
+	}
+
+	// did not find the idp of name
```
marun, line 174: (No action required) Where possible, consider using a boolean conditional whose name documents your intent in preference to a statement that requires a comment e.g.

```golang
notFound := i == len(config.Spec.IdentityProviders)
if notFound {
  return
}
```

test/library/gitlabidp.go
```diff
@@ -0,0 +1,316 @@
...
+			{
+				Name:      "configdir",
+				SubPath:   "gitlab/data",
+				MountPath: "/var/opt/gitlab",
+			},
+		},
+		corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				// it's a bit resource-heavy, it will probably eat more during start-up
+				// max observed cpu: 1420m
```
marun, line 77: Yikes!

test/library/gitlabidp.go
```diff
@@ -0,0 +1,316 @@
...
+			for _, c := range cleanups {
+				c()
+			}
+		}
+	}()
+	require.NoError(t, err)
+
+	gitlabClient := GitLabClientFor(t, transport, gitlabURL)
+
+	// TODO: future us might try to configure whitelist IPs so that we can configure
```
marun, line 101: (No action required) Is there a convention to ensure ownership of TODOs e.g. `TODO(marun)`? I recall this from upstream but maybe this isn't common downstream.

test/library/gitlabidp.go
```diff
@@ -0,0 +1,316 @@
...
+			},
+			Data: map[string][]byte{
+				"clientSecret": []byte(appSecret),
+			},
+		},
+		metav1.CreateOptions{},
+	)
+	require.NoError(t, err, "failed to create gitlab client secret")
+	cleanups = append(cleanups, func() {
+		kubeClients.CoreV1().Secrets("openshift-config").Delete(context.TODO(), "gitlab-secret", metav1.DeleteOptions{})
```
marun, line 132: Log on failure?

Separately, what are your thoughts on introducing a static checker to catch these kinds of mechanical errors (e.g. implicitly ignored return values)?

pkg/operator2/idp.go
```diff
@@ -293,6 +321,75 @@ func (c *authOperator) transportForCARef(ctx context.Context, ca configv1.Config
...
+	query.Add("username", "test")
+	query.Add("password", "test")
+	body := strings.NewReader(query.Encode())
+
+	req, err := http.NewRequest("POST", tokenURL, body)
+	if err != nil {
+		return false, err
+	}
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	// explicitly set Accept to 'application/json' as that's the expected deserializable output
```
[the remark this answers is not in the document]
stlaz: > Would the oidc server return something different otherwise?
it might, IIRC some servers defaulted to urlencoded

test/library/certs.go
```diff
@@ -93,3 +98,23 @@ func NewCertificateAuthorityCertificate(t *testing.T, parent *CryptoMaterials) *
...
+// syncDefaultIngressCA sychronizes the openshift-config-managed/default-ingress-cert
+// wherever needed
+// useful when deploying an IDP behind a reencrypt/edge-termination route
+// returns a cleanup function for the CM
+func SyncDefaultIngressCAToConfig(t *testing.T, cmClient corev1client.ConfigMapsGetter, name string) func() {
+	ca, err := cmClient.ConfigMaps("openshift-config-managed").Get(context.TODO(), "default-ingress-cert", metav1.GetOptions{})
+	require.NoError(t, err)
+
+	ca.ObjectMeta = metav1.ObjectMeta{Name: name, Labels: CAOE2ETestLabels()}
+	ca.Data["ca.crt"] = ca.Data["ca-bundle.crt"]
```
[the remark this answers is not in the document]
stlaz: We expect the "ca.crt" field in our IdP configs, it's described in the API and enforced in configuration handling as in https://github.com/openshift/cluster-authentication-operator/blob/0f699e19a878b4c0140caca2be311ed3b3c63c24/pkg/operator2/idp.go#L178

Whereas most of OpenShift chose "ca-bundle.crt", we went with the kube's "ca.crt" instead back then ("ca.crt" is used by Kube in SA-token secrets, for example).

test/library/operand.go
stlaz: this was probably something else originally I used during debugging and I removed wrong lines, I'll fix that

test/library/idpdeployment.go
stlaz: it's a "don't panic and don't try to needlessly set up cleanups for the other resources" kind of comment, I like to remind myself of obvious stuff every now and then

pkg/operator2/idp.go
```diff
@@ -217,6 +242,9 @@ func (c *authOperator) convertProviderConfigToIDPData(ctx context.Context, provi
 	return data, nil
 }
 
+// discoverOpenIDURLs retrieves basic information about an OIDC server with hostname
+// given by the `issuer` argument. It returns the authentication, token and user
+// info urls in *osinv1.OpenIDUrls structure
```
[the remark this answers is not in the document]
stlaz: I guess that's just an old habit from python where it's not always that obvious, I'll remove that part

pkg/operator2/idp.go
```diff
@@ -293,6 +321,75 @@ func (c *authOperator) transportForCARef(ctx context.Context, ca configv1.Config
...
+	if !ok || len(clientSecret) == 0 {
+		return false, fmt.Errorf("the referenced secret does not contain the 'clientSecret' key")
+	}
+
+	transport, err := c.transportForCARef(ctx, caRererence, corev1.ServiceAccountRootCAKey)
+	if err != nil {
+		return false, fmt.Errorf("couldn't get a transport for the referenced CA: %v", err)
+	}
+
+	// prepare the grant-checking query, "openid" scope should be enough
```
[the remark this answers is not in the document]
stlaz: it's the minimal scop we always required, I'll improve the comment

test/library/waits.go
stlaz: will add logging here

test/library/waits.go
stlaz: I'll improve this to only log whenever status changes

test/library/idpdeployment.go
stlaz: I usually stick just with comments to avoid unnecessary variable allocation

test/library/gitlabidp.go
stlaz: I've always thought of TODOs as of useful ideas if you want to improve parts of the code but don't currently have time/willingness to deal with that, and anyone can do that if they find time for it.

If there was a TODO that I would like to own as I really just hacked around something and want to fix that soon, I would go with TODO(owner) as you suggest.

test/library/gitlabidp.go
stlaz: Will add the logging.

I haven't used static checkers in golang too much, outside of what my IDE provides at least, I think that they make much more sense for scripting languages, whereas go with its static typing and strong pressure on the code structure may not need it that much, and could be more of a nuisance than help (mostly with adding error ignoring comments).

pkg/operator2/idp.go
```diff
@@ -32,6 +33,11 @@ var (
 	scheme  = runtime.NewScheme()
 	codecs  = serializer.NewCodecFactory(scheme)
 	encoder = codecs.LegacyCodec(osinv1.GroupVersion) // TODO I think there is a better way to do this
+
+	// save latest OIDC password grant check not to bomb the provider with
+	// login requests each sync loop
+	// map is checkedSecretResourceVersion -> passwordGrantsAllowed
+	oidcPasswordChecks = map[string]bool{}
```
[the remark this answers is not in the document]
stlaz: The worst case scenario is n-times accessing the remote OIDC server in a short interval, where n is the number of workers and I'd expect (n << 10) and the map being written the same value n-times.

I do not see how anyone could trivially add more than one worker for the purpose of what this is doing :thinking: I can add a comment to the map that anyone using it needs to be careful that accessing it is not thread-safe, I'd expect anyone trying to do that would need to tear it off the operator first anyway, so it might be a better spot to comment here?

test/library/certs.go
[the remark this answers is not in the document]
marun: Maybe add a comment reflecting this so that future maintainers can discover your intent?

test/library/idpdeployment.go
marun: I encourage you to avoid this kind of premature optimization. Code is for people first and foremost, and where optimization is required it generally requires performance analysis to determine bottlenecks (e.g. hot loops or stale references).

test/library/gitlabidp.go
marun: While I share your concern at poorly configured tools causing problems, I've seen some positive outcomes. Most recently in openshift/kubernetes, enabling upstream static analysis checks that we previously had disabled revealed a number of mechanical gotchas (similar to the one here) in our persistent carries. A good tool supports disabling checks that are not useful for a codebase. I think at a minimum ensuring that return values need to be explicitly ignored (e.g. `_, _ := myfunc()` vs `myfunc()`) would be worth enabling since it is such a common cause of accidentally ignoring errors.

marun:
/lgtm

Thank you for cleanup!

For large PRs (and I think this qualifies), what do you think of using fixup commits to simplify follow-up reviews?

stlaz:
You're correct, probably should have done that and squash them before merge, hopefully I'll remember that the next time

test/library/idpdeployment.go
stlaz: Yet comments are there for the people and not the machine, and if the variable is not reused anywhere, a comment still makes more sense to me than a new variable

stlaz:
The BZ for adding Keycloak e2e: https://bugzilla.redhat.com/show_bug.cgi?id=1841039

PR #250, opened 2020-02-11
 move to openshift/build-machinery-go 

stlaz:
/cc @vareti @marun


stlaz:
this also fixes the test failure from https://github.com/openshift/release/pull/7115 (needed to do `git add -f` on a file that's gitignored inside one of the vendored deps)

marun:
@stlaz I this for 4.4 or 4.5?

PR #251, opened 2020-02-21
fix gosum

stlaz:
somehow my private repo got to go.sum, fixing this should hopefully finally make it possible to run verify-deps to prevent this in the future


stlaz:
/retest
CI might be broken?

PR #252, opened 2020-02-28
The oauth server should wait until it is out of rotation to shut down

smarterclayton:
When a pod is marked deleted, endpoints are updated instantly and
propagate to load balancers, routers, and nodes. A component that
wishes to remain available during upgrades must wait longer than
the default propagation interval for these changes to avoid having
requests delivered to pods that are shutting down.

Change the oauth server to wait 25s before terminating the serving
process and up to 40s on the node to ensure all front ends have
time to drain. The minimum interval here is how long an average
connection can take behind the router to drain, once new connections
stop getting created. I.e.

    wait = time to propagate endpoints (5s) +
           time for router reload (5s) +
           time for longest request to finish (15s)
         = 25s

This should result in the oauth route not reporting any downtime
during upgrade.

Downtime documented in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade/19447

### DIFF
--- a/bindata/oauth-openshift/deployment.yaml
+++ b/bindata/oauth-openshift/deployment.yaml
@@ -17,6 +17,7 @@ spec:
       labels:
         app: oauth-openshift
     spec:
+      terminationGracePeriodSeconds: 40
       serviceAccountName: oauth-openshift
       nodeSelector:
         node-role.kubernetes.io/master: ''
@@ -102,6 +103,16 @@ spec:
             periodSeconds: 10
             successThreshold: 1
             failureThreshold: 3
+          lifecycle:
+            # delay shutdown by 25s to ensure existing connections are drained
+            # * 5s for endpoint propagation on delete
+            # * 5s for route reload
+            # * 15s for the longest running request to finish
+            preStop:
+              exec:
+                command:
+                - sleep
+                - "25"
           terminationMessagePolicy: FallbackToLogsOnError
           resources:
             requests:
--- a/pkg/operator2/assets/bindata.go
+++ b/pkg/operator2/assets/bindata.go
@@ -64,6 +64,7 @@ spec:
       labels:
         app: oauth-openshift
     spec:
+      terminationGracePeriodSeconds: 40
       serviceAccountName: oauth-openshift
       nodeSelector:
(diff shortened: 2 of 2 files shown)


smarterclayton:
See https://github.com/openshift/console-operator/pull/385 as well (these are the two routes we monitor)

smarterclayton:
/hold

while testing

sttts:
Is it still the case that kube-proxy might get an iptables conflict and try another time 30 sec later? cc @squeed

smarterclayton:
Router hits oauth service directly over pod network, so iptables changes cannot affect it.  It could affect the router, but the router waits much longer.

smarterclayton:
Verified during upgrade route ingresses remain available https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade/19806 when testing all three PRs together.

This is ready for review.

sttts:
/lgtm
/approve

Any reason not to cherry-pick this into 4.4, 4.3 and 4.2 ?

smarterclayton:
4.4 bug is 1809667 and 4.3 bug is 1809668

smarterclayton:
4.3 backport is https://github.com/openshift/cluster-authentication-operator/pull/257

bindata/oauth-openshift/deployment.yaml
```diff
@@ -102,6 +103,16 @@ spec:
             periodSeconds: 10
             successThreshold: 1
             failureThreshold: 3
+          lifecycle:
+            # delay shutdown by 25s to ensure existing connections are drained
+            # * 5s for endpoint propagation on delete
+            # * 5s for route reload
+            # * 15s for the longest running request to finish
```
sttts, line 110: When the pod is deleted, its endpoint is removed and we give it 5s to propagate. Existing connections keep serving, especially http/2 and keepalive ones. *New* requests will keep coming all these 25s, until the process gets a signal and it closes the connection.

Afaik we disallow http/2. What about keepalive, i.e. http connection reuse?

Does the CI monitor test force new connections for each request?

PR #253, opened 2020-03-02
Bug 1809193: Resync auth operator informer every 30 sec to tighten sync loop for .well-known check

sttts:
The old resync period was 20 min, far too long to reliably check the .well-known endpoint.


mfojtik:
/lgtm

If this passes the tests.

PR #257, opened 2020-03-03
Bug 1824167: The oauth server should wait until it is out of rotation to shut down

smarterclayton:
When a pod is marked deleted, endpoints are updated instantly and
propagate to load balancers, routers, and nodes. A component that
wishes to remain available during upgrades must wait longer than
the default propagation interval for these changes to avoid having
requests delivered to pods that are shutting down.

Change the oauth server to wait 25s before terminating the serving
process and up to 40s on the node to ensure all front ends have
time to drain. The minimum interval here is how long an average
connection can take behind the router to drain, once new connections
stop getting created. I.e.

    wait = time to propagate endpoints (5s) +
           time for router reload (5s) +
           time for longest request to finish (15s)
         = 25s

### DIFF
--- a/pkg/operator2/deployment.go
+++ b/pkg/operator2/deployment.go
@@ -98,6 +98,7 @@ func defaultDeployment(
 	klog.V(4).Infof("tracked resource versions: %s", rvs)
 	rvsHash := sha512.Sum512([]byte(rvs))
 	rvsHashStr := base64.RawURLEncoding.EncodeToString(rvsHash[:])
+	gracePeriod := int64(40)
 
 	// make sure ApplyDeployment knows to update
 	meta := defaultMeta()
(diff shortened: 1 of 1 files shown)


stlaz:
/hold
the failing job should not even be triggered for this branch, holding to stop the CI spam

bbguimaraes:
The [failed job](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/256/pull-ci-openshift-cluster-authentication-operator-master-verify-deps/17) is from another PR (#256), seems like a bug in Prow.

/skip

stlaz:
/retitle [4.3] Bug 1824167: The oauth server should wait until it is out of rotation to shut down

eparis:
Bugs were self verified, no actual QA has happened. We can not merge like this. QA needs to happen before we try to merge to z-streams. If you need help understanding why we have such processes feel free to talk to your manager, team lead, or group lead.

sdodson:
/retitle Bug 1824167: The oauth server should wait until it is out of rotation to shut down

sdodson:
4.4 BZ has been properly verified and shipped via errata, approving.

PR #243, opened 2020-01-20
workload controller for CAO

p0lyn0mial:
it's based on the workload controller in `cas-o` except it
- tries to be generic at least in (CAO)
- doesn't take `operator.Generation` into account because `CAO` will manage multiple operands and that would make them compete.

TODO:
- [x] add config support
- [x] add unit tests
- [x] wire into `CAO`
- [x] create a sperate package for each operator
- [x] use static controller (https://github.com/openshift/cluster-openshift-apiserver-operator/pull/304)
- [x] create a generic workload controller and reuse it in this operator and in `openshif-apiserver-operator`
- [x] figure out how to wire the serving cert because at the moment I see in the logs `serving.go:306] Generated self-signed cert (apiserver.local.config/certificates/apiserver.crt, apiserver.local.config/certificates/apiserver.key)`
- [x] add revision support (for etcd encryption)
- [x] wire `APIServiceController` https://github.com/openshift/cluster-authentication-operator/pull/260
- [x] move the prune controller to APIServerControllerSet (library-go) (https://github.com/openshift/library-go/pull/846)
- [x] wire the config observers (https://issues.redhat.com/browse/MSTR-984)
- [x] add proxy support https://issues.redhat.com/browse/MSTR-986 (not needed)
- [x] add standard encryption tests for CAO (@p0lyn0mial ) (requires https://github.com/openshift/oauth-apiserver/pull/14)
- [x] investigate why `openapi/v2` endpoint spits back only `2` paths. (@stlaz ) (https://github.com/openshift/oauth-apiserver/pull/15)
- [x] add t[he finalizer controller](https://github.com/openshift/cluster-openshift-apiserver-operator/blob/487fef099d1dfdb1c47eea85b513c3dfdf4a449a/pkg/operator/nsfinalizercontroller/finalizer_controller.go#L40-L46)
- [x] CAO in 4.5 flips `Status.ManagingOAuthAPIServer` to `false` (OAS-O in 4.5/4.6 is ready to handle that) (@stlaz https://github.com/openshift/cluster-authentication-operator/pull/307)
- [ ] what about https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/audit-policy.md ?
- [x] figure out how to react to changes in the deployment's spec. For example when the flags changes or were changed by the user. (@sttts) (https://github.com/openshift/library-go/pull/856)
[...]

PR #260, opened 2020-03-13
testing WithAPIServiceController for CAO

p0lyn0mial:


p0lyn0mial:
builds on https://github.com/openshift/cluster-authentication-operator/pull/243 and adds https://github.com/openshift/cluster-authentication-operator/pull/260/commits/281af092ec594c0c55b413b20f2fc04127a333dc which makes `OpenShiftAPIServer` to stop managing `OAuth` related `APIs`

pkg/operator2/apiservices/sync_take_over_apiservices_oauth_apiserver.go
```diff
@@ -0,0 +1,50 @@
...
+
+	controllerFactory := factory.New()
+	authOperatorLister := authOperatorInformers.Operator().V1().Authentications().Lister()
+
+	controllerFactory.WithSync(func(ctx context.Context, controllerContext factory.SyncContext) error {
+		operator, err := authOperatorLister.Get("cluster")
+		if err != nil {
+			return err
+		}
+		if !v1helpers.IsOperatorConditionTrue(operator.Status.Conditions, "Available") {
```
p0lyn0mial: should we also check if were are `NOT` progressing and we are `NOT` degraded?

pkg/operator2/apiservices/sync_take_over_apiservices_oauth_apiserver.go
```diff
@@ -0,0 +1,50 @@
...
+			message := "authentication operator is not Available"
+			controllerContext.Recorder().Warning("PrereqNotReady", message)
+			return fmt.Errorf(message)
+		}
+
+		if !operator.Status.ManagingOAuthAPIServer {
+			operatorCopy := operator.DeepCopy()
+			operatorCopy.Status.ManagingOAuthAPIServer = true
```
p0lyn0mial: @sttts if we decide to land it in `4.5` we would have to flip it to `false` in `4.4` (downgrade scenario). So that `OAS` can claim all `groups`

p0lyn0mial: unless we need to add it `OAS-O` for `4.4` because it runs before `CAO`?

p0lyn0mial: so, yeah `OAS-O` is run before `CAO`, so I think we need a way to tell `CAO` to step down otherwise they will be fighting with yeah other, am I right?

sttts: We need a way to decide the fight for OAS to win as long as it is the latest version (don't forget rollback when there is an old OAS again).

p0lyn0mial: yeah, I am asking how it will behave on downgrades. As I understand it, we will have an old `OAS-O` and new `CAO`. Because `OAS-O` runs on a different runlevel that `CAO`

I think we need a way to claim back the API as soon as we will successfully roll back `OAS-O` and we need to tell the new `CAO` to step back?

Otherwise, they will be fighting until `CAO` is rolled back.

PR #268, opened 2020-04-10
Bug 1822666: authentication operator should handle operand namespace creation

sallyom:
Migrated operand ns and dependent resources to auth-operator managed rather than cvo.  Also, I added the yaml check that openshift-apiserver has. 
 This PR moves from manifest to bindata:
* openshift-authentication ns yaml
* operand clusterrolebinding yaml
* operand serviceaccount yaml
* branding-secret yaml 
* operand cabundle configmap yaml


pkg/operator2/operator.go
```diff
@@ -122,11 +126,12 @@ func NewAuthenticationOperator(
 
 		oauthClientClient: oauthClientClient.OAuthClients(),
 
-		services:    kubeClient.CoreV1(),
-		endpoints:   kubeClient.CoreV1(),
-		secrets:     kubeClient.CoreV1(),
-		configMaps:  kubeClient.CoreV1(),
-		deployments: kubeClient.AppsV1(),
+		kubeClientHolder: resourceapply.NewKubeClientHolder(kubeClient),
```
stlaz, line 129: you'll want to wire in the informers for the new resources

pkg/operator2/operator.go
```diff
@@ -239,6 +244,19 @@ func (c *authOperator) handleSync(ctx context.Context, operatorConfig *operatorv
 	// will cause a redeploy anyway
 	// TODO move this hash from deployment meta to operatorConfig.status.generations.[...].hash
 	resourceVersions := []string{}
+	// ensure operand namespace and resources exist
+	directResourceResults := resourceapply.ApplyDirectly(c.kubeClientHolder, c.recorder, assets.Asset,
+		"oauth-openshift/ns.yaml",
+		"oauth-openshift/cabundle.yaml",
```
stlaz, line 250: nit: remove the `release.openshift.io/create-only: "true"` from the file, it's no longer needed when you do this

sallyom: done

pkg/operator2/operator.go
sallyom: the configmap added is the `v4-0-config-system-trusted-ca-bundle` and the branding secret is `v4-0-config-system-ocp-branding-template` so those are covered by the `getPrefixFilter()` function, right? https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator2/operator.go#L170-#L171

pkg/operator2/operator.go
sallyom: done

pkg/operator2/operator.go
stlaz: yep, those should already be covered :)

bindata/oauth-openshift/authentication-clusterrolebinding.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:openshift:openshift-authentication
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin  # TODO fix, this is madness
```
stlaz, line 7: could you please change the comment to: oauth-server needs to read secrets in all namespaces so there is no point trying to tighten its permissions

pkg/operator2/operator.go
stlaz: but you are still missing an informer for the NS of the operand, the rolebinding and the SA

bindata/oauth-openshift/authentication-clusterrolebinding.yaml
sallyom: done

pkg/operator2/operator.go
```diff
@@ -161,11 +165,14 @@ func NewAuthenticationOperator(
...
 	configNameFilter := operator.FilterByNames("cluster")
 	prefixFilter := getPrefixFilter()
+	operandNSFilter := operator.FilterByNames("openshift-authentication")
 
 	return operator.New("AuthenticationOperator2", c,
 		operator.WithInformer(routeInformer, targetNameFilter),
 		operator.WithInformer(coreInformers.Services(), targetNameFilter),
 		operator.WithInformer(kubeInformersNamespaced.Apps().V1().Deployments(), targetNameFilter),
+		operator.WithInformer(kubeInformersNamespaced.Core().V1().ServiceAccounts(), targetNameFilter),
+		operator.WithInformer(kubeInformersNamespaced.Core().V1().Namespaces(), operandNSFilter),
```
sallyom, line 176: I added informer for sa and ns, but looking at our other operators (i looked at service-ca-operator, openshift-apiserver-operator, cluster-kube-controller-manager-operator), AFAICT they don't add the informer for ClusterRoleBindings()  I want to ask why before I add it here.

stlaz: That's a good question, I do not know. I can see that those attempt to use kube-auto-reconciliation, which we could perhaps use here, too, I only need to see how it works.

pkg/operator2/operator.go
sallyom: added all but clusterrolebinding, mentioned in note below

bindata/oauth-openshift/serviceaccount.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: openshift-authentication
+  name: oauth-openshift
+  labels:
+    app: oauth-openshift
```
deads2k, line 7: why the label?

pkg/operator2/operator.go
sallyom: added staticresourcecontroller instead : )

PR #269, opened 2020-04-14
Bug 1817748: Fix ingress state condition flapping

marun:
Previously the primary auth controller set all condition types when performing an update. With the addition of the ingress state and router certs controllers, this behavior resulted in status message
flapping as the primary controller raced with the new controllers. The fix is having the primary controller avoid setting the condition types managed by the other controllers.

/cc @stlaz @mfojtik

### DIFF
--- a/pkg/operator2/oauth.go
+++ b/pkg/operator2/oauth.go
@@ -27,6 +27,7 @@ func (c *authOperator) handleOAuthConfig(
 	operatorConfig *operatorv1.Authentication,
 	route *routev1.Route,
 	service *corev1.Service,
+	conditions *authConditions,
 ) (
 	*corev1.ConfigMap,
 	*configSyncData,
@@ -83,7 +84,7 @@ func (c *authOperator) handleOAuthConfig(
 			},
 		)
 	}
-	handleDegraded(operatorConfig, "IdentityProviderConfig", v1helpers.NewMultiLineAggregate(errsIDP))
+	conditions.handleDegraded("IdentityProviderConfig", v1helpers.NewMultiLineAggregate(errsIDP))
 
 	cliConfig := &osinv1.OsinServerConfig{
 		GenericAPIServerConfig: configv1.GenericAPIServerConfig{
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -196,14 +196,14 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 
 	operatorConfigCopy := operatorConfig.DeepCopy()
 
-	syncErr := c.handleSync(context.TODO(), operatorConfigCopy)
+	conditions := newAuthConditions()
+	syncErr := c.handleSync(context.TODO(), operatorConfigCopy, conditions)
 	// this is a catch all degraded state that we only set when we are otherwise not degraded
 	globalDegradedErr := syncErr
-	const globalDegradedPrefix = "OperatorSync"
-	if isDegradedIgnoreGlobal(operatorConfigCopy, globalDegradedPrefix) {
+	if conditions.hasDegraded {
 		globalDegradedErr = nil // unset because we are already degraded for some other reason
 	}
-	handleDegraded(operatorConfigCopy, globalDegradedPrefix, globalDegradedErr)
+	conditions.handleDegraded("OperatorSync", globalDegradedErr)
 
 	if _, _, err := v1helpers.UpdateStatus(c.authOperatorConfigClient, func(status *operatorv1.OperatorStatus) error {
 		// store a copy of our starting conditions, we need to preserve last transition time
@@ -216,7 +216,7 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 		status.Conditions = originalConditions
 
 		// manually update the conditions while preserving last transition time
-		for _, condition := range operatorConfigCopy.Status.Conditions {
+		for _, condition := range conditions.conditions {
 			v1helpers.SetOperatorCondition(&status.Conditions, condition)
 		}
 
@@ -231,7 +231,7 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 	return syncErr
 }
 
-func (c *authOperator) handleSync(ctx context.Context, operatorConfig *operatorv1.Authentication) error {
+func (c *authOperator) handleSync(ctx context.Context, operatorConfig *operatorv1.Authentication, conditions *authConditions) error {
 	// resourceVersions serves to store versions of config resources so that we
 	// can redeploy our payload should either change. We only omit the operator
 	// config version, it would both cause redeploy loops (status updates cause
@@ -251,7 +251,7 @@ func (c *authOperator) handleSync(ctx context.Context, operatorConfig *operatorv
 	}
 
 	route, routerSecret, reason, err := c.handleRoute(ctx, ingress)
-	handleDegradedWithReason(operatorConfig, "RouteStatus", reason, err)
+	conditions.handleDegradedWithReason("RouteStatus", err, reason)
 	if err != nil {
 		return fmt.Errorf("failed handling the route: %v", err)
 	}
@@ -295,7 +295,7 @@ func (c *authOperator) handleSync(ctx context.Context, operatorConfig *operatorv
 		return fmt.Errorf("failed applying session secret: %v", err)
 	}
 
-	expectedCLIconfig, syncData, err := c.handleOAuthConfig(ctx, operatorConfig, route, service)
+	expectedCLIconfig, syncData, err := c.handleOAuthConfig(ctx, operatorConfig, route, service, conditions)
 	if err != nil {
 		return fmt.Errorf("failed handling OAuth configuration: %v", err)
 	}
@@ -370,7 +370,7 @@ func (c *authOperator) handleSync(ctx context.Context, operatorConfig *operatorv
 
 	klog.V(4).Infof("current deployment: %#v", deployment)
 
-	if err := c.handleVersion(ctx, operatorConfig, authConfig, route, routerSecret, deployment, ingress); err != nil {
+	if err := c.handleVersion(ctx, operatorConfig, authConfig, route, routerSecret, deployment, ingress, conditions); err != nil {
 		return fmt.Errorf("error checking current version: %v", err)
 	}
 
@@ -385,6 +385,7 @@ func (c *authOperator) handleVersion(
 	routerSecret *corev1.Secret,
 	deployment *appsv1.Deployment,
 	ingress *configv1.Ingress,
+	conditions *authConditions,
 ) error {
 	// Checks readiness of all of:
 	//    - route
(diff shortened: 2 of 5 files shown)


pkg/operator2/operator.go
```diff
@@ -217,6 +225,11 @@ func (c *authOperator) Sync(obj metav1.Object) error {
 
 		// manually update the conditions while preserving last transition time
 		for _, condition := range operatorConfigCopy.Status.Conditions {
+			// Avoid setting conditions that are managed by other
+			// controllers to avoid racing with them.
+			if c.ignoredConditionTypes.Has(condition.Type) {
+				continue
+			}
 			v1helpers.SetOperatorCondition(&status.Conditions, condition)
```
stlaz, line 233: This would just delete those conditions that you keep ignored instead, wouldn't it?

Besides, I like to think of controllers as of separate logical units. I think the right thing to do would be to have `handleSync` return a slice of conditions/conditionUpdateFunctions that you could just pass to `UpdateStatus`. It would just require a minor change to `handleDegraded` and passing those updates up. WDYT?

marun: > This would just delete those conditions that you keep ignored instead, wouldn't it?

I don't believe that is the case. `status` is retrieved from the informer, and includes all conditions set by all controllers. The code here sets/updates only those conditions managed by this controller, and does not delete/update conditions set by other controllers.

> Besides, I like to think of controllers as of separate logical units. I think the right thing to do would be to have `handleSync` return a slice of conditions/conditionUpdateFunctions that you could just pass to `UpdateStatus`. It would just require a minor change to `handleDegraded` and passing those updates up. WDYT?

What you are proposing is the strategy already used by the ingress state and router certs controllers. I investigated implementing it for the primary controller, but it looked far more invasive due to the need to plumb collection and propagation of update functions through the entire callchain. Given that a refactor of the operator is ongoing, does it make sense to introduce such a change as part of that refactor rather than in the fixing of this bug?

stlaz: > The code here sets/updates only those conditions managed by this controller, and does not delete/update conditions set by other controllers.

don't know what I was thinking...

> Given that a refactor of the operator is ongoing, does it make sense to introduce such a change as part of that refactor rather than in the fixing of this bug?

this is a rare chance we get to do the right thing now rather than to write it down and postpone it to somewhen, I think it's a good excuse to do it.

marun: It's going to be really invasive. Happy to do it as a follow-up, but I really think it's orthogonal to the fix proposed here and much riskier.

marun: Also - the fix proposed here is trivial to backport (and it should be backported to 4.4 and 4.3). I'm less convinced of the wisdom of backporting the fix you are proposing.

stlaz: Why would we have to backport it to 4.3 and 4.2 when ingress state controller was added in 4.4? It sure would not make sense to backport it all the way.

Also, the idea was to just make `handleDegraded` return a condition which could then bubble all the way up to `Sync`. I am apparently missing something?

marun:
@stlaz As requested the latest update ensures controller independence. I chose to follow the example of the ingress state controller (only set condition if present and default if not set).

pkg/operator2/oauth.go
```diff
@@ -83,7 +84,9 @@ func (c *authOperator) handleOAuthConfig(
 			},
 		)
 	}
-	handleDegraded(operatorConfig, "IdentityProviderConfig", v1helpers.NewMultiLineAggregate(errsIDP))
+	if len(errsIDP) > 0 {
```
stlaz, line 87: move the err != nil check back in the `degraded` method, possibly keep the `handleDegraded` name as that might be more appropriate due to that err handling, but I'll leave that up to you

pkg/operator2/status.go
```diff
@@ -1,75 +1,113 @@
...
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
-func handleDegraded(operatorConfig *operatorv1.Authentication, prefix string, err error) {
-	handleDegradedWithReason(operatorConfig, prefix, "Error", err)
+// Each prefix included in this set will have its condition defaulted
+// to false if not explicitly set to true.
+var degradedConditionPrefixes = sets.NewString(
```
stlaz, line 12: why is this a set?

pkg/operator2/status.go
```diff
@@ -1,75 +1,113 @@
...
+	"IdentityProviderConfig",
```
stlaz, line 13: please add the `Degraded` suffix to all, no need to concat strings later

pkg/operator2/status.go
```diff
@@ -1,75 +1,113 @@
...
-			operatorv1.OperatorCondition{
-				Type:    prefix + operatorv1.OperatorStatusTypeDegraded,
-				Status:  operatorv1.ConditionTrue,
-				Reason:  reason,
-				Message: err.Error(),
+// all returns the set of conditions representing the complete status
+// of the auth operator. If a condition was not set, it will be
+// returned in its default state (degraded -> false, progressing ->
+// false, available -> true).
+func (c *authConditions) all() []operatorv1.OperatorCondition {
```
stlaz, line 36: `func (c *authConditions) allUpdates() []v1helpers.UpdateStatusFunc {` so that in `Sync()` we can just do:

```golang
if _, _, err := v1helpers.UpdateStatus(c.authOperatorConfigClient, conditions.allUpdates()...)
```
instead of the unnecessary conditions copying magic we do now

pkg/operator2/status.go
```diff
@@ -1,75 +1,113 @@
...
+	allConditions := []operatorv1.OperatorCondition{}
+
+	// Collect degraded conditions
+	for _, prefix := range degradedConditionPrefixes.UnsortedList() {
+		conditionType := prefix + operatorv1.OperatorStatusTypeDegraded
+		if degraded := v1helpers.FindOperatorCondition(c.conditions, conditionType); degraded != nil {
+			allConditions = append(allConditions, *degraded)
```
stlaz, line 43: `append(allConditions, v1helpers.UpdateConditionFn(*degraded)`

and similar later on

pkg/operator2/status.go
```diff
@@ -1,75 +1,113 @@
...
+// of the auth operator. If a condition was not set, it will be
+// returned in its default state (degraded -> false, progressing ->
+// false, available -> true).
+func (c *authConditions) all() []operatorv1.OperatorCondition {
+	allConditions := []operatorv1.OperatorCondition{}
+
+	// Collect degraded conditions
+	for _, prefix := range degradedConditionPrefixes.UnsortedList() {
+		conditionType := prefix + operatorv1.OperatorStatusTypeDegraded
+		if degraded := v1helpers.FindOperatorCondition(c.conditions, conditionType); degraded != nil {
```
stlaz, line 42: **Consider**, no action required if you feel confident about the current course of action:

Only return the condition updates that were triggered throughout the course of sync loop execution and ignore their complement in the set of all controllers conditions.

Reason (by example):
checking route health fails on GET HTTP request, you switch RouteHealthDegraded on
KAS has a temporary blip so for a few next syncs you fail earlier than in the route health, turning RouteHealthDegraded to false, instead having just OperatorSyncDegraded be true. Route is still inaccessible.
KAS decides it's ok to be a good API server again, you flip OperatorSyncDegraded to false and RouteHealthDegraded back to true

Sure, a counterexample would be route goes healthy again and KAS stays degraded so we'd be reporting wrong again, but I still tend to like the first approach a bit more. It is also the same behavior we had before, whereas this would be a change.

pkg/operator2/status.go
marun: I purposely didn't chooses the approach you are suggesting because the current call to `UpdateStatus` needs to do more than update conditions.

marun: See: https://github.com/openshift/cluster-authentication-operator/blob/fix-ingress-state-flapping/pkg/operator2/operator.go#L206

pkg/operator2/status.go
marun: ditto

pkg/operator2/oauth.go
marun: I externalized the error check from `degraded` because 4 out of 6 calls need to handle the error:

- https://github.com/marun/cluster-authentication-operator/blob/fix-ingress-state-flapping/pkg/operator2/operator.go#L252
- https://github.com/marun/cluster-authentication-operator/blob/fix-ingress-state-flapping/pkg/operator2/operator.go#L398
- https://github.com/marun/cluster-authentication-operator/blob/fix-ingress-state-flapping/pkg/operator2/operator.go#L408
- https://github.com/marun/cluster-authentication-operator/blob/fix-ingress-state-flapping/pkg/operator2/operator.go#L418

Why is it preferable to duplicate the error checking for a minority of cases?

marun:
@stlaz As per our conversation, the latest iteration is a much simpler change limited to ensuring that only operator-managed conditions are set.

pkg/operator2/status.go
stlaz: If the correct link is https://github.com/marun/cluster-authentication-operator/blob/fix-ingress-state-flapping/pkg/operator2/operator.go#L206 then I still don't see what the problem is. You set the `OperatorSyncDegraded` to true or false and then can still call `if _, _, err := v1helpers.UpdateStatus(c.authOperatorConfigClient, conditions.allUpdates()...)`

Am I missing something?

marun: Sorry, my comment was made prior to our conversation on Friday and is no longer relevant. Resolving...

pkg/operator2/status.go
```diff
@@ -1,74 +1,69 @@
...
-			strings.HasSuffix(condition.Type, operatorv1.OperatorStatusTypeDegraded) &&
-			condition.Status == operatorv1.ConditionTrue {
-			return true
-		}
+func (c *authConditions) handleDegraded(prefix string, err error) {
+	c.handleDegradedWithReason(prefix, err, "Error")
+}
+
+func (c *authConditions) handleDegradedWithReason(prefix string, err error, reason string) {
+	c.hasDegraded = true
```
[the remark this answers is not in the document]
marun: Good catch! Done.

pkg/operator2/status.go
stlaz: I think there was a bit of a misunderstanding - I was fine with the `all()` funcs, I just wanted you to form it into `allFuncs()` so tha we can do a simple `UpdateStatus()` call with it as I pointed out in https://github.com/openshift/cluster-authentication-operator/pull/269#discussion_r410163230 which would remove our custom update status logic and just reused https://github.com/openshift/library-go/blob/ccc4bb27d0320da04e5d2af6be09f030440bb152/pkg/operator/v1helpers/helpers.go#L184-L190

stlaz:
/lgtm
I did not notice we were actually mutating the status we're updating in Sync, we may want to address that in a follow-up

PR #270, opened 2020-04-14
Add component to OWNERS

sttts:


OWNERS
```diff
@@ -4,3 +4,4 @@ approvers:
   - stlaz
   - sttts
   - deads2k
+components: apiserver-auth
```
pierreprinetti, line 7: The example in the email says "component" rather than "components". Are you sure it is plural?

PR #272, opened 2020-04-16
Bug 1827050: Update to use revised ApplyDeployment

marun:
The new method hashes the spec of the deployment which removes the need for forcing rollout.

/cc @deads2k @stlaz

### DIFF
--- a/pkg/operator2/deployment.go
+++ b/pkg/operator2/deployment.go
@@ -23,6 +23,7 @@ func defaultDeployment(
 	operatorConfig *operatorv1.Authentication,
 	syncData *configSyncData,
 	proxyConfig *configv1.Proxy,
+	bootstrapUserExists bool,
 	resourceVersions ...string,
 ) *appsv1.Deployment {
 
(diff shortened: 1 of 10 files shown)


marun:
@stlaz The library-go has merged. Updated and ready for review.

@vareti Adding you as a reviewer since this PR is touching code you added.

/cc @vareti

pkg/operator2/deployment.go
```diff
@@ -47,6 +48,11 @@ func defaultDeployment(
 	}
 	deployment.Spec.Template.Annotations[deploymentVersionHashKey] = rvsHashStr
 
+	// Ensure a rollout when the bootstrap user goes away
+	if bootstrapUserExists {
+		deployment.Spec.Template.Annotations["operator.openshift.io/bootstrap-user-exists"] = "true"
```
vareti, line 53: do you think we should name annotation to imply CAO?

vareti:
one single comment, otherwise /lgtm

pkg/operator2/deployment.go
stlaz: I don't think there's a need for that now, and this is not an API so we can change it anytime in the future.

nit: you could just `deployment.Spec.Template.Annotations["operator.openshift.io/bootstrap-user-exists"] = fmt.Sprintf("%v", bootstrapUserExists)` to avoid conditions

marun: Why is it desirable to avoid conditions? I think the absence of an annotation is as good a signal as one with a value of `"false"`.

stlaz:
/approve
looks good

marun:
/retitle Bug 1827050: Update to use revised ApplyDeployment

mfojtik:
/hold

I will update tomorrow morning.

marun:
Committed with #282

PR #273, opened 2020-04-17
Bug 1824800: explicitly set oauth-server container's root file system to writable

stlaz:
We are doing root write with the copying of the trusted cert bundle
so we unfortunately need the FS to be writable.

The SA running the pod has broad privileges and if there's a more
restrictive SCC with higher priority than what anyuid has, this
would make the container fail to start as it would choose to
default to non-writable root system.

### DIFF
--- a/bindata/oauth-openshift/deployment.yaml
+++ b/bindata/oauth-openshift/deployment.yaml
@@ -62,6 +62,8 @@ spec:
             - name: https
               containerPort: 6443
               protocol: TCP
+          securityContext:
+            readOnlyRootFilesystem: false # because of the `cp` in args
           volumeMounts:
             - name: v4-0-config-system-session
               readOnly: true
--- a/manifests/07_deployment.yaml
+++ b/manifests/07_deployment.yaml
@@ -36,6 +36,8 @@ spec:
           requests:
             memory: 50Mi
             cpu: 10m
+        securityContext:
+          readOnlyRootFilesystem: false # because of the `cp` in args
         volumeMounts:
         - mountPath: /var/run/configmaps/config
           name: config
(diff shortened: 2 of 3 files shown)


bindata/oauth-openshift/deployment.yaml
```diff
@@ -62,6 +62,8 @@ spec:
             - name: https
               containerPort: 6443
               protocol: TCP
+          securityContext:
+            readOnlyRootFilesystem: false # because of the `rm` in args
```
sttts, line 66: you mean the cp

stlaz: right, the cp does rm first by `-f` which I believe is the error we're seeing, let me fix that

stlaz: ...or I just confused myself thinking of the removal, anyway

stlaz: fixed

bindata/oauth-openshift/deployment.yaml
```diff
@@ -62,6 +62,8 @@ spec:
             - name: https
               containerPort: 6443
               protocol: TCP
+          securityContext:
+            readOnlyRootFilesystem: false # because of the `cm` in args
```
sttts, line 66: cp ;-)

stlaz: :facepalm: somebody kill me please

stlaz:
/hold
need to fix the operator, too, I shouldn't start with fixes in late afternoons

bindata/oauth-openshift/deployment.yaml
```diff
@@ -62,6 +62,8 @@ spec:
             - name: https
               containerPort: 6443
               protocol: TCP
+          securityContext:
+            readOnlyRootFilesystem: false # because of the `cp` in args
```
sttts, line 66: @smarterclayton how do you feel about this? Any workaround? Note: we copy the proxy CA to /etc.

PR #276, opened 2020-04-21
Bug 1834852: Add make target to run operator with telepresence

marun:
**NOTE** https://github.com/openshift/build-machinery-go/pull/14 needs to merge before this PR.

Once telepresence has been installed (see the run-telepresence.sh script for instructions), `make telepresence` will replace the operator's pod with a local process that has access to the same environment as the pod. This can speed up development since iteration doesn't require cycles of image build/push/pull. See https://telepresence.io for more details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ origin.iml
 tags
 .envrc
 authentication-operator
+telepresence.log
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	targets/openshift/deps-gomod.mk \
 	targets/openshift/images.mk \
 	targets/openshift/bindata.mk \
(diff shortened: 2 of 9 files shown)


stlaz:
I assume this might be related to my custom setup, but it would appear that when the cluster-deployed operator tries to connect together with the local telepresence client, this never really happens.

I can see the local telepresence instance asking me for password to SSH to telepresence@127.0.0.1, but the prompt is not really interactive as the script isn't, and I can see the remote pod logging its failed attempts:

```
2020-05-13T12:53:26+0000 [Poll#error] Failed to contact Telepresence client:
2020-05-13T12:53:26+0000 [Poll#error] Connection was refused by other side: 111: Connection refused.
2020-05-13T12:53:26+0000 [Poll#warn] Perhaps it's time to exit?
```

stlaz:
/lgtm

Turns out I was using a wrong telepresence version (kudos to @marun), this works well :+1:

PR #279, opened 2020-04-22
Bug 1826341: ingress controller does not closes its connection to healthz

stlaz:


pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -310,7 +313,7 @@ func checkAddresses(ctx context.Context, addresses []corev1.EndpointAddress, che
 			}
 			messages := checkPod(ctx, address.TargetRef)
 			if len(messages) > 0 {
-				unhealthyPodCount += 1
+				unhealthyPodCount++
```
marun, line 316: Unrelated?

pkg/controller/ingressstate/ingress_state_controller.go
```diff
@@ -240,6 +240,9 @@ func checkEndpointHealthz(endpointIP string, rootCAs *x509.CertPool) error {
 	}
 	url := fmt.Sprintf("https://%s/healthz", net.JoinHostPort(endpointIP, "6443"))
 	response, err := client.Get(url)
+	if response != nil {
```
sttts, line 243: compare with 

```
resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
```

from https://golang.org/pkg/net/http/. Let's not risk a nil `response.Body`.

pkg/controller/ingressstate/ingress_state_controller.go
stlaz: my IDE underlines that which is just bothersome, I marked that as a cosmetic change in a separate commit, hopefully that's ok

stlaz:
s/io/ioutil, my bad

PR #290, opened 2020-06-25
pkg/operator2/operator: Check client and deployment operands each time

wking:
Builds on #289; you may want to review that first.

Before this commit, the auth operator checked a few operands sequentially. That means that an earlier error like `RouteHealthDegraded` would leave later operands unchecked (e.g. still reporting `ProgressingOAuthServerDeploymentNotReady` despite the deployment being ready).  This commit pivots to checking each operand each time, with the one exception being that we still skip the well-known endpoint check if the route is not ready, because a ready-route is a prerequisite for a ready well-known endpoint.  The clients and deployment, on the other hand, do not depend on a healthy route or ready well-known endpoint.

I'm also pushing the `oauth-openshift` version setting down into `checkDeploymentReady`, because we can level that as soon as the deployment is ready, without blocking on the other operands.

Setting the operator version, clearing `Progressing`, and ensuring `Available=True` still require all operands to be ready, just as they did before this commit.

### DIFF
--- a/pkg/operator2/operator.go
+++ b/pkg/operator2/operator.go
@@ -20,6 +20,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


wking:
[Failed on storage][1].

/retest

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/290/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws/1276391762310467584

stlaz:
/close
this is now obsolete, the checks have been distributed among other controller so that they are performed each time

PR #292, opened 2020-06-26
Refactor metadata to separate controller (part 1)

mfojtik:
This holds a ground work for splitting the main operator loop into more granular controllers based on controller factory.

To preserve order (temporarely), the operator progressing conditions are used. Note that this will eventually go away after all blocks are split (in this PR the first blocks is separated to its own controller).


mfojtik:
/hold

stop retesting this.

mfojtik:
/hold cancel

This is green and provide ground work.

mfojtik:
storage tests flaking

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -0,0 +1,397 @@
...
+
+// NewMetadataController assure that ingress configuration is available to determine the domain suffix that this controller use to create
+// a route for oauth. The controller then update the oauth metadata config map and update the cluster authentication config.
+// The controller use degraded condition if any part of the process fail and use the "AuthMetadataProgressing=false" condition when the controller job is done
+// and all resources exists.
+func NewMetadataController(kubeInformersForTargetNamespace informers.SharedInformerFactory, configInformer configinformers.SharedInformerFactory, routeInformer routeinformer.SharedInformerFactory,
+	secrets corev1client.SecretsGetter, configMaps corev1client.ConfigMapsGetter, route routeclient.RouteInterface, authentication configv1client.AuthenticationInterface, operatorClient v1helpers.OperatorClient,
+	recorder events.Recorder) factory.Controller {
+	c := &metadataController{
+		ingress:        configInformer.Config().V1().Ingresses().Lister(),
```
stlaz, line 61: `ingressLister`

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -0,0 +1,397 @@
...
+		authConfigNoDefaults, err = c.authentication.Create(ctx, &configv1.Authentication{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   "cluster",
+				Labels: map[string]string{},
+				Annotations: map[string]string{
+					"release.openshift.io/create-only": "true",
+				},
+			},
+		}, metav1.CreateOptions{})
+	}
```
stlaz, line 269: drop this part, go degraded

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -0,0 +1,397 @@
...
+// NewMetadataController assure that ingress configuration is available to determine the domain suffix that this controller use to create
+// a route for oauth. The controller then update the oauth metadata config map and update the cluster authentication config.
+// The controller use degraded condition if any part of the process fail and use the "AuthMetadataProgressing=false" condition when the controller job is done
+// and all resources exists.
+func NewMetadataController(kubeInformersForTargetNamespace informers.SharedInformerFactory, configInformer configinformers.SharedInformerFactory, routeInformer routeinformer.SharedInformerFactory,
+	secrets corev1client.SecretsGetter, configMaps corev1client.ConfigMapsGetter, route routeclient.RouteInterface, authentication configv1client.AuthenticationInterface, operatorClient v1helpers.OperatorClient,
+	recorder events.Recorder) factory.Controller {
+	c := &metadataController{
+		ingress:        configInformer.Config().V1().Ingresses().Lister(),
+		secrets:        secrets,
```
stlaz, line 62: you can use just a lister for secrets

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -0,0 +1,397 @@
...
+
+func routeHasCanonicalHost(route *routev1.Route, canonicalHost string) bool {
+	for _, ingress := range route.Status.Ingress {
+		if ingress.Host != canonicalHost {
+			continue
+		}
+		admitted := false
+		for _, condition := range ingress.Conditions {
+			if condition.Type == routev1.RouteAdmitted && condition.Status == corev1.ConditionTrue {
+				admitted = true
```
stlaz, line 307: `return true`

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	return nil
+}
+
+func routeHasCanonicalHost(route *routev1.Route, canonicalHost string) bool {
+	for _, ingress := range route.Status.Ingress {
+		if ingress.Host != canonicalHost {
+			continue
+		}
+		admitted := false
```
stlaz, line 304: remove this line

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -0,0 +1,397 @@
...
+		for _, condition := range ingress.Conditions {
+			if condition.Type == routev1.RouteAdmitted && condition.Status == corev1.ConditionTrue {
+				admitted = true
+			}
+		}
+		if !admitted {
+			continue
+		}
+		return true
```
stlaz, line 313: remove

pkg/operator2/operator.go
```diff
@@ -361,6 +358,19 @@ func (c *authOperator) handleSync(ctx context.Context, operatorConfig *operatorv
...
+		return fmt.Errorf("unable to get router secret: %v", err)
+	}
+	ingress, err := c.ingress.Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to get ingress: %v", err)
+	}
+
 	if err := c.handleVersion(ctx, operatorConfig, authConfig, route, routerSecret, deployment, ingress, conditions); err != nil {
 		return fmt.Errorf("error checking current version: %v", err)
 	}
```
stlaz, line 376: for the below code (`handleVersion()`) - you moved metadata and route handling outside, you should include checking them (route, well-known) in the new metadata controller.

mfojtik:
/retest

infra registry flake

pkg/operator2/operator.go
mfojtik: not sure what you mean... block 2,3,4 controllers will have to do a pre-checks if the route/config/etc. exists...

i suggest they go degraded or progressing until their pre-requirements are satisfied.

stlaz: What I meant is that you moved the oauth-metadata sync code outside. This code may affect availability and correctness of the KAS /.well-known/oauth-authorization-server endpoint, therefore it would make sense to move at least
```golang
	wellknownReady, wellknownMsg, err := c.checkWellknownEndpointsReady(ctx, authConfig, route)
	conditions.handleDegraded("WellKnownEndpoint", err)
	if err != nil {
		return fmt.Errorf("unable to check the .well-known endpoint: %v", err)
	}
	if !wellknownReady {
		conditions.setProgressingTrueAndAvailableFalse("WellKnownNotReady", wellknownMsg)
		return nil
	}
```
to the new controller.

The route health-check should be moved to the same controller that handles the route as well.

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -0,0 +1,384 @@
...
+	for _, ingress := range route.Status.Ingress {
+		if ingress.Host != canonicalHost {
+			continue
+		}
+		for _, condition := range ingress.Conditions {
+			if condition.Type == routev1.RouteAdmitted && condition.Status == corev1.ConditionTrue {
+				return true
+			}
+		}
+		return true
```
stlaz, line 300: you should remove this one line, too

mfojtik: fixed

pkg/operator2/operator.go
mfojtik: added

mfojtik:
/retest

aws flake

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+func (c *wellKnownReadyController) getAuthConfig() (*configv1.Authentication, []operatorv1.OperatorCondition) {
+	operatorConfig, err := c.authLister.Get("cluster")
+	if err != nil && os.IsNotExist(err) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, []operatorv1.OperatorCondition{
+			{
+				Type:    "ReadinessControllerDegraded",
```
stlaz, line 179: wrong name (wouldn't be unset by your update all conditions logic)

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	route, routeConditions := c.getRoute()
+	foundConditions = append(foundConditions, routeConditions...)
+
+	if authConfig != nil && route != nil {
+		// TODO: refactor this to return conditions
+		ready, endpointsMessage, err := c.isWellknownEndpointsReady(authConfig, route)
+		if !ready {
+			if err != nil {
+				foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+					Type:    "WellKnownNotReadyProgressing",
```
stlaz, line 101: `WellKnownProgressing` and `WellKnownAvailable` please, otherwise this is just confusing

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+	metadataJSON := getOAuthMetadata(route.Spec.Host)
+	err := json.Unmarshal([]byte(metadataJSON), &ret)
+	if err != nil {
+		// should never happen unless the static metadata is broken
+		panic(err)
+	}
+
+	return ret
+}
```
stlaz, line 298: you could get rid of this duplication by listing the CM and comparing its content to what KAS returns. I can do that in a follow-up

pkg/controllers/readiness/wellknown_ready_controller.go
mfojtik: @stlaz this matches the existing condition names... unfortunately we can't remove them otherwise during upgrade if they are set they will never change... we might think of a stale condition controller that will allow us to change these.

pkg/controllers/readiness/wellknown_ready_controller.go
mfojtik: hups, good catch!

pkg/controllers/readiness/wellknown_ready_controller.go
mfojtik: +1 for follow up

pkg/operator2/starter.go
```diff
@@ -206,6 +206,10 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 
 			// As of 4.6, this will appear as a configObserver error
 			"IdentityProviderConfigDegraded",
+
+			// This was changed to WellKnownProgressing and WellKnownAvailable in 4.6.
+			"WellKnownNotReadyProgressing",
+			"WellKnownNotReadyAvailable",
```
stlaz, line 212: You got me confused - these two condition types never really existed, no need to add them here.

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	if authConfig != nil && route != nil {
+		// TODO: refactor this to return conditions
+		ready, endpointsMessage, err := c.isWellknownEndpointsReady(authConfig, route)
+		if !ready {
+			if err != nil {
+				foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+					Type:    "WellKnownNotReadyProgressing",
+					Status:  operatorv1.ConditionTrue,
+					Reason:  "RouteCheckFailed",
+					Message: fmt.Sprintf("Route %q is failing health check: %v", route.Name, err),
```
stlaz, line 105: you're not health-checking routes but KAS instances

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	operatorClient v1helpers.OperatorClient
+	authLister     configv1lister.AuthenticationLister
+	routeLister    routev1lister.RouteLister
+}
+
+// knownConditionNames lists all condition types used by this controller.
+// These conditions are operated and defaulted by this controller.
+// Any new condition used by this controller sync() loop should be listed here.
+var knownConditionNames = sets.NewString(
+	"WellKnownEndpointDegraded",
```
stlaz, line 62: You're not actually setting this value anywhere

mfojtik:
@stlaz updated

pkg/controllers/readiness/wellknown_ready_controller.go
stlaz: Since you don't set it to true, move it to Stale conditions

mfojtik: i just forgot to remove it, thanks

stlaz: You still need to move `WellKnownEndpointDegraded` among the other stale conditions in the stale condition controller, this one is the only pre-existing type

PR #294, opened 2020-06-29
Use config observer from library-go

vareti:
Bumps openshift/api and openshift/library-go and reuses config observer from https://github.com/openshift/library-go/pull/816

### DIFF
--- a/pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
+++ b/pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
@@ -7,6 +7,7 @@ import (
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/configobserver/apiserver"
+	configobserveroauth "github.com/openshift/library-go/pkg/operator/configobserver/oauth"
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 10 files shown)


pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -67,6 +68,7 @@ func NewConfigObserver(
 		oauth.ObserveIdentityProviders,
 		oauth.ObserveTemplates,
 		oauth.ObserveTokenConfig,
+		configobserveroauth.ObserveAccessTokenInactivityTimeout,
```
stlaz, line 71: you need to remove the field observation from the other observer, otherwise you've got two entities setting the same field

vareti: Sure, will do. I am in the process of doing that.

pkg/controllers/configobservation/interfaces.go
```diff
@@ -22,7 +22,7 @@ type Listers struct {
 	APIServerLister_     configlistersv1.APIServerLister
 	ConsoleLister        configlistersv1.ConsoleLister
 	InfrastructureLister configlistersv1.InfrastructureLister
-	OAuthLister          configlistersv1.OAuthLister
+	OAuthConfigLister    configlistersv1.OAuthLister
```
stlaz, line 25: `OAuthLister_` to match the other fields, please

PR #296, opened 2020-06-30
Bug 1852312: fix template mounting to the oauth-server pods

stlaz:
fallout of a recent refactoring

/cc @vareti @marun

### DIFF
--- a/test/e2e/templates_test.go
+++ b/test/e2e/templates_test.go
@@ -1,0 +1,138 @@
+package e2e
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
+
+	e2e "github.com/openshift/cluster-authentication-operator/test/library"
+)
+
+func TestTemplatesConfig(t *testing.T) {
+	kubeConfig := e2e.NewClientConfigForTest(t)
+
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+	configClient, err := configclient.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+	routeClient, err := routeclient.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+
+	configSecretsClient := kubeClient.CoreV1().Secrets("openshift-config")
+
+	cleanup := createSecret(t, configSecretsClient, "login", "login.html", "login test")
+	defer cleanup()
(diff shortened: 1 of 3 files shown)


test/e2e/templates_test.go
```diff
@@ -0,0 +1,138 @@
...
+	require.NoError(t, err, "failed to wait for the authentication operator to become available")
+
+	route, err := routeClient.RouteV1().Routes("openshift-authentication").Get(context.TODO(), "oauth-openshift", metav1.GetOptions{})
+	require.NoError(t, err)
+	oauthURL, err := url.Parse("https://" + route.Spec.Host)
+	require.NoError(t, err)
+
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
```
vareti, line 97: (no action required) Do we have any test environment in our e2e that needs this flag to be set? Does this need to be set for all test involving oauth-server?

test/e2e/templates_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// modify URL to go to the IdP login
+	oauthURL.Path = "/oauth/token/display"
+	oauthURL.RawQuery = fmt.Sprintf("client_id=openshift-browser-client&idp=super+duper+production-ready+htpasswd+idp&redirect_uri=%s&response_type=code", url.QueryEscape(oauthURL.String()))
+	oauthURL.Path = "/oauth/authorize"
+	resp, err = httpClient.Get(oauthURL.String())
+	require.NoError(t, err)
+	body, err = ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Contains(t, string(body), "login test")
```
vareti, line 118: Just wondering why the case to test `error.html` was not added?

test/e2e/templates_test.go
```diff
@@ -0,0 +1,138 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
+
+	e2e "github.com/openshift/cluster-authentication-operator/test/library"
```
vareti, line 23: sort imports?

test/e2e/templates_test.go
stlaz: I'd rather have it set in case someone wants to try running the tests with a proxy. It definitely needs to appear in origin tests that involve http calls outside the cluster/to routes.

test/e2e/templates_test.go
stlaz: I was not sure how to force an error on the server, it did not feel so important with the other tests and the payload not erroring.

test/e2e/templates_test.go
stlaz: I think they are sorted, aren't they?
```
sys deps

outside deps

kube deps

openshift deps

local deps
```

stlaz:
/retest
infra

test/e2e/templates_test.go
vareti: Interesting, Good to know. Thanks!

test/e2e/templates_test.go
vareti: I was under the impression that imports must be sorted in the order

```
sys deps

github deps

kube deps
```

Didn't know that they could sorted this way too.

PR #299, opened 2020-07-09
Finalize operator refactoring

mfojtik:
This is the final refactor PR:

* The `operator2` package is gone
* Each handler is now separate, independent controller
* Common code (getters) moved to common package


pkg/controllers/targetversion/target_version_controller.go
```diff
@@ -0,0 +1,436 @@
...
+
+	routeSecret, routeSecretConditions := c.getRouteSecret()
+	foundConditions = append(foundConditions, routeSecretConditions...)
+
+	deployment, deploymentConditions := c.getDeployment()
+	foundConditions = append(foundConditions, deploymentConditions...)
+
+	foundConditions = append(foundConditions, c.checkRouteHealthy(route, routeSecret, ingressConfig)...)
+	foundConditions = append(foundConditions, c.checkDeploymentReady(deployment)...)
+	foundConditions = append(foundConditions, c.oauthClientsReady(ctx)...)
```
stlaz: None of this belongs here, all of these checks should be spread to the controllers that have these resources as an output. 

Eventually, if you want to handle version like this, by a controller, you'll get the conditions from the state of the operator and set/not set the version based on that information.

pkg/controllers/common/ingress_config.go
```diff
@@ -0,0 +1,30 @@
+package common
+
+import (
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	v1 "github.com/openshift/client-go/config/listers/config/v1"
```
stlaz, line 8: nit: configv1lister

pkg/controllers/common/route.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"github.com/davecgh/go-spew/spew"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
+
+	v1 "github.com/openshift/client-go/route/listers/route/v1"
```
stlaz, line 19: nit: routev1lister

stlaz: nit: belongs to previous block

pkg/controllers/deployment/deployment_controller.go
```diff
@@ -310,64 +230,58 @@ func (c *deploymentController) sync(ctx context.Context, syncContext factory.Syn
...
-				Reason:  "ReplicasNotReady",
-				Message: fmt.Sprintf("Waiting for all OAuth server replicas to be ready (%d not ready)", deployment.Status.Replicas-deployment.Status.UpdatedReplicas),
-			},
-			{
-				Type:    "OAuthServerAvailable",
-				Status:  operatorv1.ConditionTrue,
-				Reason:  "AsExpected",
-				Message: fmt.Sprintf("%d available replicas found for OAuth Server", deployment.Status.AvailableReplicas),
-			},
+func routeHasCanonicalHost(route *routev1.Route, canonicalHost string) bool {
```
stlaz, line 237: you've got common for this now

pkg/controllers/deployment/deployment_controller.go
```diff
@@ -227,6 +211,11 @@ func (c *deploymentController) sync(ctx context.Context, syncContext factory.Syn
 		return err
 	}
 
+	if len(foundConditions) > 0 {
+		klog.V(4).Infof("Retrying because conditions: %s", spew.Sdump(foundConditions))
+		return factory.SyntheticRequeueError
+	}
```
stlaz, line 217: I missed this bit with the other controllers also. Shouldn't this be assumed by the events that you connect the controller to, instead of looping on a errors?

pkg/controllers/targetversion/target_version_controller.go
stlaz: nvm, discard, I guess if we want to be checking these in parallel and not be stopped by other effects, this is fine

pkg/controllers/deployment/deployment_controller.go
mfojtik: @stlaz i think retrying faster when there are degraded conditions makes sense, but now that most of the controllers are driven by listers I think it is fine if we remove this?

pkg/controllers/targetversion/target_version_controller.go
mfojtik: i thought the purpose of this controller is to set version **only** is all these pre-checks are satisfied (iow. the operand is fully deployed and available). is that wrong assumption?

pkg/controllers/deployment/deployment_controller.go
stlaz: the events you attach to have always been driven by informers, correct? I would assume that in most cases you would just see that an object your controller watches changed and a new loop is started to handle the event.

I can think of just one exception which is the KAS grabbing the metadata configmap and performing rollouts that are slow and we're not watching the static pods, but in that case, time resync should be used (and we used it everywhere now).

mfojtik: right, initially, some of these were driven by clients which means you wanted to retry when something bad happened... as most of these are now listers driven, I think the risk is low and the periodic resync should handle wedged controller case (if there are any).

mfojtik: (removed)

pkg/controllers/targetversion/target_version_controller.go
stlaz: it's a correct assumption, the source of those prereqs could be different (the operator's status itself?), but if a controller stops on not being able to retrieve route, then I guess it won't care about setting/unsetting the RouteHealthDegraded condition.

mfojtik: status can be changed/cleared by external actors, so it is not the best thing to rely on (I did that in refactoring steps temporarily ;-)

pkg/controllers/targetversion/target_version_controller.go
```diff
@@ -27,6 +17,13 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	appsv1 "k8s.io/api/apps/v1"
```
stlaz, line 20: order!

mfojtik: now I know why @deads2k hate this comments :-)

stlaz:
/approve
/lgtm
/hold
for the part 3 merge

PR #301, opened 2020-07-13
controllers: abstract status setting logic

stlaz:
Further simplifies the syncs among controllers

/assign @mfojtik


mfojtik:
/lgtm

At some point we can move this into factory.

PR #311, opened 2020-07-30
[wip] add token timeout syncing controller

stlaz:
/assign @sttts


pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		c.clientTimeouts[client.Name] = timeout
+	}
+
+	return nil
+}
+
+func (c *tokenTimeoutSyncController) updateToken(ctx context.Context, token *oauthv1.OAuthAccessToken, globalChanged bool) error {
+	tokenClientName := token.ClientName
+	if len(tokenClientName) == 0 {
```
vareti, line 114: any specific reason why the token is not updated with `globalTimeout`?

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil
+	}
+
+	tokenClient, err := c.oauthClientLister.Get(tokenClientName)
+	if err != nil {
+		return err
```
vareti, line 120: same as above. any specific not to update it?

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
stlaz: client name is not optional so this case should never happen, I added a log here

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
stlaz: you don't know what to update it to in this case, do you?

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+	configinformer "github.com/openshift/client-go/config/informers/externalversions"
+	configv1lister "github.com/openshift/client-go/config/listers/config/v1"
+	oauthv1client "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	oauthinformer "github.com/openshift/client-go/oauth/informers/externalversions"
+	oauthv1lister "github.com/openshift/client-go/oauth/listers/oauth/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+type tokenTimeoutSyncController struct {
```
sttts, line 23: document what this is doing

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+	oauthConfigLister configv1lister.OAuthLister
+	oauthClientLister oauthv1lister.OAuthClientLister
+	accessTokens      oauthv1client.OAuthAccessTokenInterface
+
+	globalTimeout  time.Duration
```
sttts, line 28: document what -1 means.

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+	c := &tokenTimeoutSyncController{
+		accessTokens:      accessTokenClient,
+		oauthConfigLister: configInformers.Config().V1().OAuths().Lister(),
+		oauthClientLister: oauthInformers.Oauth().V1().OAuthClients().Lister(),
+
+		globalTimeout:  -1,
+		clientTimeouts: map[string]*time.Duration{},
+	}
+
+	return factory.New().ResyncEvery(30*time.Second).WithInformers(
```
sttts, line 56: is 30s resync interval really common? pretty often for this kind of controller.

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
sttts, line 28: s/globalTimeout/lastSeenConfigTimeout/

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
stlaz: @mfojtik told me we want it in case informers get stuck, I can live w/o it

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+func (c *tokenTimeoutSyncController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	var globalChanged bool
+
+	oauthConfig, err := c.oauthConfigLister.Get("cluster")
+	if err != nil {
+		return err
+	}
+
+	if oauthConfig.Spec.TokenConfig.AccessTokenInactivityTimeout != nil {
+		if newTimeout := oauthConfig.Spec.TokenConfig.AccessTokenInactivityTimeout.Duration; c.globalTimeout != newTimeout { // don't do anything on first run
```
sttts, line 71: remove the trailing comment. It's not adding anything

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+			// only update if this is not the first run and the new new timeout is lower than the previous
+			globalChanged = c.globalTimeout != -1 && ((c.globalTimeout > 0 && newTimeout < c.globalTimeout) || (c.globalTimeout == 0 && newTimeout > 0))
+			c.globalTimeout = newTimeout
+		}
+	} else {
+		globalChanged = c.globalTimeout != 0
```
sttts, line 77: this fires on first run. Should it?

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+		c.globalTimeout = 0
+	}
+
+	tokenList, err := c.accessTokens.List(ctx, metav1.ListOptions{LabelSelector: labels.Everything().String()})
```
sttts, line 81: can't one leave the list options empty?

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, t := range tokenList.Items {
+		if err := c.updateToken(ctx, &t, globalChanged); err != nil {
+			// can't log token until we are sure only sha256: tokens exist
+			klog.Errorf("failed to upate token")
```
sttts, line 89: one failing out of a thousand is not much. Do we retry?

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+		globalChanged = c.globalTimeout != 0
+		c.globalTimeout = 0
+	}
+
+	tokenList, err := c.accessTokens.List(ctx, metav1.ListOptions{LabelSelector: labels.Everything().String()})
+	if err != nil {
+		return err
+	}
+
+	for _, t := range tokenList.Items {
```
sttts, line 86: why not skip this if `!globalChanged`?

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
stlaz: I removed the "don't fire on 1st round" for clientChanges but forgot about global change. I think we can actually attempt the updates even on the first round as no token should have timeout greater than `now + whicheverTimeoutApplies` so no actual updates would be generated

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
stlaz: clients could have changed their timeouts

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+
+	for _, client := range oauthClients {
+		var timeout *time.Duration
+		if client.AccessTokenInactivityTimeoutSeconds == nil {
+			timeout = nil
+		} else {
+			t := time.Duration(*client.AccessTokenInactivityTimeoutSeconds) * time.Second
+			timeout = &t
+		}
+		c.clientTimeouts[client.Name] = timeout
```
sttts, line 106: why don't you update first? And even compare whether anything changed? Would prefer to avoid listing tokens all the time.

pkg/controllers/tokentimeoutsync/tokentimeoutsync.go
```diff
@@ -0,0 +1,179 @@
...
+	if timeout == 0 {
+		changed = tokenCopy.InactivityTimeoutSeconds != 0
+		tokenCopy.InactivityTimeoutSeconds = 0
+		return changed
+	}
+
+	// count a new timeout based on the current time
+	newTimeout := int32((time.Now().Sub(tokenCopy.CreationTimestamp.Time) + timeout) / time.Second)
+	if oldTimeout := tokenCopy.InactivityTimeoutSeconds; oldTimeout == 0 || newTimeout < tokenCopy.InactivityTimeoutSeconds {
+		tokenCopy.InactivityTimeoutSeconds = newTimeout
```
sttts, line 159: this is wrong. We cannot reset all of them. We have to add the difference between new and old (global/client) value.

PR #314, opened 2020-08-04
Bug 1859591: version controller: honor system CA bundle when checking route health

stlaz:
/assign @marun

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"context"
+	"io/ioutil"
 	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/operator/starter.go
```diff
@@ -336,13 +338,19 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 		controllerContext.EventRecorder,
 	)
 
+	systemCABundle, err := ioutil.ReadFile("/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem")
+	if err != nil {
+		// this may fail route-health checks in proxy environments
+		klog.Warningf("Unable to read system CA from /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem: %v", err)
```
mfojtik, line 344: why not make this check during the route healthcheck and report it in message to user?

stlaz: it's not a check, you're loading the full file which might be quite long, you don't want this io operation appearing all the time

mfojtik: i think a simple "file exists" check (os.Stat) would be cheap?

stlaz: heh, I see why this is confusing, I added a new argument but did not wire it :facepalm: 

The contents of the file are being used. I do not care whether it exists or not.

PR #317, opened 2020-08-06
Bug 1860922: adds OAuthAPIServer pruner controller

p0lyn0mial:
This PR adds a new controller that will remove `openshift-oauth-apiserver` namespace. The namespace holds the `OAuth API Server` and all necessary resources that are valid only for` 4.6+` clusters. This controller is intended to clean up in case of a downgrade from 4.6 to 4.5.

This PR also makes sure that the new conditions that are valid only for 4.6 will be pruned.

This PR builds on https://github.com/openshift/cluster-authentication-operator/pull/307


p0lyn0mial:
please review only the last 3 commits.

pkg/controller/oauthapiserverpruner/oauth_apiserver_pruner.go
```diff
@@ -0,0 +1,137 @@
...
+		apiregistrationv1Lister: apiregistrationv1Informer.Lister(),
+		authOperatorLister:      authInformers.Lister(),
+		namespaceLister:         kubeInformersForNamespaces.InformersFor("").Core().V1().Namespaces().Lister(),
+	}
+
+	controllerFactory := factory.New()
+	controllerFactory.WithInformers(authInformers.Informer(), kubeInformersForNamespaces.InformersFor("").Core().V1().Namespaces().Informer(), apiregistrationv1Informer.Informer())
+	controllerFactory.WithSync(c.sync)
+
+	return controllerFactory.ToController(name, eventRecorder.WithComponentSuffix("oauth-apiserver-cleaner-controller"))
```
stlaz, line 67: no action needed:
this equals:
```golang
return factory.New().
    WithInformers(
        authInformers.Informer(),
        kubeInformersForNamespaces.InformersFor("").Core().V1().Namespaces().Informer(),
        apiregistrationv1Informer.Informer()).
    WithSync(c.sync).
    ToController(name, eventRecorder.WithComponentSuffix("oauth-apiserver-cleaner-controller"))
```

sdodson:
@p0lyn0mial what's the difference between this and #307 ?

sdodson:
By that I mean, both are in the queue, which should be accepted?

stlaz:
This one adds 3 more commits on top of #307 so that @p0lyn0mial does not have to rebase it once the other PR merges because he might be on PTO. These are two PRs fixing two different problems. Both should be accepted, #307 should be merged before this one.

PR #319, opened 2020-08-07
oauthserver: grab unsupported overrides from a prefixed tree

stlaz:
Also removes duplicate code

/assign @vareti


stlaz:
/remove-lifecycle stale
this is still relevant

PR #321, opened 2020-08-07
logs when the encryption deployer hasn't yet converged

p0lyn0mial:
it turns out that `!converged` will also be reported during the oauth apiserver installation

### DIFF
--- a/pkg/operator/apiservices/sync_take_over_apiservices_oauth_apiserver.go
+++ b/pkg/operator/apiservices/sync_take_over_apiservices_oauth_apiserver.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/apiservices/sync_take_over_apiservices_oauth_apiserver.go
```diff
@@ -63,7 +64,8 @@ func syncManageAPIServicesController(deployer statemachine.Deployer, authOperato
 			return err
 		}
 		if !converged && !operator.Status.ManagingOAuthAPIServer {
-			syncContext.Recorder().Warning("PrereqNotReady", "the encryption deployer hasn't yet converged, retrying in 5 minutes")
+			// !converged will also be reported during the oauth apiserver installation
```
p0lyn0mial, line 67: here https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/deployer/revisionedpod.go#L168

p0lyn0mial: alternatively, we could export `progressing` and simply not fire an event in that case.

PR #322, opened 2020-08-11
Bug 1867908: add connection timeouts and proxy from env to http client

vareti:
Add client side timeout to not wait ~2mins to report a failure. Request is expected to be served within 5 seconds now. Also add the flag to get proxy configuration from the environment. This flag should resolve the issues we are seeing in proxy environments.

### DIFF
--- a/pkg/controllers/endpointaccessible/endpoint_accessible_controller.go
+++ b/pkg/controllers/endpointaccessible/endpoint_accessible_controller.go
@@ -183,10 +183,15 @@ func (c *endpointAccessibleController) sync(ctx context.Context, syncCtx factory
 			req.WithContext(reqCtx)
 
 			// we don't really care  if anyone lies to us. We aren't sending important data.
-			insecureTransport := &http.Transport{
-				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			client := &http.Client{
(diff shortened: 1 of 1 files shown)


pkg/controllers/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -183,10 +183,17 @@ func (c *endpointAccessibleController) sync(ctx context.Context, syncCtx factory
...
 
 			// we don't really care  if anyone lies to us. We aren't sending important data.
-			insecureTransport := &http.Transport{
-				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			client := &http.Client{
+				Transport: &http.Transport{
+					Proxy: http.ProxyFromEnvironment,
+					DialContext: (&net.Dialer{
+						Timeout: 5 * time.Second,
+					}).DialContext,
```
stlaz, line 191: It feels like it would be simpler to just set the `Timeout` attribute of the `http.Client` struct, is there any reason not to?

vareti: @stlaz I had only tcp timeout in my mind when I was writing this. But, I guess client `Timeout` also works in this case and would be simpler too. Update the code. Please take a look.

PR #323, opened 2020-08-11
authentication operator is marked available because the deployment is missing

deads2k:
Since the CVO treats availability as "past the post", this cause a blip which disturbs the installation flow.

### DIFF
--- a/pkg/controllers/endpointaccessible/endpoint_accessible_controller.go
+++ b/pkg/controllers/endpointaccessible/endpoint_accessible_controller.go
@@ -4,134 +4,23 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"net"
 	"net/http"
-	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


pkg/controllers/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -4,134 +4,23 @@ import (
...
-	"net"
 	"net/http"
-	"strconv"
 	"sync"
 	"time"
 
-	routev1informers "github.com/openshift/client-go/route/informers/externalversions/route/v1"
-	routev1listers "github.com/openshift/client-go/route/listers/route/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+
```
stlaz, line 12: nit: no need for this empty line

stlaz:
/hold
should have looked at the tests sooner, they all appear to fail with
```
Cluster operator authentication Available is False with WellKnown_NotReady: WellKnownAvailable: The well-known endpoint is not yet available: failed to get API server IPs: failed to get kube api server service: service \"kubernetes\" not found
```

PR #327, opened 2020-08-26
Don't bump operator version when the oauth-apiserver deployment is not yet ready

stlaz:
Updates the target version controller to take the oauth API server deployment into account.

In the ideal world, we should only check the clusteroperator's "Available" condition and bump the version when that is true (==> the version controller would depend on all the other controllers properly reporting status)


stlaz:
/close
the target version controller is getting removed

PR #328, opened 2020-08-28
Bug 1870565: deployment: avoid degraded due to conflict error

sttts:
Deployment controller was possibly constantly updating the observed generation and replica count. Moreover, it was not protected to conflict errors from the peer controllers.


sttts:
Ingress is not healthy.

/retest

PR #330, opened 2020-08-31
Bug 1866818: routercerts: always trust the default ingress ca

stlaz:
Sometimes, the CA cert does not seem to appear in the routercerts secret. We should have always trusted the default ingress cert for router certs, adding that here.

/assign @deads2k 
/cc @Miciah

### DIFF
--- a/pkg/controllers/routercerts/controller.go
+++ b/pkg/controllers/routercerts/controller.go
@@ -32,6 +32,7 @@ type routerCertsDomainValidationController struct {
 	operatorClient  v1helpers.OperatorClient
 	ingressLister   configv1listers.IngressLister
 	secretLister    corev1listers.SecretLister
+	configMapLister corev1listers.ConfigMapLister
 	targetNamespace string
 	secretName      string
 	routeName       string
@@ -44,6 +45,7 @@ func NewRouterCertsDomainValidationController(
 	eventRecorder events.Recorder,
 	ingressInformer configv1informers.IngressInformer,
 	secretInformer corev1informers.SecretInformer,
+	configMapInformer corev1informers.ConfigMapInformer,
 	targetNamespace string,
 	secretName string,
 	routeName string,
@@ -52,6 +54,7 @@ func NewRouterCertsDomainValidationController(
 		operatorClient:  operatorClient,
 		ingressLister:   ingressInformer.Lister(),
 		secretLister:    secretInformer.Lister(),
+		configMapLister: configMapInformer.Lister(),
 		targetNamespace: targetNamespace,
 		secretName:      secretName,
 		routeName:       routeName,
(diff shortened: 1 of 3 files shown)


Miciah:
From [the ingress operator logs](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/330/pull-ci-openshift-cluster-authentication-operator-master-e2e-upgrade/1300382027652009984/artifacts/e2e-upgrade/pods/openshift-ingress-operator_ingress-operator-785cdf498f-ctkh2_ingress-operator.log), it looks like e2e-upgrade failed because of [BZ#1873728](https://bugzilla.redhat.com/show_bug.cgi?id=1873728).

Miciah:
https://github.com/openshift/cluster-ingress-operator/pull/446, which should fix [BZ#1873728](https://bugzilla.redhat.com/show_bug.cgi?id=1873728), has merged.

/test e2e-upgrade

pkg/controllers/routercerts/controller.go
```diff
@@ -123,23 +145,13 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
-			verifyOptions.Intermediates.AddCert(certificate)
-		default:
-			serverCerts = append(serverCerts, certificate)
-		}
-	}
-
-	if len(verifyOptions.Roots.Subjects()) == 0 {
-		return newRouterCertsDegradedf("NoRootCARouterCerts", "secret/%v.spec.data[%v] -n %v: no root CA certificates found in secret or system", c.secretName, ingressDomain, c.targetNamespace)
+	// ignore the server cert from the default cert bundle
+	populateVerifyOptionsFromCertSlice(&verifyOptions, ingressCACerts)
```
sanchezl, line 149: Even though the comment above states so, it might be easier to grep at first glance for future reviewers if you explicitly ignore the return value.
```suggestion
	_ = populateVerifyOptionsFromCertSlice(&verifyOptions, ingressCACerts)
```
Also (nit), consider renaming the method to better describe the full functionality:
```suggestion
	_ = populateVerifyOptionsAndGetServerCertsFromCertSlice(&verifyOptions, ingressCACerts)
```

sanchezl (approved):

pkg/controllers/routercerts/controller.go
sttts: +1 for `_ = ...`

Can be a follow-up.

Miciah:
Can we backport this PR?  It would be helpful to have the change in earlier releases in order to obviate the need to use the proxy trusted CA as a side channel for injecting the CA for a custom PKI.  As noted in [the `default-ingress-cert-configmap` enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/ingress/default-ingress-cert-configmap.md), this workaround was supposed to be unnecessary as of OCP 4.4.0.  We were planning to remove the documentation for this workaround when the issue that this PR fixes was discovered: https://github.com/openshift/openshift-docs/pull/24979.

PR #332, opened 2020-09-02
Bug 1874713: deployment: don't panic when applying deployment fails

stlaz:
In case there was an error during `ApplyDeployment`, the deployment would be `nil`, the controller would panic never to recover again.

/assign @mfojtik

### DIFF
--- a/pkg/controllers/deployment/deployment_controller.go
+++ b/pkg/controllers/deployment/deployment_controller.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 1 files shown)


pkg/controllers/deployment/deployment_controller.go
```diff
@@ -362,3 +352,33 @@ func (c *deploymentController) ensureBootstrappedOAuthClients(ctx context.Contex
...
+
+// updateOperatorDeploymentInfo updates the operator's Status .ReadyReplicas, .Generation and the
+// .Generetions field with data about the oauth-server deployment
+func (c *deploymentController) updateOperatorDeploymentInfo(
+	ctx context.Context,
+	syncContext factory.SyncContext,
+	operatorConfig *operatorv1.Authentication,
+	deployment *appsv1.Deployment,
+) error {
+	operatorStatusOutdated := operatorConfig.Status.ObservedGeneration != operatorConfig.Generation || operatorConfig.Status.ReadyReplicas != deployment.Status.UpdatedReplicas
```
p0lyn0mial, line 364: why we are not using [the generic workload controller](https://github.com/openshift/library-go/blob/master/pkg/operator/apiserver/controller/workload/workload.go) for this deployment? 

The generic controller is used to deploy the openshift-apiserver and the new oauth-apiserver

sttts:
/lgtm

@p0lyn0mial your question is a good one. Would like to hear the answer. Let's fix the panic first though.

stlaz:
We don't use it because it only came after we started using the current logic. I bet there are side effects that would need to be solved (setting the operand version, for example, there might be more).

It's (probably) not impossible to use the generic workload controller, I just wanted to have this fixed fast.

mfojtik:
/lgtm

@stlaz sounds like a tech debt Jira or bugzilla (your choice) :-) not urgent.

PR #334, opened 2020-09-03
bug 1875038: endpointaccessible: do not panic when no endpoints are found

mfojtik:


pkg/controllers/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -55,6 +55,12 @@ func (c *endpointAccessibleController) sync(ctx context.Context, syncCtx factory
 		return err
 	}
 
+	// the endpoint list function can return empty results without error when the endpoints does not exists
+	// requeue in that case and wait until we get valid endpoints to check.
+	if len(endpoints) == 0 {
+		return factory.SyntheticRequeueError
```
deads2k, line 61: if this is the case, we should be degraded.  This shouldn't be a synthentic error.

deads2k: > if this is the case, we should be degraded. This shouldn't be a synthentic error.

actuall, set the AVailable to false.

mfojtik: @deads2k updated, we go available=false in case no endpoints are returned.

PR #336, opened 2020-09-07
Bug 1876471: Add ingress-operator as provider for router-certs

mfojtik:


pkg/operator/starter.go
```diff
@@ -200,7 +200,7 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 	// add syncing for router certs for all cluster ingresses
 	if err := operatorCtx.resourceSyncController.SyncSecret(
 		resourcesynccontroller.ResourceLocation{Namespace: "openshift-authentication", Name: "v4-0-config-system-router-certs"},
-		resourcesynccontroller.ResourceLocation{Namespace: "openshift-config-managed", Name: "router-certs"},
+		resourcesynccontroller.ResourceLocation{Namespace: "openshift-config-managed", Name: "router-certs", Provider: "ingress-operator"},
```
sttts, line 203: @soltysh @hexfusion @adambkaplan for better error message that users can understand, we need similar PRs and add the providers everywhere

soltysh: /cc @tnozicka & @damemi

PR #340, opened 2020-09-10
readiness: suggest checking kube apiserver operator when min number of API server is not met

mfojtik:
…f API server is not met

### DIFF
--- a/pkg/controllers/readiness/wellknown_ready_controller.go
+++ b/pkg/controllers/readiness/wellknown_ready_controller.go
@@ -185,7 +185,7 @@ func (c *wellKnownReadyController) isWellknownEndpointsReady(spec *operatorv1.Op
 		return nil
 	}
 	if expectedMinNumber := getExpectedMinimumNumberOfMasters(spec); len(ips) < expectedMinNumber {
-		return fmt.Errorf("need at least %d kube-apiservers, got %d", expectedMinNumber, len(ips))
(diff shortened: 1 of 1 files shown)


pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -185,7 +185,7 @@ func (c *wellKnownReadyController) isWellknownEndpointsReady(spec *operatorv1.Op
 		return nil
 	}
 	if expectedMinNumber := getExpectedMinimumNumberOfMasters(spec); len(ips) < expectedMinNumber {
-		return fmt.Errorf("need at least %d kube-apiservers, got %d", expectedMinNumber, len(ips))
+		return fmt.Errorf("need at least %d kube-apiservers, got %d (see kube-apiserver operator for more information)", expectedMinNumber, len(ips))
```
sttts, line 188: isn't this not only about existance, but about kube-apiservers all being up-to-date? Or is that error string somewhere else?

sttts:
```
waiting for at least %d kube-apiservers, kube-apiserver operators reports only %d (see installer logs and check why not all master nodes are up)
```

wking:
/retitle readiness: suggest checking kube apiserver operator when min number of API server is not met

@mfojtik , might want to fix "mim" -> "min" in the commit message to :shrug:

PR #341, opened 2020-09-10
Bug 1877803: don't error out on non-json output of password grant attempt

stlaz:
Some OIDCs return HTML page when they don't support password grants.


stlaz:
/cc @marun @vareti 
for review

stlaz:
/hold cancel
CI infra fails should be resolved now

PR #344, opened 2020-09-11
Bug 1878776: ingressnodesavailable: add controller that checks if router can schedule pods

mfojtik:
The purpose of this controller is to check whether nodes are available to schedule ingress pods (which require workload schedulable nodes).
Not having ingress available is very common failure case for authentication but it lacks a good signal (from ingress operator) and the authentication operator is not available resulting in longer bug triage and red herring.

This controller handle also case when masters are schedulable, however the workers schedulable is best-effort as there could be taints and toleration that can take effect and cause router pods to not be able to schedule.

### DIFF
--- a/pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
+++ b/pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
@@ -1,0 +1,107 @@
+package ingressnodesavailable
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/common"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var knownConditionNames = sets.NewString(
+	"ReadyIngressNodesAvailable",
+)
+
+// ingressNodesAvailableController validates that router certs match the ingress domain
+type ingressNodesAvailableController struct {
+	operatorClient v1helpers.OperatorClient
+	nodeLister     corev1listers.NodeLister
+}
+
+func NewIngressNodesAvailableController(
+	operatorClient v1helpers.OperatorClient,
+	eventRecorder events.Recorder,
+	nodeInformer corev1informers.NodeInformer,
+) factory.Controller {
+	controller := &ingressNodesAvailableController{
+		operatorClient: operatorClient,
+		nodeLister:     nodeInformer.Lister(),
+	}
+
+	return factory.New().
+		WithInformers(
+			operatorClient.Informer(),
+			nodeInformer.Informer(),
+		).
+		WithSync(controller.sync).
+		ResyncEvery(1*time.Minute).
+		ToController("IngressNodesAvailableController", eventRecorder)
+}
+
+func countReadyWorkerNodes(nodes []*corev1.Node) int {
+	readyNodes := 0
+	for _, n := range nodes {
+		for _, c := range n.Status.Conditions {
+			if c.Type == "Ready" && c.Status == "True" {
+				readyNodes++
+			}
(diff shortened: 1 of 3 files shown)


pkg/controllers/worker/worker_available_controller.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// this is best-effort, workers can be tainted and not schedulable, which will result in auth operator to fail
+	// as the router need schedulable worker nodes.
+	readyWorkers := countReadyNodes(workers)
+
+	// we don't have any worker nodes schedulable, but we can run clusters that have master nodes schedulable, so we need to check that
+	// before going available==false
+	var masters []*corev1.Node
+	if readyWorkers == 0 {
```
sttts, line 79: can't customers have both? schedulable masters and nodes?

pkg/controllers/worker/worker_available_controller.go
```diff
@@ -0,0 +1,108 @@
...
+					break
+				}
+			}
+			if isSchedulable {
+				readyWorkers++
+			}
+		}
+	}
+
+	if len(workers) == 0 {
```
sttts, line 98: `if readyWorkers == 0`, no?

pkg/controllers/worker/worker_available_controller.go
```diff
@@ -0,0 +1,108 @@
...
+		foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+			Type:    "WorkerNodesAvailable",
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "NoIngressNodes",
+			Message: fmt.Sprint("Authentication require functional ingress which require at least one schedulable worker node, got %d workers and %d master nodes", len(workers), len(masters)),
```
sttts, line 103: say something about schedulable and/or ready

pkg/controllers/worker/worker_available_controller.go
```diff
@@ -0,0 +1,108 @@
...
+				readyWorkers++
+			}
+		}
+	}
+
+	if len(workers) == 0 {
+		foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+			Type:    "WorkerNodesAvailable",
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "NoIngressNodes",
```
sttts, line 102: NoReadyIngressNodes

pkg/controllers/worker/worker_available_controller.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+			if isSchedulable {
+				readyWorkers++
+			}
+		}
+	}
+
+	if len(workers) == 0 {
+		foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+			Type:    "WorkerNodesAvailable",
```
sttts, line 100: ReadyIngressNodesAvailable

pkg/operator/starter.go
```diff
@@ -6,6 +6,9 @@ import (
 	"os"
 	"time"
 
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/worker"
+	"k8s.io/client-go/informers"
+
```
sttts, line 11: order

marun:
Seems more than a little cray that the auth operator needs to take responsibility for this.

pkg/controllers/worker/worker_available_controller.go
mfojtik: good question, I think we can check all nodes then and find if we have something router pods can schedule on.

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
```diff
@@ -0,0 +1,107 @@
...
+			workloadReadyNodes++
+		}
+	}
+
+	if workloadReadyNodes == 0 {
+		foundConditions = append(foundConditions, operatorv1.OperatorCondition{
+			Type:   "ReadyIngressNodesAvailable",
+			Status: operatorv1.ConditionFalse,
+			Reason: "NoReadyIngressNodes",
+			Message: fmt.Sprintf("Authentication require functional ingress which require at least one schedulable and ready node. Got %d worker nodes and %d master nodes (none are schedulable or ready for ingress pods).",
```
sttts, line 101: requires

sttts:
Make edge team owner of this.

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
mfojtik: fixed

mfojtik:
adding lgtm back, fixed stucked informer

mfojtik:
OWNERS file copied from ingress-operator

PR #345, opened 2020-09-14
Give more details on oauth deployment pods status in error conditions

mfojtik:
Instead of just showing how many replicas of oauth deployment are not ready, print also details about those not ready containers and pods to provide better signal and help debugging.

Details include all information from container status, including networking conditions and terminated containers.

TODO: Move the helper code to library-go (but we only have one deployment based control plane operator so far).

### DIFF
--- a/pkg/controllers/common/deployment.go
+++ b/pkg/controllers/common/deployment.go
@@ -2,11 +2,49 @@ package common
 
 import (
 	"fmt"
+	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 )
+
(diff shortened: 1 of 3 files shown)


pkg/controllers/common/deployment.go
```diff
@@ -2,13 +2,51 @@ package common
...
+	if err != nil {
+		return nil, err
+	}
+	deploymentPodsStates := []string{}
+	for i := range deploymentPods {
+		containerStates := []string{}
+		for _, c := range append(deploymentPods[i].Status.ContainerStatuses, deploymentPods[i].Status.InitContainerStatuses...) {
+			switch {
+			case c.State.Running != nil:
+				containerStates = append(containerStates, fmt.Sprintf("container %q is running since %s", c.Name, c.State.Running.StartedAt))
```
sttts, line 27: do we need the running ones?

mfojtik: yes, if 2 and running I want to know when they started and why there is 1 pending for 20 minutes :-) more details here the better.

PR #347, opened 2020-09-15
Bug 1874056: inline apiserver-lib-go ToFlagSlice to prevent k/k dep

stlaz:


stlaz:
looks like something broke in golang 1.15 and I'm missing some files so that I can `go mod` properly :thinking:

stlaz:
/hold cancel
the failure was caused by a partial modcache clean yesterday, full cache clean fixed the weird issue

soltysh (approved):
/lgtm
/retest

PR #351, opened 2020-09-17
Bug 1880345: Provide more details for deployment pods

mfojtik:


pkg/operator/starter.go
```diff
@@ -147,6 +148,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	operatorCtx.resourceSyncController = resourceSyncer
 	operatorCtx.operatorClient = operatorClient
 	operatorCtx.operatorConfigInformer = configinformer.NewSharedInformerFactoryWithOptions(configClient, resync)
+	operatorCtx.kubeInformers = informers.NewSharedInformerFactory(operatorCtx.kubeClient, 10*time.Minute)
```
sttts, line 151: why do we need kubeInformersForNamespaces and kubeInformers ?

sttts: be sure that you don't watch pods for all namespaces.

mfojtik:
/hold

the informers needs to be reconsidered

PR #325, opened 2020-08-24
bug 1871890: add current profile annotations to CVO manifests

deads2k:
This is matches https://github.com/openshift/enhancements/pull/414 and doesn't change existing behavior

PR #352, opened 2020-09-22
bug 1871890: add current profile annotations to CVO manifests

retroflexer:
Cluster profiles are a way to support different deployment models for OpenShift clusters. A profile is an identifier that the Cluster Version Operator uses to determine which manifests to apply. Operators can be excluded completely or can have different manifests for each supported profile.

To support above, a deployment model in which not all operators rendered by the CVO by default is needed. The use case includes IBM Public Cloud, in which a hosted control plane is used. Potentially it can also be used for Code Ready Containers.

The following annotation may be used to include manifests for a given profile:

```
include.release.openshift.io/[identifier]=true
```

This would make the CVO render this manifest only when CLUSTER_PROFILE=[identifier] has been specified.

This PR is just a rebase of the original PR https://github.com/openshift/cluster-authentication-operator/pull/325 created by @deads2k .

This matches https://github.com/openshift/enhancements/pull/414 and doesn't change existing behavior.

### DIFF
--- a/manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
+++ b/manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
@@ -4,6 +4,8 @@ kind: Role
 metadata:
   name: prometheus-k8s
   namespace: openshift-authentication-operator
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
 rules:
 - apiGroups:
   - ""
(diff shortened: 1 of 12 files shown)


stlaz:
/hold
fix the commit message so that it explains what's going on, I want to have a clear explanation about why the annotations are there when I look into the history, a link somewhere does not help me.

retroflexer:
> /hold
> fix the commit message so that it explains what's going on, I want to have a clear explanation about why the annotations are there when I look into the history, a link somewhere does not help me.

I updated the description in the PR. Hope that helps.

stlaz:
@retroflexer https://github.com/openshift/cluster-authentication-operator/pull/352/commits/0b5a52f82432c72f3aec5826c3c66d863a044df8 still shows the old commit message, did you forget to force-push? :)

PR #353, opened 2020-09-29
Bug 1883769: Add validation for logLevel

dmage:
openshift/api#736 added restrictions for logLevel values. This PR vendors new openshift/api to get new CRDs.


dmage:
/retitle Bug 1883769: Add validation for logLevel

PR #355, opened 2020-10-02
Add a controller to report misconfigured proxy

vareti:
This is an attempt to address https://issues.redhat.com/browse/MSTR-1011

Adds a new controller to report degraded conditions when the proxy is misconfigured.

### DIFF
--- a/pkg/controllers/proxyconfig/proxyconfig_controller.go
+++ b/pkg/controllers/proxyconfig/proxyconfig_controller.go
@@ -1,0 +1,203 @@
+package proxyconfig
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	"golang.org/x/net/http/httpproxy"
+
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	corev1lister "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
+
+	routeinformer "github.com/openshift/client-go/route/informers/externalversions/route/v1"
+	v1 "github.com/openshift/client-go/route/listers/route/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"github.com/openshift/library-go/pkg/route/routeapihelpers"
+)
+
+// proxyConfigChecker reports bad proxy configurations.
+type proxyConfigChecker struct {
+	caPool               *x509.CertPool
+	routeLister          v1.RouteLister
+	configMapLister      corev1lister.ConfigMapLister
+	routeName            string
+	routeNamespace       string
+	caConfigmapName      string
+	caConfigmapNamespace string
+	resourceVersion      string
+}
+
+func NewProxyConfigChecker(
+	routeInformer routeinformer.RouteInformer,
+	configMapInformer corev1informers.ConfigMapInformer,
+	routeNamespace string,
+	routeName string,
+	caConfigmapNamespace string,
+	caConfigmapName string,
+	recorder events.Recorder,
+	operatorClient v1helpers.OperatorClient) factory.Controller {
+	p := proxyConfigChecker{
+		routeLister:          routeInformer.Lister(),
+		configMapLister:      configMapInformer.Lister(),
+		routeName:            routeName,
+		routeNamespace:       routeNamespace,
+		caConfigmapName:      caConfigmapName,
+		caConfigmapNamespace: caConfigmapNamespace,
+	}
+
+	return factory.New().
+		WithSync(p.sync).
+		WithInformers(
+			routeInformer.Informer(),
+			configMapInformer.Informer(),
+		).
+		ResyncEvery(5*time.Minute).
+		WithSyncDegradedOnError(operatorClient).
+		ToController("ProxyConfigController", recorder.WithComponentSuffix("proxy-config-controller"))
+}
+
+// sync attempts to connect to route using configured proxy settings and reports any error.
+func (p *proxyConfigChecker) sync(ctx context.Context, _ factory.SyncContext) error {
+	proxyConfig := httpproxy.FromEnvironment()
+	if !isProxyConfigured(proxyConfig) {
+		// If proxy is not configured, then it is a no-op.
+		return nil
+	}
+
+	route, err := p.routeLister.Routes(p.routeNamespace).Get(p.routeName)
+	if err != nil {
+		return err
+	}
+
+	routeURL, _, err := routeapihelpers.IngressURI(route, "")
+	if err != nil {
+		return err
+	}
+	routeURL.Path = "healthz"
+
+	clientWithProxy, clientWithoutProxy, err := p.createHTTPClients()
+	if err != nil {
+		return err
+	}
+
+	return checkProxyConfig(ctx, routeURL, proxyConfig.NoProxy, clientWithProxy, clientWithoutProxy)
+}
+
+// checkProxyConfig determines any mis-configuration in proxy settings by attempting
+// to connect to endpoint directly and via proxy and comparing the results with expectations.
+func checkProxyConfig(ctx context.Context, endpointURL *url.URL, noProxy string, clientWithProxy, clientWithoutProxy *http.Client) error {
+	withProxyErr := isEndpointReachable(ctx, endpointURL.String(), clientWithProxy)
+	withoutProxyErr := isEndpointReachable(ctx, endpointURL.String(), clientWithoutProxy)
+	noProxyMatchesEndpoint := strings.Contains(noProxy, endpointURL.Hostname())
+
+	if noProxyMatchesEndpoint && withoutProxyErr != nil {
+		if withProxyErr == nil {
+			return fmt.Errorf("failed to reach endpoint(%q) found in NO_PROXY(%q) with error: %v", endpointURL.String(), noProxy, withoutProxyErr)
+		}
+		return fmt.Errorf("endpoint(%q) found in NO_PROXY(%q) is unreachable with proxy(%v) and without proxy(%v)", endpointURL.String(), noProxy, withProxyErr, withoutProxyErr)
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,134 @@
...
+
+	withProxySuccess := p.isEndpointReachable(ctx, "https://"+route.Spec.Host+"/healthz", p.proxyFunc)
+	withoutProxySuccess := p.isEndpointReachable(ctx, "https://"+route.Spec.Host+"/healthz", nil)
+	noProxyMatchesEndpoint := strings.Contains(p.proxyConfig.NoProxy, endpointURL.Hostname())
+
+	if noProxyMatchesEndpoint && !withoutProxySuccess {
+		if withProxySuccess {
+			return fmt.Errorf("endpoint(%q) found in NO_PROXY(%q) but can only be reached with proxy", endpointURL.Hostname(), p.proxyConfig.NoProxy)
+		}
+		return fmt.Errorf("endpoint(%q) found in NO_PROXY(%q) is not reachable with/without proxy", endpointURL.Hostname(), p.proxyConfig.NoProxy)
```
stlaz, line 70: maybe just `%q is unreachable`?

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	if !noProxyMatchesEndpoint && !withProxySuccess {
+		if withoutProxySuccess {
+			return fmt.Errorf("endpoint(%q) absent in NO_PROXY(%q) can only be reached without proxy", endpointURL.Hostname(), p.proxyConfig.NoProxy)
+		}
+		return fmt.Errorf("endpoint(%q) is not reachable with/without proxy", endpointURL.Hostname())
```
stlaz, line 77: maybe just `%q is unreachable`?

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	reqCtx, cancel := context.WithTimeout(ctx, 10*time.Second) // avoid waiting forever
+	defer cancel()
+	req.WithContext(reqCtx)
+
+	client := &http.Client{
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
+			Proxy: proxyFunc,
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
```
stlaz, line 98: what if the CA in proxyconfig is misconfigured?

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		klog.V(4).Infof("failed to parse https proxy %q", p.proxyConfig.HTTPSProxy)
+	}
+
+	proxyURL, err := url.Parse(p.proxyConfig.HTTPProxy)
+	if err != nil {
+		return nil, err
+	}
+	return proxyURL, nil
+}
```
stlaz, line 134: or you don't cache the httproxy.FromEnvironment result and just use the function instead

stlaz (commented):
a few comments and variable name nits

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	proxyConfig *httpproxy.Config
+	routeLister v1.RouteLister
+}
+
+func NewProxyConfigChecker(
+	routeInformer routeinformer.RouteInformer,
+	recorder events.Recorder,
+	operatorClient v1helpers.OperatorClient) factory.Controller {
+	p := proxyConfigChecker{
+		proxyConfig: httpproxy.FromEnvironment(),
```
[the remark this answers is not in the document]
vareti: @stlaz Thanks for the review. My understanding is that env variables are set at pod startup and cannot be changed later. Is there way to override them dynamically at runtime?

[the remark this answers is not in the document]
stlaz: That might be true but retrieving environment variables is no resource-heavy operation so there's no need to cache something that's actually already cached by the process.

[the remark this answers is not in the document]
vareti: I dont have a strong argument against yours.So, I updated the code to not cache it now.

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+
+	endpointURL, err := url.Parse(route.Spec.Host)
+	if err != nil {
+		return err
+	}
+
+	withProxySuccess := p.isEndpointReachable(ctx, "https://"+route.Spec.Host+"/healthz", p.proxyFunc)
```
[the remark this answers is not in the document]
vareti: I renamed them to `reachableWithProxy` and `reachableWithoutProxy`. Hopefully it is clear now.

pkg/controllers/proxyconfig/proxyconfig_controller.go
vareti: done

pkg/controllers/proxyconfig/proxyconfig_controller.go
vareti: done

pkg/controllers/proxyconfig/proxyconfig_controller.go
vareti: Good point. Now it uses ca to validate proxyconfig.

pkg/controllers/proxyconfig/proxyconfig_controller.go
vareti: I don't think we can use `httpproxy.FromEnvironment` here. With endpoint matches NO_PROXY, then this function would not use the proxy. But for one specific case, this controller should reach endpoint via proxy even if endpoint is present in NO_PROXY. Does that make sense?

stlaz: I thought this was more about trying to connect to the endpoint with and without the proxy configuration. Why would we try ignore specifically configured NO_PROXY?

vareti: Imagine the case where NO_PROXY is misconfigured with endpoint. In that case, connecting to an endpoint with proxy will help troubleshoot this issue. Otherwise, testing the endpoint without proxy will not be any different from testing with proxy and endpoint configured in NO_PROXY. Does that sound reasonable to you? It is possible that I misunderstood the requirement too.

stlaz: That makes sense

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, err
+	}
+
+	caPool := x509.NewCertPool()
+	if ok := caPool.AppendCertsFromPEM([]byte(caConfigmap.Data["ca-bundle.crt"])); !ok {
+		return nil, fmt.Errorf("unable to append system trust ca bundle")
+	}
+
+	return caPool, nil
+}
```
stlaz, line 153: if this is the whole bundle, you may not want to load it on each run. Maybe cache it based on a resource version?

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,200 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		klog.V(4).Infof("failed to make request to endpoint: %v", err)
+		return false
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		klog.V(4).Infof("%q returned %q", endpointURL, resp.Status)
+		return false
+	}
```
stlaz, line 175: consider keeping the errors to go into the status so that you've got more information right away without having to look into the logs

pkg/operator/starter.go
```diff
@@ -390,6 +391,17 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 		controllerContext.EventRecorder,
 	)
 
+	proxyConfigController := proxyconfig.NewProxyConfigChecker(
+		routeInformersNamespaced.Route().V1().Routes(),
+		operatorCtx.kubeInformersForNamespaces.InformersFor("openshift-authentication-operator").Core().V1().ConfigMaps(),
+		"oauth-openshift",
+		"openshift-authentication",
+		"trusted-ca-bundle",
+		"openshift-authentication-operator",
```
stlaz, line 400: nit: you may want to keep the order here as
```
stuff,
ns, name,
ns, name,
other_stuff
```

stlaz (commented):
small things, looks ok in general

stlaz (commented):
nits, I forgot to submit the review yesterday :facepalm:

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,205 @@
...
+		// If proxy is not configured, then it is a no-op.
+		return nil
+	}
+
+	route, err := p.routeLister.Routes(p.routeNamespace).Get(p.routeName)
+	if err != nil {
+		return err
+	}
+
+	_, ingress, err := routeapihelpers.IngressURI(route, "")
```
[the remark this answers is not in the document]
vareti: sounds better. made the change.

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,205 @@
...
+		routeName:            routeName,
+		routeNamespace:       routeNamespace,
+		caConfigmapName:      caConfigmapName,
+		caConfigmapNamespace: caConfigmapNamespace,
+	}
+
+	return factory.New().
+		WithSync(p.sync).
+		WithInformers(routeInformer.Informer()).
+		WithInformers(configMapInformer.Informer()).
```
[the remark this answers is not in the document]
vareti: done.

vareti:
@stlaz addressed the nits. please take a look.

PR #356, opened 2020-10-05
flowschema for oauth server and its operator

tkashem:
- `SAR` or `tokenreviews` from oauth 
- Any other requests from oauth server.
- Traffic from oauth operator.

See https://github.com/openshift/cluster-kube-apiserver-operator/pull/966 for more details.

### DIFF
--- a/manifests/09_flowschema.yaml
+++ b/manifests/09_flowschema.yaml
@@ -1,0 +1,89 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: FlowSchema
+metadata:
+  name: openshift-oauth-apiserver-sar
+spec:
+  distinguisherMethod:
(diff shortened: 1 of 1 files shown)


manifests/09_flowschema.yaml
```diff
@@ -0,0 +1,89 @@
...
+  distinguisherMethod:
+    type: ByUser
+  matchingPrecedence: 2000
+  priorityLevelConfiguration:
+    name: openshift-control-plane-operators
+  rules:
+  - resourceRules:
+    - apiGroups:
+      - '*'
+      clusterScope: true
```
deads2k, line 78: will this include all namespaced requrests too?

tkashem: Yes, that's how it was designed I think.

PR #361, opened 2020-10-12
Bug 1887428: add ServiceMonitor for the oauth apiserver

stlaz:


stlaz:
/bugzilla refresh
let's see whether it's just an accident or BZ API changed

manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
```diff
@@ -38,6 +38,24 @@ rules:
   - list
   - watch
 ---
+# oauth-apiserver role
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-oauth-apiserver
```
vareti, line 46: Is this annotation intentionally ignored from all the resources?

`include.release.openshift.io/self-managed-high-availability: "true"`

As I understand it, CVO would not create the resource if the annotation was missing in default cluster profile mode. Please correct me if I am wrong.

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -53,3 +53,46 @@ spec:
   selector:
     matchLabels:
       app: oauth-openshift
+---
```
vareti, line 56: I don't have enough expertise to review this file but I trust you with what you are doing.

vareti: A comment explaining why specific labels are dropped will really help a person like me who is  unfamiliar with these metrics.

vareti (commented):
Couple of nits otherwise look good to me.

manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
stlaz: I think that should still be in. Good catch!

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
stlaz: Honestly, I am just as unfamiliar with the metrics as you are. This is copied from the openshift-apiserver operator.

sttts:
Lgtm.

@tkashem @vareti for final lgtm

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -53,3 +53,48 @@ spec:
...
+      - __name__
+    - action: drop
+      regex: apiserver_request_duration_seconds_bucket;(0.15|0.25|0.3|0.35|0.4|0.45|0.6|0.7|0.8|0.9|1.25|1.5|1.75|2.5|3|3.5|4.5|6|7|8|9|15|25|30|50)
+      sourceLabels:
+      - __name__
+      - le
+    relabelings:
+    - action: replace
+      targetLabel: apiserver
+      replacement: openshift-oauth-apiserver
```
tkashem, line 89: This is good, when we have the api dashboard,  `openshift-oauth-apiserver` will show up as an option and we can view its metrics.

manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
```diff
@@ -53,3 +53,48 @@ spec:
...
+  namespace: openshift-oauth-apiserver
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    interval: 30s
+    metricRelabelings:
+    - action: drop
+      regex: etcd_(debugging|disk|request|server).*
```
tkashem, line 70: we can remove the `request` from the drop list. Similar to kube-apiserver `etcd_(debugging|disk|server).*`
https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml#L52

I realize that I didn't fix it for openshift-apiserver. https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/manifests/0000_90_openshift-apiserver-operator_04_servicemonitor-apiserver.yaml#L49. 

This is basically a histogram that shows the apiserver -> etcd latencies, comes handy in troubleshooting.

PR #365, opened 2020-10-27
Bug 1891758: deployment: update event spam

stlaz:
When authentication.operator observedGeneration and readyReplicas
get set before the observed deployment's generation, the last element
would have never gotten set and the ApplyDeployment function would have
kept reporting deployment update event endlessly.

### DIFF
--- a/pkg/controllers/deployment/deployment_controller.go
+++ b/pkg/controllers/deployment/deployment_controller.go
@@ -360,7 +360,9 @@ func (c *deploymentController) updateOperatorDeploymentInfo(
 	operatorConfig *operatorv1.Authentication,
 	deployment *appsv1.Deployment,
 ) error {
-	operatorStatusOutdated := operatorConfig.Status.ObservedGeneration != operatorConfig.Generation || operatorConfig.Status.ReadyReplicas != deployment.Status.UpdatedReplicas
+	operatorStatusOutdated := operatorConfig.Status.ObservedGeneration != operatorConfig.Generation ||
+		operatorConfig.Status.ReadyReplicas != deployment.Status.UpdatedReplicas ||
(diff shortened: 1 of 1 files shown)


pkg/controllers/deployment/deployment_controller.go
```diff
@@ -360,7 +360,9 @@ func (c *deploymentController) updateOperatorDeploymentInfo(
 	operatorConfig *operatorv1.Authentication,
 	deployment *appsv1.Deployment,
 ) error {
-	operatorStatusOutdated := operatorConfig.Status.ObservedGeneration != operatorConfig.Generation || operatorConfig.Status.ReadyReplicas != deployment.Status.UpdatedReplicas
+	operatorStatusOutdated := operatorConfig.Status.ObservedGeneration != operatorConfig.Generation ||
+		operatorConfig.Status.ReadyReplicas != deployment.Status.UpdatedReplicas ||
+		resourcemerge.ExpectedDeploymentGeneration(deployment, operatorConfig.Status.Generations) != deployment.Generation
```
sttts, line 365: this increases spam, doesn't it?

sttts: or maybe I don't get what the new line shoul dbe doing

stlaz: why would it? It compares the previous generation for the deployment with its current, and if they don't match, it sets them to be equal, therefore https://github.com/openshift/cluster-authentication-operator/blob/82e8f1801f879133199d4be9610c14f92f7cbe49/pkg/controllers/deployment/deployment_controller.go#L164-L165 no longer spams an update event in every loop.

sttts: got it

PR #366, opened 2020-10-27
Restrict log permissions

JAORMX:
While the current log directory permissions are restrictive and correct,
the file permissions are too permissive and it raises flags on
evaluations of the deployment. Let's instead change the permissions to
0600 which is more appropriate for these types of logs.


JAORMX:
This is similar to the fix proposed in openshift/cluster-openshift-apiserver-operator#397

JAORMX:
Pull-request updated, HEAD is now 0ffcc428a132c21f84ba4292fd5b00baf9d831ca

dtaylor113:
Apologies, once this [PR](https://github.com/openshift/console/pull/7053) merges it should resolve your CI issue.

stlaz:
/hold cancel
the PR merged

PR #368, opened 2020-10-30
Bug 1893386: Revert "workers: add controller that checks if router can schedule pods"

wking:
This reverts commit 88e6eba123b74c8fd73d282c8acc2d77772299e7, #344 and its follow-up ffcd51d18d41fd4b8418423881ffae8ba3e64064, #348.

The logic assumes the router will be scheduled on `worker`-labeled nodes.  That leads to false-positives when there are no vanilla `worker` compute nodes, but are schedulable [compute nodes that have custom names][1] (`infra`, `compute`, etc.).  Instead of trying to second-guess the scheduler and the ingress-operator, let [the ingress operator handle reporting this issue][3].

[1]: https://github.com/openshift/machine-config-operator/blob/0170e082a8b8228373bd841d17555fff2cfb51b7/docs/custom-pools.md
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=1881155


sttts:
/hold

We need a replacement. It cannot be that ingress operator is happy, but auth struggles because ingress is NOT ready.

wking:
my bug links [the bug series][1] where ingress adds its own check, so the replacement is already in place.

/hold cancel

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1881155

stlaz:
/bugzilla refresh
hasn't passed QA yet

wking:
Bot seems fine with the ON_QA blocker, but if you want a team call to block on it getting verified, please encourage QE to pick it up (it has been ON_QE almost a month already).

wking:
Also, this PR targets master.  Can we land if folks think it is a clean enough revert, and then auth and ingress/routing and QE can fight out the timing of the 4.6 backports?

deads2k:
/hold

This throws out the baby with the bathwater.  The message overall is incredibly useful.  You found a bug.  No reason to overreact.  

https://github.com/openshift/cluster-ingress-operator/blob/c6491b2c5d04f81aa36b3c409de1c49a5c228ae3/pkg/operator/controller/ingress/deployment.go#L343 is pretty clearly missed here.  A must-gather or insights data for your failing cluster would help confirm.

wking:
The message should be redundant since openshift/cluster-ingress-operator#465's more robust check.  Why would we want the auth operator *and* the ingress operator both watching the ingress operand long-term?

deads2k:
https://github.com/openshift/cluster-authentication-operator/pull/370 addresses the bug without removing value for tracking nodes (which the ingress operator does not do, it tracks scheduling messages)

/close

PR #371, opened 2020-11-03
[4.6] Bug 1893803: update ingress node available to handle custom placement

stlaz:


mfojtik:
/lgtm

overriding invalid-bug to speed up delivering this fix to unblock 4.6.1 upgrades (it has unit test attached and should be checked by QE anyway)

PR #373, opened 2020-11-04
Bug 1893386: Fix wrong operator config informer

mfojtik:
@deads2k @stlaz we pass a wrong lister to `NewIngressNodesAvailableController` which ONLY include operator config resources that match field selector "meta.name=cluster". This won't work for "ingress" config that has name "default"....


mfojtik:
/override ci/prow/e2e-agnostic 
/override ci/prow/e2e-operator-encryption

These flake on unrelated infra issues, QE verified this fix solve the problem the bug was adressing. Merging with button.

PR #375, opened 2020-11-05
Make ready ingress nodes degraded instead of unavailable

mfojtik:


mfojtik:
/close

we need this only in 4.6

deads2k:
I would follow the normal process here.  It can change further in 4.7, but this isn't so bad to have in 4.7

PR #380, opened 2020-11-23
always claim control over oauth and user APIs

stlaz:
/assign @p0lyn0mial 

depends on https://github.com/openshift/cluster-openshift-apiserver-operator/pull/417


stlaz:
/retest
weird build failures

pkg/operator/starter.go
```diff
@@ -654,3 +615,38 @@ func singleNameListOptions(name string) func(opts *metav1.ListOptions) {
...
+	}
+
+	ret := []*apiregistrationv1.APIService{}
+	for _, apiServiceGroupVersion := range apiServiceGroupVersions {
+		obj := &apiregistrationv1.APIService{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: apiServiceGroupVersion.Version + "." + apiServiceGroupVersion.Group,
+				Annotations: map[string]string{
+					"service.alpha.openshift.io/inject-cabundle":   "true",
+					"authentication.operator.openshift.io/managed": "true",
```
p0lyn0mial, line 633: I think we also don't need this annotation. It was used in OAS-O to check if it is safe to stop managing an API group. 
I think that we have at least two options:

`Option 1`: we remove it in 4.7 then:

In 4.6 the fight is over and CAO is responsible for these groups.
In 4.7 nothing changes.

An Upgrade from 4.6 to 4.7 will work
A downgrade from 4.7 to 4.6 might be a bit disruptive since CAO and OSA-O needs to decide which one is managing. 

`Option 2`: we remove it in 4.8 then a downgrade from 4.7 to 4.6 is smoother, no disruption.

pkg/operator/apiservices/sync_apiservices_oauth_apiserver.go
```diff
@@ -1,88 +0,0 @@
...
-//
-// TODO: change this function in 4.7 to return initial/full/authoritative list of APIs to manage
-func (a *APIServicesToManage) GetAPIServicesToManage() ([]*apiregistrationv1.APIService, error) {
-	authOperator, err := a.authOperatorLister.Get("cluster")
-	if err != nil {
-		klog.V(4).Infof("unable to determine if auth operator should take OAuth APIs over due to %v", err)
-		return nil, err
-	}
-
-	if !authOperator.Status.ManagingOAuthAPIServer {
```
p0lyn0mial, line 41: we also need to decide how to depreciate `ManagingOAuthAPIServer` field. 

The situation here is a bit different because we are going to remove it from the API so it will be pruned and removed from the database on an update.

We could remove it in 4.8 and set it to `true` in 4.7 so that a downgrade from 4.7 to 4.6 is smooth.

p0lyn0mial (commented):
I think we should also remove https://github.com/openshift/cluster-authentication-operator/blob/0859046d8df0d21976def7b3a51b376eeb80041c/pkg/operator/apiservices/sync_take_over_apiservices_oauth_apiserver.go#L26


which set `perator.Status.ManagingOAuthAPIServer` to `true`. This was necessary in 4.6 to decide which component is in charge, we don't need this anymore.

pkg/operator/apiservices/sync_apiservices_oauth_apiserver.go
stlaz: Makes sense

p0lyn0mial: we decided to remove it in `4.8` along with the annotation.

p0lyn0mial:
/hold 

shouldn't we merge https://github.com/openshift/cluster-openshift-apiserver-operator/pull/417 first ?

stlaz:
spot on, I thought that was already in... CI is obstructive once again

stlaz:
/hold cancel
the other PR merged

PR #383, opened 2020-11-26
Annotate manifests for single-node-developer cluster profile

guillaumerose:
This implements phase 1 of openshift/enhancements#482

### DIFF
--- a/manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
+++ b/manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
@@ -7,6 +7,7 @@ metadata:
   annotations:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
 rules:
 - apiGroups:
   - ""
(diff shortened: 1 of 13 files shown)


rkukura:
Do the manifests under vendor/github.com/openshift/api also need to be updated to pickup the profile annotations?

stlaz:
Please explain in details of what the annotation actually does in the commit message, link to an enhancement won't do for anyone who wants to quickly understand why the annotation's there.

I haven't read the enhancement and I do not know the consequences of this annotation. Single-node cluster sounds contradictive. Does that mean the operator's code will need to change? Who is going to implement the necessary changes?

guillaumerose:
I updated the commit message.

In a future release, cluster profile will be used to determine which manifests will be applied by the CVO. The default profile is `self-managed-high-availability`. 
CRC will use a new one, `single-node-developer` and edge production cluster  `single-node-production`.

> Does that mean the operator's code will need to change?

Yes. Right now, we don't modify operators but in the future we might have to open PRs to fine tune operators for our profile. 
For this operator, a good thing for us would be to customize https://github.com/openshift/cluster-authentication-operator/blob/master/manifests/02_config.cr.yaml and directly include `useUnsupportedUnsafeNonHANonProductionUnstableOAuthServer`. 

> Who is going to implement the necessary changes?

Teams that need it. For the single-node-developer profile, CRC team.

stlaz:
/hold cancel

> > Who is going to implement the necessary changes?

> Teams that need it. For the single-node-developer profile, CRC team.

I'm taking your word for it, then ;)

cfergeau:
I don't think the CI failures are related to this PR? Are there ongoing issues with CI?

PR #384, opened 2020-12-01
add 'single-node-production-edge' annotations to CVO manifests.

osherdp:
This PR adds annotation for the single-node-production-edge cluster profile. There's a growing requirement from several customers to enable creation of single-node (not high-available) Openshift clusters.
In stage one (following openshift/enhancements#504) there should be no implication on components logic.
In the next stage, the component's behavior will match a non high-availability profile if the customer is specifically interested in one.
This PR is separate from the 'single-node-developer' work, which will implement a different behavior and is currently on another stage of implementation.

For more info, please refer to the enhancement link and participate in the discussion.

### DIFF
--- a/manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
+++ b/manifests/0000_90_cluster-authentication-operator_01_prometheusrbac.yaml
@@ -8,6 +8,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    include.release.openshift.io/single-node-production-edge: "true"
 rules:
 - apiGroups:
   - ""
@@ -30,6 +31,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    include.release.openshift.io/single-node-production-edge: "true"
 rules:
 - apiGroups:
   - ""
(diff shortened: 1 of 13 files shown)


stlaz:
Please explain in details of what the annotation actually does in the commit message, link to an enhancement won't do for anyone who wants to quickly understand why the annotation's there.

I haven't read the enhancement and I do not know the consequences of this annotation. Single-node cluster sounds contradictive. Does that mean the operator's code will need to change? Who is going to implement the necessary changes?

How is this different from the other "single-node" PR? Why does it have to be 2 PRs instead of one with two commits?

osherdp:
@stlaz updated description, please take a look

stlaz:
@osherdp  Awesome, please also remove the JIRA reference from the commit message.

I am also concerned about
> In the next stage, the component's behavior will match a non high-availability profile if the customer is specifically interested in one.

What does it mean "will match a non high-availability profile"? CAO makes assumptions about high-availability of kube-apiserver as it may restart it as a reaction to user configuration. What happens to the platform when there is no kube-apiserver? How do you make this component "match a non high-availability profile"? Is that something that will require changes to the component? Who will implement them?

osherdp:
@stlaz not only CAO, but many components are doing restarts. We have MCO which also does a complete node reboot and we do know to handle that
I assume most changes are pretty trivial and can be made by our team (edge), some edge cases will need help from the responsible maintainers of the component

stlaz:
/hold cancel
/lgtm
@osherdp ok, I think that addresses all my concerns for now. Thanks.

osherdp:
@stlaz can you please re-lgtm?

stlaz:
/hold
`Merge branch 'master' into enhancement/single-node-annotation` <- remove this

osherdp:
@stlaz done
keep on hold until they'll merge api's PR

osherdp:
Shifting strategy based on https://github.com/openshift/enhancements/pull/560

PR #395, opened 2021-01-07
Bug 1905329: oauth-apiserver deployment: add cpu and memory requests

stlaz:
The number is arbitrary, modify if it proves wrong.

### DIFF
--- a/bindata/oauth-apiserver/deploy.yaml
+++ b/bindata/oauth-apiserver/deploy.yaml
@@ -39,6 +39,10 @@ spec:
           command: ['sh', '-c', 'chmod 0700 /var/log/oauth-apiserver && touch /var/log/oauth-apiserver/audit.log && chmod 0600 /var/log/oauth-apiserver/*']
           securityContext:
             privileged: true
+          resources:
+            requests:
+              cpu: 15m
+              memory: 50Mi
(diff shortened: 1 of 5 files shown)


bindata/oauth-apiserver/deploy.yaml
```diff
@@ -39,6 +39,10 @@ spec:
           command: ['sh', '-c', 'chmod 0700 /var/log/oauth-apiserver && touch /var/log/oauth-apiserver/audit.log && chmod 0600 /var/log/oauth-apiserver/*']
           securityContext:
             privileged: true
+          resources:
+            requests:
+              cpu: 15m
+              memory: 50Mi
```
sttts, line 45: what do we have in openshift-apiserver?

Have you checked a real cluster? Console should show values.

stlaz: This specific container just comes and goes so it's hard to retrieve what it really attempts to request. I wasn't able to get that with `oc adm top pod --containers`, nor with the console.

The openshift-apiserver does not seem to be setting the requests here either.

sttts: oh, didn't see it's not the main container. All fine.

PR #399, opened 2021-01-15
Bug 1916949: manifests: exclude rbac resources in openshift-oauth-apiserver namespace from hosted profile

sjenning:
@csrwng


sjenning:
had to add the `ServiceMonitor` as well

sttts:
This needs lgtm from monitoring team, who own this domain.

/assign @s-urbaniak

PR #401, opened 2021-01-27
Bug 1841039: e2e: add a test for password grant checking with keycloak

stlaz:
Finally, after all that time.

/assign @sttts


test/e2e/password_grant_check_test.go
```diff
@@ -32,3 +38,82 @@ func TestGitLabAsOIDCPasswordGrantCheck(t *testing.T) {
 
 	require.Equal(t, true, gitlabIDPConfig.UseAsChallenger, "gitlab is not configured as challenger")
 }
+
+func TestKeycloakAsOIDCPasswordGrantCheck(t *testing.T) {
+	testContext := context.Background()
```
sttts, line 43: defer cancel ?

test/e2e/password_grant_check_test.go
```diff
@@ -32,3 +38,82 @@ func TestGitLabAsOIDCPasswordGrantCheck(t *testing.T) {
...
+	// bump the configured secret since the operator caches the password grant check
+	// based on it
+	configSecretName := configIDP.OpenID.ClientSecret.Name
+	configSecret, err := kubeClients.CoreV1().Secrets("openshift-config").Get(testContext, configSecretName, v1.GetOptions{})
+	require.NoError(t, err)
+
+	if configSecret.Annotations == nil {
+		configSecret.Annotations = map[string]string{}
+	}
+	configSecret.Annotations["bumped"] = "true"
```
sttts, line 105: random value?

test/e2e/password_grant_check_test.go
```diff
@@ -32,3 +38,82 @@ func TestGitLabAsOIDCPasswordGrantCheck(t *testing.T) {
...
+	_, err = kubeClients.CoreV1().Secrets("openshift-config").Update(testContext, configSecret, v1.UpdateOptions{})
+	require.NoError(t, err)
+
+	err = library.WaitForOperatorToPickUpChanges(t, configClient, "authentication")
```
sttts, line 109: is this race free to wait after the change above just a moment ago?

stlaz: I don't think it's exactly race free, but until now I think it has always worked.

stlaz: but it seems that perhaps this time, it fails in CI, yet it never failed for me. I'll try to look for a better way.

slaskawi:
There's a tracking ticket for this on Keycloak side: https://issues.redhat.com/browse/KEYCLOAK-14886

test/e2e/password_grant_check_test.go
```diff
@@ -32,3 +40,83 @@ func TestGitLabAsOIDCPasswordGrantCheck(t *testing.T) {
...
+	var configIDP configv1.IdentityProvider
+	for _, idp := range oauthConfig.Spec.IdentityProviders {
+		if idp.Name == idpName {
+			configIDP = idp
+			break
+		}
+	}
+
+	// Get a keycloak client for the external KC URL
+	// skip TLS verify on the route's cert
```
slaskawi, line 83: This might be a bit risky I think.

Keycloak generates it own, sef-signed cert if none is given (see [Keycloak container manual](https://github.com/keycloak/keycloak-containers/tree/master/server#setting-up-tlsssl)). The route pointing to Keycloak can support either Passthrough or Re-encrypt TLS Termination. So it's hard to tell what certificate (the one on the route or the one from Keycloak) you're ignoring here. 

Is there any way to supply a proper, OpenShift-generated certificate to Keycloak (see [how our Operator does it](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/keycloak_service.go#L21)) and use Re-Encrypt on the route? This way all certs should be fine.

test/library/keycloakidp.go
```diff
@@ -0,0 +1,374 @@
...
+
+	configClient, err := configv1client.NewForConfig(kubeconfig)
+	require.NoError(t, err)
+
+	// configclients, err := configv1client.NewForConfig(kubeconfig)
+	// require.NoError(t, err)
+
+	nsName, keycloakHost, cleanup := deployPod(t, kubeClients, routeClient,
+		"keycloak",
+		"docker.io/jboss/keycloak:latest",
```
slaskawi, line 43: We prefer to use Quay: https://quay.io/repository/keycloak/keycloak

test/library/keycloakidp.go
```diff
@@ -0,0 +1,374 @@
...
+		[]corev1.VolumeMount{
+			{
+				Name:      "certkeypair",
+				MountPath: "/etc/x509/https",
+				ReadOnly:  true,
+			},
+		},
+		corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				"cpu":    resource.MustParse("500m"),
```
slaskawi, line 70: You might want to use more CPU for better bootup time. I usually use 1 vCPU.

The memory could be adjusted as well. For a 1 user installation probably 600-700 MB should be enough.

slaskawi (commented):
Most of the bits looks good to me. Great work!

One thing I'm curious though - could we use Keycloak Operator here? Relying solely on the Docker image might be a bit limiting for the future.

test/e2e/password_grant_check_test.go
stlaz: I can add a trust to the router certificate instead, although it did not seem necessary for the test.

test/library/keycloakidp.go
stlaz: Thanks!

test/library/keycloakidp.go
stlaz: Good to know. From me playing with this, I could already see Keycloak booting pretty fast (compared to GitLab from the other test, but GitLab ofc brings way more than just an OIDC), but I'll ajdust these.

test/e2e/password_grant_check_test.go
stlaz: actually, I can see that the operator went degraded, but this should otherwise still be safe. I added degraded reporting for this and I'll check why that is.

stlaz:
@slaskawi for this test, I'd rather go as simple as possible, avoid the operator and use the image instead.

Perhaps the experience I got with configuring Keycloak here could later be used if we wanted to test Keycloak as a replacement to the oauth-server (which is not the case here yet), and perhaps we could then expand this test to leverage the operator's capabilities. I think this could actually be a pretty good springboard to that kind of future testing.

test/e2e/password_grant_check_test.go
stlaz: Found the issue, I forgot to wire gitlab cleanup with the new refactor. This method of checking whether the operator picked up the changes still seems safe, then.

test/e2e/password_grant_check_test.go
stlaz: Fixed, it was actually simpler to do it that way

stlaz:
/retest
lets' see whether infra healed over the weekend

sttts:
Approved.

@slaskawi can you lgtm?

slaskawi:
Yep, LGTM!

stlaz:
/bugzilla refresh

Thank you, Seb! :slightly_smiling_face:

PR #402, opened 2021-01-28
Bug 1921275: deploymentController: treat error retrieving authentication.operator/cluster

stlaz:
This could cause panics in a rare case where the resource cannot be
found

/assign @sttts


stlaz:
/retest
infra

PR #308, opened 2020-07-29
follow up changes to clean up endpoint accessible controller

vareti:
this is a follow up change set to clean up the changes made by this [PR](https://github.com/openshift/cluster-authentication-operator/pull/306). It removes the duplicate health checks to route and oauth endpoints. It also adds support for using ca data when making a connection to an endpoint.

PR #404, opened 2021-02-04
follow up changes to clean up endpoint accessible controller

stlaz:
Picked up https://github.com/openshift/cluster-authentication-operator/pull/308, rebased, added Proxy and Timeout to the tested client


pkg/libs/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -55,6 +58,10 @@ func (c *endpointAccessibleController) sync(ctx context.Context, syncCtx factory
 		return err
 	}
 
+	client, err := c.buildTLSClient()
```
p0lyn0mial, line 61: this is quite ineffective as we are building the HTTP client on every `sync` loop.
You don't want to reuse connections?

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -72,17 +79,6 @@ func (c *endpointAccessibleController) sync(ctx context.Context, syncCtx factory
...
-				Timeout: 5 * time.Second,
-				Transport: &http.Transport{
-					Proxy: http.ProxyFromEnvironment,
-					TLSClientConfig: &tls.Config{
-						InsecureSkipVerify: true,
-					},
-				},
-			}
-
 			resp, err := client.Do(req)
```
p0lyn0mial, line 82: close the resp.body if err != nil

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -72,17 +79,6 @@ func (c *endpointAccessibleController) sync(ctx context.Context, syncCtx factory
 			defer cancel()
```
p0lyn0mial, line 79: > reqCtx, cancel := context.WithTimeout(ctx, 10*time.Second) // avoid waiting forever

does setting the request timeout has any effect since you are setting the timeout on the HTTP client (Timeout: 5 * time.Second)

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -20,23 +19,27 @@ import (
...
+	getTLSConfigFn         EndpointTLSConfigFunc
 	availableConditionName string
 }
 
 type EndpointListFunc func() ([]string, error)
+type EndpointTLSConfigFunc func() (*tls.Config, error)
 
 // NewEndpointAccessibleController returns a controller that checks if the endpoints
 // listed by endpointListFn are reachable
 func NewEndpointAccessibleController(
```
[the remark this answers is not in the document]
p0lyn0mial: we could reuse https://github.com/openshift/library-go/blob/master/pkg/monitor/health/prober.go#L31

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+	}
+
+	return &tls.Config{
+		RootCAs: rootCAs,
+	}, nil
+}
+
+func getOAuthEndpointTLSConfig(secretLister corev1listers.SecretLister, recorder events.Recorder) (*tls.Config, error) {
+	serviceSecret, err := secretLister.Secrets("openshift-authentication").Get("v4-0-config-system-serving-cert")
```
p0lyn0mial, line 202: shouldn't we load a public cert of the authority that signed the serving cert for `oauth-server` instead?

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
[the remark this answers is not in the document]
stlaz: > mhm, this controller could be moved to library-go as a generic way of checking if an endpoint is "OK"
that was the original idea I believe (notice that this is @vareti's work originally)

I wouldn't mind moving to another lib-go mechanism, although I'd prefer that as a follow-up to this PR when everything is in place as Siva originally meant it to be and I can do a 1:1 comparisons to the other mechanism

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,234 @@
...
+
+func listOAuthRoutes(routeLister routev1listers.RouteLister, recorder events.Recorder) ([]string, error) {
+	var results []string
+	route, err := routeLister.Routes("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		recorder.Warningf("OAuthRouteCheck", "failed to get oauth route: %v", err)
+		return nil, err
+	}
+	for _, ingress := range route.Status.Ingress {
+		if len(ingress.Host) > 0 {
```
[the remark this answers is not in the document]
stlaz: good point

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,234 @@
...
+		// so accessing the endpoint via IP would otherwise result
+		// in validation failure.
+		ServerName: "oauth-openshift.openshift-authentication.svc",
+	}, nil
+}
+
+func toHealthzURL(urls []string) []string {
+	var res []string
+	for _, url := range urls {
+		res = append(res, "https://"+url+"/healthz")
```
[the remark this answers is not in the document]
stlaz: this is a reverse operation to `url.Parse`. `url.Parse` would actually fail on the `url` variable here since it's not going to contain any scheme.

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -131,10 +127,22 @@ func (c *endpointAccessibleController) sync(ctx context.Context, syncCtx factory
...
 
-func toHealthzURL(urls []string) []string {
-	var res []string
-	for _, url := range urls {
-		res = append(res, "https://"+url+"/healthz")
+func (c *endpointAccessibleController) buildTLSClient() (*http.Client, error) {
+	transport := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
stlaz: that's the case here, isn't it? Given the code below

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
stlaz: I wouldn't reuse connections, shouldn't the point be that we want to make sure new connection can be created?

stlaz: but we might reuse clients to some degree, possibly caching on the TLS config?

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	return toHealthzURL(results), nil
+}
+
+func listOAuthRoutes(routeLister routev1listers.RouteLister, recorder events.Recorder) ([]string, error) {
+	var results []string
+	route, err := routeLister.Routes("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		recorder.Warningf("OAuthRouteCheck", "failed to get oauth route: %v", err)
+		return nil, err
```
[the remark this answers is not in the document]
stlaz: maybe going "available=false" would make more sense? The error is likely to be NotFound

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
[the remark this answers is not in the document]
p0lyn0mial: ah, it gets overwritten

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
p0lyn0mial: yeah, seems that the TLS config won't change very often

p0lyn0mial:
sidenote: we have probes (kublet) and the endpoint controller (KCM) yet we still write code to monitor some components.

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
stlaz: I guess not much, I guess just making sure? Perhaps @vareti would know why he did that, but we may never know

stlaz:
first batch of fixes

stlaz:
second batch of fixes, that should now be all comments addressed

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, fmt.Errorf("route status does not have host address")
+	}
+	return toHealthzURL(results), nil
+}
+
+func getOAuthRouteTLSConfig(cmLister corev1listers.ConfigMapLister, secretLister corev1listers.SecretLister, ingressLister configv1lister.IngressLister, systemCABundle []byte, recorder events.Recorder) (*tls.Config, error) {
+	// get default router CA cert cm
+	defaultIngressCertCM, err := cmLister.ConfigMaps("openshift-config-managed").Get("default-ingress-cert")
+	if err != nil {
+		recorder.Warningf("OAuthRouterCACerts", "failed to retrieve the default router CA certs: %v", err)
```
p0lyn0mial, line 185: we usually fire events on state changes, this is not a state change

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, err
+	}
+
+	defaultRouterCAPEM, ok := defaultIngressCertCM.Data["ca-bundle.crt"]
+	if !ok {
+		klog.Info("the openshift-config-managed/default-ingress-cert CM does not contain the \"ca-bundle.crt\" key")
+		return nil, err
```
p0lyn0mial, line 192: create a new `err` and return.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	ingress, err := ingressLister.Get("cluster")
+	if err != nil {
+		recorder.Warningf("OAuthRouteSecret", "failed to get ingress config: %v", err)
```
p0lyn0mial, line 197: same as above.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, err
+	}
+	if len(ingress.Spec.Domain) == 0 {
+		return nil, fmt.Errorf("ingress config domain cannot be empty")
+	}
+
+	routerSecret, err := secretLister.Secrets("openshift-authentication").Get("v4-0-config-system-router-certs")
+	if err != nil {
+		recorder.Warningf("OAuthRouteSecret", "failed to get oauth route ca cert: %v", err)
```
p0lyn0mial, line 206: log instead.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, err
+	}
+
+	// find the domain that matches our route
+	routerCertKey, ok := routerSecret.Data[ingress.Spec.Domain]
+	if !ok {
+		klog.Infof("unable to find router certs for domain %s", ingress.Spec.Domain)
```
p0lyn0mial, line 213: if this is a normal state then we shouldn't log at all

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, nil
+	}
+
+	rootCAs := x509.NewCertPool()
+	if ok := rootCAs.AppendCertsFromPEM([]byte(defaultRouterCAPEM)); !ok {
+		klog.Infof("the default ingress CA bundle did not contain any PEM certificates %s", defaultRouterCAPEM)
```
p0lyn0mial, line 219: too verbose

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, nil
+	}
+
+	if ok := rootCAs.AppendCertsFromPEM(routerCertKey); !ok {
+		klog.Infof("failed to parse router certs for domain %s", ingress.Spec.Domain)
```
p0lyn0mial, line 224: same here

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, nil
```
p0lyn0mial, line 225: should this be "fatal"? We still have `defaultRouterCAPEM`

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+
+func getOAuthRouteTLSConfig(cmLister corev1listers.ConfigMapLister, secretLister corev1listers.SecretLister, ingressLister configv1lister.IngressLister, systemCABundle []byte, recorder events.Recorder) (*tls.Config, error) {
+	// get default router CA cert cm
+	defaultIngressCertCM, err := cmLister.ConfigMaps("openshift-config-managed").Get("default-ingress-cert")
+	if err != nil {
+		recorder.Warningf("OAuthRouterCACerts", "failed to retrieve the default router CA certs: %v", err)
+		return nil, err
+	}
+
+	defaultRouterCAPEM, ok := defaultIngressCertCM.Data["ca-bundle.crt"]
```
p0lyn0mial, line 189: move down where it is actually used.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+	for _, ingress := range route.Status.Ingress {
+		if len(ingress.Host) > 0 {
+			for _, condition := range ingress.Conditions {
+				if condition.Type == routev1.RouteAdmitted && condition.Status == corev1.ConditionTrue {
+					results = append(results, ingress.Host)
+					break
+				}
+			}
+
```
p0lyn0mial, line 171: rm empty line

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+func listOAuthServices(serviceLister corev1listers.ServiceLister, recorder events.Recorder) ([]string, error) {
+	var results []string
+	service, err := serviceLister.Services("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		recorder.Warningf("OAuthServiceCheck", "failed to get oauth service: %v", err)
+		return nil, err
+	}
+	for _, port := range service.Spec.Ports {
+		results = append(results, net.JoinHostPort(service.Spec.ClusterIP, strconv.Itoa(int(port.Port))))
```
p0lyn0mial, line 148: should we take only ports with `https` name?

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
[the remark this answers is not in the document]
p0lyn0mial: was thinking about `url.Parse(fmt.Sprintf("https://%s/%s", url, "healthz"))`, it would fail iff `url` for some reason would be broken.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	endpointsListFunc := func() ([]string, error) {
+		return listOAuthServices(kubeInformersForTargetNS.Core().V1().Services().Lister(), recorder)
+	}
+
+	getTLSConfigFunc := func() (*tls.Config, error) {
+		return getOAuthEndpointTLSConfig(kubeInformersForTargetNS.Core().V1().ConfigMaps().Lister(), recorder)
+	}
+
+	return endpointaccessible.NewEndpointAccessibleController(
+		"OAuthServerService",
```
p0lyn0mial, line 84: why not combine `OAuthServerService` and `OAuthServerServiceEndpoints`?

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: it's not expected, I'll move it to `V(5)`

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: It's usually not a parsing error, just none certificates are present in the bundle. I'll improve the message.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: I'd be inclusive

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: they can be perceived as two different layers

pkg/controllers/targetversion/target_version_controller.go
```diff
@@ -134,19 +101,6 @@ func (c *targetVersionController) sync(ctx context.Context, syncContext factory.
 	return common.UpdateControllerConditions(c.operatorClient, knownConditionNames, foundConditions)
 }
 
-func (c *targetVersionController) getRouteSecret() (*corev1.Secret, []operatorv1.OperatorCondition) {
-	routerSecret, err := c.secretLister.Secrets("openshift-authentication").Get("v4-0-config-system-router-certs")
-	if err != nil {
-		return nil, []operatorv1.OperatorCondition{{
-			Type:    "OAuthVersionRouteSecretDegraded",
```
p0lyn0mial, line 141: should we add this to the `staleConditions` list ?

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,268 @@
...
+		},
+		recorder)
+}
+
+func listOAuthServiceEndpoints(endpointsLister corev1listers.EndpointsLister, recorder events.Recorder) ([]string, error) {
+	var results []string
+	endpoints, err := endpointsLister.Endpoints("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		recorder.Warningf("OAuthServiceEndpointsCheck", "failed to get oauth service endpoints: %v", err)
+		return results, nil
```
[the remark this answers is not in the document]
stlaz: sorry, I forgot this other function was doing the same :facepalm:

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	defaultRouterCAPEM, ok := defaultIngressCertCM.Data["ca-bundle.crt"]
+	if !ok {
+		return nil, fmt.Errorf("the openshift-config-managed/default-ingress-cert CM does not contain the \"ca-bundle.crt\" key")
+	}
+
+	rootCAs := x509.NewCertPool()
+	if ok := rootCAs.AppendCertsFromPEM([]byte(defaultRouterCAPEM)); !ok {
+		klog.V(5).Infof("the default ingress CA bundle did not contain any PEM certificates %s", defaultRouterCAPEM)
```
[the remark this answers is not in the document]
stlaz: I guess we'd rather not

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	if len(results) == 0 {
+		recorder.Warningf("OAuthRouteCheck", "route status does not have host address")
+		return nil, fmt.Errorf("route status does not have host address")
+	}
+	return toHealthzURL(results), nil
+}
+
+func getOAuthRouteTLSConfig(cmLister corev1listers.ConfigMapLister, secretLister corev1listers.SecretLister, ingressLister configv1lister.IngressLister, systemCABundle []byte, recorder events.Recorder) (*tls.Config, error) {
+	// get default router CA cert cm
+	defaultIngressCertCM, err := cmLister.ConfigMaps("openshift-config-managed").Get("default-ingress-cert")
```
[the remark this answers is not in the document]
stlaz: I'd rather keep the function separated into two parts where in the first you retrieve stuff, in the second you work with it. Makes it way better readable to me.

Besides, this first batch of variables are just pointers to already existing structures so we don't need to be that careful around what we're initializing.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	rootCAs := x509.NewCertPool()
+	if ok := rootCAs.AppendCertsFromPEM([]byte(defaultRouterCAPEM)); !ok {
+		klog.V(5).Infof("the default ingress CA bundle did not contain any PEM certificates %s", defaultRouterCAPEM)
+		return nil, nil
+	}
+
+	// find the domain that matches our route
+	routerCertKey, ok := routerSecret.Data[ingress.Spec.Domain]
+	if !ok {
+		klog.V(5).Infof("unable to find router certs for domain %s", ingress.Spec.Domain)
```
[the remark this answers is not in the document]
stlaz: ingress-operator malfunction/new feature we don't keep track of

pkg/controllers/targetversion/target_version_controller.go
stlaz: that's a good observation, yes

pkg/operator/starter.go
```diff
@@ -655,3 +644,27 @@ func apiServices() []*apiregistrationv1.APIService {
...
+	if err != nil {
+		// this may fail route-health checks in proxy environments
+		klog.Warningf("Unable to read system CA from /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem: %v", err)
+		return nil, nil // trust noone
+	}
+
+	// test that the cert pool actually contains certs
+	certPool := x509.NewCertPool()
+	if ok := certPool.AppendCertsFromPEM(systemCABundle); !ok {
+		return nil, fmt.Errorf("no PEM certificates found in the system trust store (/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem)")
```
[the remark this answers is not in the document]
stlaz: The latter means there could be unexpected data in the file. The line is indeed thin, I wouldn't mind going either direction.

pkg/controllers/targetversion/target_version_controller.go
```diff
@@ -51,68 +45,41 @@ var knownConditionNames = sets.NewString(
...
-		configInformers.Config().V1().Ingresses().Informer(),
-		routeInformer.Informer(),
+		kubeInformersNamespaced.Core().V1().Pods().Informer(),
 	).WithSync(c.sync).ToController("TargetVersion", recorder.WithComponentSuffix("target-version-controller"))
 }
 
 func (c *targetVersionController) sync(ctx context.Context, syncContext factory.SyncContext) error {
 	foundConditions := []operatorv1.OperatorCondition{}
 
-	ingressConfig, ingressConditions := common.GetIngressConfig(c.ingressLister, "OAuthVersionIngressConfig")
```
stlaz, line 100: this needs to go to the stale conditions, too

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,268 @@
...
+
+	ingress, err := ingressLister.Get("cluster")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ingress config: %w", err)
+	}
+	if len(ingress.Spec.Domain) == 0 {
+		return nil, fmt.Errorf("ingress config domain cannot be empty")
+	}
+
+	routerSecret, err := secretLister.Secrets("openshift-authentication").Get("v4-0-config-system-router-certs")
```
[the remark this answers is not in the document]
stlaz: as above

pkg/operator/starter.go
```diff
@@ -237,23 +239,9 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
...
-			"WellKnownRouteDegraded",
-			"WellKnownAuthConfigDegraded",
-			"WellKnownProgressing",
-			"OperatorSyncDegraded",
-			"RouteHealthDegraded",
-			"RouteStatusDegraded",
-			"OAuthServerAvailable",
+			"OAuthRouteCheckEndpointAccessibleControllerAvailable",
+			"OAuthServiceCheckEndpointAccessibleControllerAvailable",
+			"OAuthServiceEndpointsCheckEndpointAccessibleControllerAvailable",
```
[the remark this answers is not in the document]
stlaz: absolutely, added that

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	var results []string
+	route, err := routeLister.Routes("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get oauth route: %w", err)
+	}
+	for _, ingress := range route.Status.Ingress {
+		if len(ingress.Host) > 0 {
+			for _, condition := range ingress.Conditions {
+				if condition.Type == routev1.RouteAdmitted && condition.Status == corev1.ConditionTrue {
+					results = append(results, ingress.Host)
```
p0lyn0mial, line 164: nit: we could return directly, since we break on first match

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, fmt.Errorf("no valid oauth services found")
+	}
+	return toHealthzURL(results), nil
+}
+
+func listOAuthRoutes(routeLister routev1listers.RouteLister) ([]string, error) {
+	var results []string
+	route, err := routeLister.Routes("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get oauth route: %w", err)
```
p0lyn0mial, line 158: you sure it will be correctly interpreted by `NewEndpointAccessibleController` ?

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, fmt.Errorf("oauth service endpoints are not ready")
+	}
+	return toHealthzURL(results), nil
+}
+
+func listOAuthServices(serviceLister corev1listers.ServiceLister) ([]string, error) {
+	var results []string
+	service, err := serviceLister.Services("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get oauth service: %w", err)
```
p0lyn0mial, line 143: > you sure it will be correctly interpreted by NewEndpointAccessibleController ?

same applies here

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,263 @@
...
+			kubeInformersForTargetNS.Core().V1().ConfigMaps().Informer(),
+		},
+		recorder)
+}
+
+func listOAuthServiceEndpoints(endpointsLister corev1listers.EndpointsLister) ([]string, error) {
+	var results []string
+	endpoints, err := endpointsLister.Endpoints("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get oauth service endpoints: %w", err)
```
p0lyn0mial, line 124: > you sure it will be correctly interpreted by NewEndpointAccessibleController ?

here too

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	for _, subset := range endpoints.Subsets {
+		for _, address := range subset.Addresses {
+			for _, port := range subset.Ports {
+				results = append(results, net.JoinHostPort(address.IP, strconv.Itoa(int(port.Port))))
+			}
+		}
+	}
+	if len(results) == 0 {
+		return nil, fmt.Errorf("oauth service endpoints are not ready")
```
p0lyn0mial, line 134: this will be `degraded` not `progressing`, is that okay?

p0lyn0mial: if yes, then we could update the msg

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: probably not, I'll remove the wrapper

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: there's another wrapping `for` loop

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: I think I'm currently fine with both the message and the status change, we can see if it actually ever becomes an issue and fix it later if it is

stlaz:
/retest
none of the failures seems related

stlaz:
/retest
pods are failing to create sandboxes

PR #407, opened 2021-02-09
Remove the 'authentication.operator.openshift.io/managed' annotation from user,oauth apiservices

stlaz:
/assign @p0lyn0mial


p0lyn0mial:
/lgtm

https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md has been fully implemented.

I have also opened https://github.com/openshift/cluster-authentication-operator/pull/409

PR #409, opened 2021-02-10
removes the encryption provider

p0lyn0mial:
In 4.7 we switched CAO to manage its own encryption configuration. 

That means we can replace a dynamic provider with one that maintains a static list.


pkg/operator/starter.go
```diff
@@ -438,6 +438,18 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 	return nil
 }
 
+type encryptionProvider []schema.GroupResource
```
p0lyn0mial, line 441: /hold

I'm going to move it to library-go.

A static provider will be used by oas, cao and kas-o

p0lyn0mial: https://github.com/openshift/library-go/pull/992

p0lyn0mial: merged.

/hold cancel

pkg/operator/starter.go
```diff
@@ -28,12 +28,29 @@ import (
...
+	oauthapiconfigobservercontroller "github.com/openshift/cluster-authentication-operator/pkg/operator/configobservation"
+	"github.com/openshift/cluster-authentication-operator/pkg/operator/revisionclient"
+	"github.com/openshift/cluster-authentication-operator/pkg/operator/workload"
 	"github.com/openshift/library-go/pkg/authentication/bootstrapauthenticator"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	libgoassets "github.com/openshift/library-go/pkg/operator/apiserver/audit"
 	workloadcontroller "github.com/openshift/library-go/pkg/operator/apiserver/controller/workload"
 	apiservercontrollerset "github.com/openshift/library-go/pkg/operator/apiserver/controllerset"
 	libgoetcd "github.com/openshift/library-go/pkg/operator/configobserver/etcd"
+	"github.com/openshift/library-go/pkg/operator/encryption"
```
p0lyn0mial, line 53: this is new, the rest was moved up.

stlaz: nit: please keep the local deps in a separate block

stlaz (changes requested):

pkg/operator/starter.go
p0lyn0mial: it will be hard to keep track since the local deps match "github.com/opeshisft" prefix

p0lyn0mial: anyway, I moved them to the bottom of the import statement :)

PR #410, opened 2021-02-22
[proof] bump lib-go for refactored workloads controller

stlaz:


stlaz:
/close
the PR that was being proved merged, the bump is going to be a part of https://github.com/openshift/cluster-authentication-operator/pull/414

PR #411, opened 2021-02-23
e2e: new gitlab version broke initial root user login

stlaz:
I couldn't find a specific change that would break this in the release notes. It is possible they started checking `client_id` even for these logins.


stlaz:
/retest
infra

stlaz:
/close
this was fixed in the #413, we'll need a BZ to investigate what's failing

PR #416, opened 2021-03-03
Bug 1965929: wire certs as the authentication means to openshift webhook authenticator

stlaz:
This is replacing the current authentication means KAS is using for webhook token authenticators (SA token) with client certificates. As a result, a client-auth cert needs to be requested using the CSR API and then placed into the kubeconfig the kube-apiserver will eventually use to reach out to the oauth-apiserver.

We're doing this in order to prevent additional requests from oauth-apiserver to KAS during authentication in order to validate the SA token.


stlaz:
Apparently, the `rsa.PublicKey` is missing the `Equal()` method in golang 1.15 :neutral_face:

stlaz:
/retest
infra

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	_, err = c.csrClient.Create(ctx,
+		&certapiv1.CertificateSigningRequest{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: csrName,
+			},
+			Spec: certapiv1.CertificateSigningRequestSpec{
+				Request:    csrPEM,
+				SignerName: certapiv1.KubeAPIServerClientSignerName,
```
p0lyn0mial, line 293: @stlaz how will `oauth-apiserver` verify a cert signed by `KubeAPIServerClientSignerName`?

Haven't seen it accepting a custom client CA.

stlaz: that's not a custom CA, it's the CA already used for client certificate verification

p0lyn0mial: Haven't seen `oauth-apiserver` accepting any CA :)

stlaz: https://github.com/openshift/oauth-apiserver/blob/44c67e17695c437ea96c3715f498728077b60322/vendor/k8s.io/apiserver/pkg/server/options/authentication.go#L305-L313

p0lyn0mial: thanks, didn't know it contains `client-ca-file`

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+		key, err = rsa.GenerateKey(rand.Reader, 4096)
+		if err != nil {
+			return nil, nil, false, fmt.Errorf("failed to generate a private key: %w", err)
+		}
+		keyPEM, err := pemEncodeRSAPrivate(key)
+		if err != nil {
+			return nil, nil, false, err
+		}
+
+		if err := c.secretClient.Delete(ctx, certSecretName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
p0lyn0mial, line 223: is this secret immutable? Why not to `Update` ?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	if err = c.resubmitCSR(ctx, key); err != nil {
+		return nil, err
+	}
+
+	return pbool(true), nil
+}
+
+func (c *webhookAuthenticatorCertRequester) getKeyCertFromSecret(ctx context.Context) (key *rsa.PrivateKey, cert *x509.Certificate, keyRotated bool, err error) {
```
p0lyn0mial, line 192: this becomes more like `getAndCreate`

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+	certSecret, err := c.secretLister.Secrets("openshift-oauth-apiserver").Get(certSecretName)
```
p0lyn0mial, line 193: how often gets this method called?

if the informer is slow we might be regenerating the key.

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+			v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:    "AuthenticatorCSRProgressing",
+				Status:  progressingStatus,
+				Reason:  progressingReason,
+				Message: progressingText,
+			}))
+
+		if statusErr != nil {
+			if err != nil {
+				klog.Error("failed to update operator status: %v", err)
```
p0lyn0mial, line 100: pass `statusErr` and remove the condition

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+		} else if !key.PublicKey.Equal(cert.PublicKey.(*rsa.PublicKey)) {
+			klog.Error("the public key of the certificate used for the openshift-authenticator does not match the stored private key, new certificate will be requested: %v", err)
+			deleteCert = true
+		}
+
+		if deleteCert {
+			certSecretCopy := certSecret.DeepCopy()
+			delete(certSecretCopy.Data, "tls.crt")
+			_, err := c.secretClient.Update(ctx, certSecretCopy, metav1.UpdateOptions{})
+			return key, nil, false, err
```
p0lyn0mial, line 257: should we resubmit if we deleted the cert?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+
+	return key, cert, false, nil
+}
+
+// resubmitCSR deletes the old CSR API object and creates a new one for the given key
+func (c *webhookAuthenticatorCertRequester) resubmitCSR(ctx context.Context, key *rsa.PrivateKey) error {
+	err := c.csrClient.Delete(ctx, csrName, metav1.DeleteOptions{})
```
p0lyn0mial, line 266: you sure CSR are not auto removed ? (https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#request-signing-process)

p0lyn0mial: in that case, we would get `NotFound` right?

pkg/controllers/webhookcerts/cert_requester_test.go
```diff
@@ -0,0 +1,322 @@
...
+	csr, err := x509.ParseCertificateRequest(csrPEM.Bytes)
+	if err != nil {
+		t.Fatalf("failed to parse CSR: %v", err)
+	}
+
+	return csr
+}
+
+func defaultCSRWithStatus(t *testing.T, csrStatus *certapiv1.CertificateSigningRequestStatus) *certapiv1.CertificateSigningRequest {
+	key, err := rsa.GenerateKey(rand.Reader, 4096)
```
p0lyn0mial, line 262: rand is a global var and doesn't need initialization,  right?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+			klog.Error("failed to decode the private key used for the openshift-authenticator, new key will be generated: %v", err)
+			// delete the secret and regenerate it in the next step with a new key
+			err = c.secretClient.Delete(ctx, certSecretName, metav1.DeleteOptions{})
+			if err != nil {
+				return nil, nil, false, err
+			}
+		}
+	}
+	if key == nil {
+		key, err = rsa.GenerateKey(rand.Reader, 4096)
```
p0lyn0mial, line 214: rand is a global var and doesn't need initialization, right?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+		&certapiv1.CertificateSigningRequest{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: csrName,
+			},
+			Spec: certapiv1.CertificateSigningRequestSpec{
+				Request:    csrPEM,
+				SignerName: certapiv1.KubeAPIServerClientSignerName,
+				Usages: []certapiv1.KeyUsage{
+					certapiv1.UsageDigitalSignature,
+					certapiv1.UsageKeyEncipherment,
```
p0lyn0mial, line 296: why do we need this flag?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+	if err := pem.Encode(&buffer,
+		&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csrDER},
+	); err != nil {
+		return nil, fmt.Errorf("failed to convert DER CSR into a PEM-formatted block: %w", err)
+	}
+
+	return buffer.Bytes(), nil
+}
+
+func pemEncodeRSAPrivate(key *rsa.PrivateKey) ([]byte, error) {
```
p0lyn0mial, line 330: we don't have any convenience functions in `library-go` for crypto?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	if cert != nil {
+		// check that the certificate is not expiring
+		certValidityDuration := cert.NotAfter.Sub(cert.NotBefore)
+		if time.Now().Before(cert.NotBefore.Add(time.Duration(0.8 * float64(certValidityDuration)))) {
+			return pbool(false), nil
+		}
+
+		klog.V(5).Info("the authenticator's certificate is almost expired, will schedule a new CSR creation")
```
p0lyn0mial, line 130: maybe this could be an event?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	if csrObj != nil {
+		if len(csrObj.Status.Conditions) == 0 {
+			return pbool(true), nil
+		}
+
+		for _, cond := range csrObj.Status.Conditions {
+			switch cond.Type {
+			case certapiv1.CertificateDenied, certapiv1.CertificateFailed:
```
p0lyn0mial, line 152: wondering what does `CertificateFailed` mean in this context?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+					return pbool(true), nil
+				}
+			}
+		}
+
+		for _, cond := range csrObj.Status.Conditions {
+			if cond.Type == certapiv1.CertificateApproved && cond.Status == corev1.ConditionTrue {
+				csrCert := csrObj.Status.Certificate
+				if len(csrCert) == 0 {
+					return pbool(true), nil
```
p0lyn0mial, line 168: would be nice if we could store progress in the status, like `csr got approved, waiting for cert`

pkg/controllers/webhookcerts/cert_requester.go
stlaz: `Update` won't work if the secret does not exist, will it?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+		// check that the certificate is not expiring
+		certValidityDuration := cert.NotAfter.Sub(cert.NotBefore)
+		if time.Now().Before(cert.NotBefore.Add(time.Duration(0.8 * float64(certValidityDuration)))) {
+			return pbool(false), nil
+		}
+
+		klog.V(5).Info("the authenticator's certificate is almost expired, will schedule a new CSR creation")
+
+		// delete the cert/key secret and schedule a new sync
+		if err := c.secretClient.Delete(ctx, certSecretName, metav1.DeleteOptions{}); err != nil {
```
p0lyn0mial, line 133: simple update would do as well

pkg/controllers/webhookcerts/cert_requester.go
stlaz: the cache should be synced before `sync`, no?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+				if err != nil {
+					return nil, err
+				}
+				certSecret.Data["tls.crt"] = csrCert
+				_, err = c.secretClient.Update(ctx, certSecret, metav1.UpdateOptions{})
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
```
p0lyn0mial, line 181: return at the end otherwise you will resubmit (line 185)

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,380 @@
...
+
+		for _, cond := range csrObj.Status.Conditions {
+			if cond.Type == certapiv1.CertificateApproved && cond.Status == corev1.ConditionTrue {
+				csrCert := csrObj.Status.Certificate
+				if len(csrCert) == 0 {
+					return pbool(true), nil
+				}
+				// do a live Get, the cache might be stale at this point if we did an Update removing the cert before
```
p0lyn0mial, line 170: maybe we should `Update` instead `Delete, Create`

pkg/controllers/webhookcerts/cert_requester.go
stlaz: If the certificate already exists in our CSR, it's going to replace this one, otherwise we either wait for the current CSR to get a cert, or if there is no CSR, we submit a new one.

pkg/controllers/webhookcerts/cert_requester.go
stlaz: exactly, we may not want to always wait 1 hour for a Denied/Failed CSR to disappear

pkg/controllers/webhookcerts/cert_requester.go
stlaz: These are quite common for client cert auth, but may not be all that necessary. I can remove the usage if you want.

pkg/controllers/webhookcerts/cert_requester.go
stlaz: none that are exposed for private key PEM-encoding, none exist for CSRs and RSA key parsing.

pkg/controllers/webhookcerts/cert_requester.go
stlaz: `rand` is just a package in fact, `Reader` reads from system's crypto-safe random resource

pkg/controllers/webhookcerts/cert_requester.go
stlaz: Not exactly sure, but it's better to treat it in case the signer fails for a rather random reason

pkg/controllers/webhookcerts/cert_requester.go
stlaz: I did not want to overcomplicate the condition handling, because I expect we would see a very very fast (about 1s each state) succession of:
1. I submitted the CSR
2. The CSR is pending
3. The CSR got approved, waiting for cert
4. AsExpected

pkg/controllers/webhookcerts/cert_requester.go
stlaz: the update must delete all the keys in the secret's `data`, deleting seems simpler

pkg/controllers/webhookcerts/cert_requester.go
stlaz: good catch :+1:

pkg/controllers/webhookcerts/cert_requester.go
stlaz: I thought I removed that comment. As you can see, the code's using a lister already, the original logic was a bit dirtier

p0lyn0mial (commented):
wondering if we could slightly change `webhookAuthenticatorCertRequester` to the following:

```
1. read the secret
     #  case 1.1 - if the secret is sealed, check expiry of the cert .
     #  case 1.2 - if expired, delete the CSR (if exists) and the secret (transition to case 2.1) .
    #. case 1.3 - otherwise just finish - since the cert was sealed no one has tampered with it .
     
2. the secret is not sealed  or doesn't exists
  # case 2.1 - we don’t have a key (don’t check if the key is valid, just read bytes) and the secret we need one and we need to submit a CSR 
  # case 2.2- the key exists and we need to wait for the CSR, if there is no CSR - create one that means we failed in 2.1 
  # case 2.3 - the CSR observed  
  - check if it matches the private key, if it doesn’t or the private key is broken or the CSR failed then                                         
  - remove the CSR                                         
  - remove the secret - we are back to case `2.1 `                                     
  - otherwise (all good)                                         
  - put the cert in the secret                                         
  - seal it and update
```

what do you think, would it make a bit simpler?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,386 @@
...
+			if cond.Type == certapiv1.CertificateApproved && cond.Status == corev1.ConditionTrue {
+				csrCert := csrObj.Status.Certificate
+				if len(csrCert) == 0 {
+					return pbool(true), nil
+				}
+				certSecret, err := c.secretLister.Secrets("openshift-oauth-apiserver").Get(certSecretName)
+				if err != nil {
+					return nil, err
+				}
+				certSecret.Data["tls.crt"] = csrCert
```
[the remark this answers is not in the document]
stlaz: how do you make secrets immutable with what we've got today?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,386 @@
...
+		WithSync(c.syncWithConditions).
+		WithSyncDegradedOnError(operatorClient).
+		WithInformers(operatorClient.Informer()).
+		WithFilteredEventsInformers(nameFilter(csrName), csrInformers.Informer()).
+		WithFilteredEventsInformers(nameFilter(certSecretName), secretInformers.Informer()).
+		ToController("WebhookAuthenticatorCertRequester", eventsRecorder.WithComponentSuffix("webhook-authenticator-cert-requester"))
+}
+
+func (c *webhookAuthenticatorCertRequester) syncWithConditions(ctx context.Context, syncCtx factory.SyncContext) error {
+	progressing, err := c.sync(ctx, syncCtx)
```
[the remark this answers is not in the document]
stlaz: isn't finished just `progressing == false`?

progressing is a pointer so we've got 3 states:
1. true = either stay progressing or go progressing
2. false = you're no longer progressing (== finished)
3. nil = keep whichever state you're in

pkg/controllers/webhookcerts/cert_requester.go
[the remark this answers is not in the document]
p0lyn0mial: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable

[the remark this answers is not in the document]
stlaz: Interesting, I was looking for "sealed secrets" but that brought something else. I did not know about this. Shouldn't we wait until it goes stable before we start using it, though?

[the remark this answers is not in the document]
p0lyn0mial: it went `GA` in 1.21 - https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1412-immutable-secrets-and-configmaps/kep.yaml#L31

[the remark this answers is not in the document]
stlaz: awesome, I'll try to rework the controller then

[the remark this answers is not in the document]
p0lyn0mial: just a suggestion, it is okay if you don't like it :)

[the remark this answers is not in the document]
stlaz: oh no, I like the suggestion. Although, as I started implementing it I noticed, that it may not bring that much value adding the immutable field in here. You still have no way of preventing someone to stopping the controller, replacing the secret, and starting it again, so you still need to perform all the checks.

pkg/controllers/webhookcerts/cert_requester.go
p0lyn0mial: `deleting` means that we would issue at least `2` requests, one for delete and the other one for create
`updating` would mean a single request - we could delete the old data and regen the new key.

pkg/controllers/webhookcerts/cert_requester.go
p0lyn0mial: is this `Delete` necessary? You already delete in `213` (https://github.com/openshift/cluster-authentication-operator/pull/416/files#diff-bbeb19bb0219c14be678e5136642e92a4958f4c6b330fbce402690efe4b5205bR213)

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,398 @@
...
+		key, err = rsa.GenerateKey(rand.Reader, 4096)
+		if err != nil {
+			return nil, nil, false, fmt.Errorf("failed to generate a private key: %w", err)
+		}
+		keyPEM, err := pemEncodeRSAPrivate(key)
+		if err != nil {
+			return nil, nil, false, err
+		}
+
+		if err := c.secretClient.Delete(ctx, certSecretName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
p0lyn0mial, line 241: is this `Delete` necessary? 

`key == nil` means that there was not secret.

stlaz: ...or the secret was there but had `tls.key` empty

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,398 @@
...
+		for _, cond := range csrObj.Status.Conditions {
+			if cond.Type == certapiv1.CertificateApproved && cond.Status == corev1.ConditionTrue {
+				csrCert := csrObj.Status.Certificate
+				if len(csrCert) == 0 {
+					return pbool(true), nil
+				}
+
+				certObj, err := pemDecodeCert(csrCert)
+				if err != nil {
+					return nil, fmt.Errorf("failed to parse the certiticate retrieved from CSR: %w", err)
```
p0lyn0mial, line 172: this is not fatal?

pkg/controllers/webhookcerts/csr_approver.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	if approved, denied := getCertApprovalCondition(&csr.Status); approved || denied {
+		return nil
+	}
+
+	csrCopy := csr.DeepCopy()
+	csrPEM, _ := pem.Decode(csrCopy.Spec.Request)
+	if csrPEM == nil {
+		return c.denyCSR(ctx, csrCopy, "NoCSRFound", "failed to PEM-parse the CSR block in .spec.request: no CSRs were found")
```
p0lyn0mial, line 81: should we set `CertificateFailed` instead?

pkg/controllers/webhookcerts/csr_approver.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	csrObj, err := x509.ParseCertificateRequest(csrPEM.Bytes)
+	if err != nil {
+		return c.denyCSR(ctx, csrCopy, "CSRParsingFailed", fmt.Sprintf("failed to parse the CSR bytes: %v", err))
```
p0lyn0mial, line 86: `CertificateFailed` ?

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -116,49 +127,27 @@ func (c *webhookAuthenticatorController) sync(ctx context.Context, syncCtx facto
...
-		return nil, fmt.Errorf("service-account token secret is missing the 'token' key")
+		return nil, fmt.Errorf("failed to read service-ca crt bundle: %w", err)
 	}
 
 	kubeconfigBytes, err := assets.Asset("oauth-apiserver/authenticator-kubeconfig.yaml")
 	if err != nil {
 		return nil, fmt.Errorf("failed to read kubeconfig template: %w", err)
 	}
 
+	// TODO: get service-ca CA bundle from somewhere
```
p0lyn0mial, line 145: so, what's inside `caBundle` ?

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +165,56 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
 	secret, _, err := resourceapply.ApplySecret(c.secrets, recorder, requiredSecret)
 	return secret, err
 }
+
+func (c *webhookAuthenticatorController) getAuthenticatorCertKeyPair(ctx context.Context) (key, cert []byte, err error) {
```
p0lyn0mial, line 169: `ctx` not used

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +165,56 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
...
+			}
+
+			if len(*waitingForCertKey) > 0 {
+				cond.Status = operatorv1.ConditionTrue
+				cond.Reason = "WaitingForCertKey"
+				cond.Message = *waitingForCertKey
+			}
+
+			if _, _, statusErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(cond)); statusErr != nil {
+				klog.Error("failed to update operator status: %v", statusErr)
```
p0lyn0mial, line 187: we don't want to return `statusErr` instead?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,422 @@
...
+		}
+	}
+
+	if key != nil {
+		return key, false, false, err
+	}
+
+	// either the key has always been nil, or the certificate is expiring
+	key, err = createSecretWithNewKey(ctx, c.secretClient)
+	return key, false, false, err
```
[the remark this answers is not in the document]
stlaz: yes, I accidentally removed it from here, but this is the spot

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,422 @@
...
+	return pbool(false), nil
+}
+
+func (c *webhookAuthenticatorCertRequester) checkKeyCertSecret(ctx context.Context, recorder events.Recorder) (key *rsa.PrivateKey, secretComplete, shouldResubmit bool, err error) {
+	key, cert, err := c.getKeyCertFromSecret(ctx)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, false, false, err
+	}
+
+	if key == nil && !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
stlaz: exactly, this is where you get when you don't have the key but the secret actually exists

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,422 @@
...
+
+	// either we did not have a key or the CSR failed
+	if shouldResubmit {
+		if err := c.resubmitCSR(ctx, key); err != nil {
+			return nil, err
+		}
+		return pbool(true), nil
+	}
+
+	return pbool(false), nil
```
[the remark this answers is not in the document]
stlaz: it shouldn't under normal circumstances

pkg/controllers/webhookcerts/csr_approver.go
stlaz: No, the signer sets that

pkg/controllers/webhookcerts/csr_approver.go
stlaz: ditto

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
stlaz: probably an old note :neutral_face:

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
stlaz: good point

pkg/controllers/webhookcerts/csr_approver.go
stlaz: https://github.com/openshift/cluster-authentication-operator/blob/f3ce0832d72f133018b14e6767c59e34c6fb2f35/vendor/k8s.io/api/certificates/v1/types.go#L197-L202

pkg/controllers/webhookcerts/csr_approver.go
p0lyn0mial: we set it :)

```
func (c *csrApproverController) denyCSR(ctx context.Context, csrCopy *certapiv1.CertificateSigningRequest, reason, message string) error {
	csrCopy.Status.Conditions = append(csrCopy.Status.Conditions,
		certapiv1.CertificateSigningRequestCondition{
			Type:    certapiv1.CertificateDenied,
			Status:  corev1.ConditionTrue,
			Reason:  reason,
			Message: message,
		},
	)

	_, err := c.csrClient.UpdateApproval(ctx, csrName, csrCopy, v1.UpdateOptions{})
	return err
}
```

p0lyn0mial: we could set `Type` to `CertificateFailed`

p0lyn0mial: this is nice to have, we don't have to change it.

pkg/operator/starter.go
```diff
@@ -531,9 +531,6 @@ func prepareOauthAPIServerOperator(ctx context.Context, controllerContext *contr
 		[]string{
 			"oauth-apiserver/ns.yaml",
 			"oauth-apiserver/apiserver-clusterrolebinding.yaml",
-			"oauth-apiserver/authenticator-sa.yaml",
-			"oauth-apiserver/authenticator-clusterrole.yaml",
-			"oauth-apiserver/authenticator-clusterrolebinding.yaml",
```
sttts, line 532: are they cleaned up anywhere?

stlaz: they were never there in 4.7 so there's no need for a cleanup

pkg/controllers/webhookcerts/csr_approver.go
p0lyn0mial: aha, now I understand, there is a limitation, sorry :)

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,422 @@
...
+
+func (c *webhookAuthenticatorCertRequester) checkKeyCertSecret(ctx context.Context, recorder events.Recorder) (key *rsa.PrivateKey, secretComplete, shouldResubmit bool, err error) {
+	key, cert, err := c.getKeyCertFromSecret()
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, false, false, err
+	}
+
+	if key == nil && !apierrors.IsNotFound(err) {
+		// remove the current secret to replace it with a new one
+		if err = c.secretClient.Delete(ctx, certSecretName, metav1.DeleteOptions{}); err != nil {
```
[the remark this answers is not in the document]
stlaz: good idea, I completely forgot about it

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +164,57 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
 	secret, _, err := resourceapply.ApplySecret(c.secrets, recorder, requiredSecret)
 	return secret, err
 }
+
+func (c *webhookAuthenticatorController) getAuthenticatorCertKeyPair() (key, cert []byte, err error) {
+	var waitingForCertKey *string
```
sttts, line 169: this is not a CertKey. This is a message? Use a better variable name.

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +164,57 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
...
+	defer func() {
+		if waitingForCertKey != nil {
+			cond := operatorv1.OperatorCondition{
+				Type:    "AuthenticatorCertKeyProgressing",
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "AsExpected",
+				Message: "All is well",
+			}
+
+			if len(*waitingForCertKey) > 0 {
```
sttts, line 179: this three-state logic is hard to read

pkg/controllers/webhookcerts/cert_requester.go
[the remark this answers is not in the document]
sttts: I don't think it will:
- at line 118 shouldResubmit==true => line 140 or 142  will return.
- at line 118 shouldResubmit==true => shouldResubmit is overwritten in line 120. If it is false now, line 132 return. If it is true now, line 140 or 142 return.

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,412 @@
...
+		// check that the certificate is not expiring
+		certValidityDuration := cert.NotAfter.Sub(cert.NotBefore)
+		if time.Now().Before(cert.NotBefore.Add(time.Duration(0.8 * float64(certValidityDuration)))) {
+			return key, true, false, nil
+		}
+		// set key == nil to trigger new key generation
+		key = nil
+
+		expiryMsg := fmt.Sprintf("the authenticator's certificate is almost expired (%s), will schedule a new CSR creation", cert.NotAfter.String())
+		recorder.Eventf("ClientCertExpiringSoon", expiryMsg)
```
sttts, line 164: this line has potential to keep creating events continuously. Only create events on an edge, not based on state like here. What is the edge?

sttts: Would move it after a successful createSecretWithNewKey.

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,412 @@
...
+		return key, false, false, err
+	}
+
+	// either the key has always been nil, or the certificate is expiring
+	key, err = createSecretWithNewKey(c.secretClient, recorder)
+	return key, false, true, err
+
+}
+
+func (c *webhookAuthenticatorCertRequester) handleCSR(key *rsa.PrivateKey, recorder events.Recorder) (cert []byte, shouldResubmit bool, err error) {
```
sttts, line 178: what does "handle" mean? Create a cert for the key?

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,412 @@
...
+			if cond.Status == corev1.ConditionTrue {
+				recorder.Eventf("CSRFailure", "previous CSR got denied or it otherwise failed: {%s - %s: %s}", cond.Type, cond.Reason, cond.Message)
+				klog.Warning("the previous CSR got denied or it otherwise failed: %v", cond)
+				return nil, true, nil
+			}
+		}
+	}
+
+	for _, cond := range csrObj.Status.Conditions {
+		if cond.Type == certapiv1.CertificateApproved && cond.Status == corev1.ConditionTrue {
```
sttts, line 203: nit: avoid deep indentions by continuouing early.

pkg/controllers/webhookcerts/cert_requester.go
```diff
@@ -0,0 +1,412 @@
...
+				return nil, false, nil
+			}
+
+			certObj, err := pemDecodeCert(csrCert)
+			if err != nil {
+				return nil, false, fmt.Errorf("failed to parse the certiticate retrieved from CSR: %w", err)
+			}
+
+			if !key.PublicKey.Equal(certObj.PublicKey) {
+				recorder.Eventf("CSRBadCertKey", "the public key of the CSR certificate did not match the current private key")
```
sttts, line 215: events based on state are bad

pkg/controllers/webhookcerts/cert_requester.go
stlaz: The certificate gets removed in the next step.

I arbitrarily picked 80% of the lifetime of the certificate to be the edge. I don't think there's a chance we get here at the very moment when the the certificate reaches that lifetime edge.

pkg/controllers/webhookcerts/cert_requester.go
stlaz: if I remove it, we're likely to never find out why the rotation was triggered :|

pkg/controllers/webhookcerts/cert_requester.go
stlaz: process CSR's conditions, grab the certificate if it exists

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
stlaz: I added an explanation to the variable, should make this easy to understand

pkg/controllers/webhookcerts/cert_requester.go
stlaz: I fixed this by adding this event call in defer only if `err == nil`

pkg/controllers/webhookcerts/cert_requester.go
sttts: there is always a delay, also later in the runtime

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +164,61 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
...
 }
+
+func (c *webhookAuthenticatorController) getAuthenticatorCertKeyPair() (key, cert []byte, err error) {
+	// waitingForCertKeyMsg lets us decide whether we should set progressing condition to true
+	// 1. nil: keep "progressing the same"
+	// 2. non-empty string: progress with the message in the string
+	// 3. empty string: stop progressing
+	var waitingForCertKeyMsg *string
+	defer func() {
+		if waitingForCertKeyMsg != nil {
```
mfojtik, line 175: nit:

```
if waitingForCertKeyMsg == nil {
  return
}
```

will make this 1 nesting level better :-)

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +164,61 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
...
+
+			if len(*waitingForCertKeyMsg) > 0 {
+				cond.Status = operatorv1.ConditionTrue
+				cond.Reason = "WaitingForCertKey"
+				cond.Message = *waitingForCertKeyMsg
+			}
+
+			if _, _, statusErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(cond)); statusErr != nil {
+				klog.Errorf("failed to update operator status: %v", statusErr)
+				err = statusErr
```
mfojtik, line 191: i hate named return values :-)

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +164,61 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
...
+	// 2. non-empty string: progress with the message in the string
+	// 3. empty string: stop progressing
+	var waitingForCertKeyMsg *string
+	defer func() {
+		if waitingForCertKeyMsg != nil {
+			cond := operatorv1.OperatorCondition{
+				Type:    "AuthenticatorCertKeyProgressing",
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "AsExpected",
+				Message: "All is well",
```
mfojtik, line 180: please use the same message as other operators for the sake of consistency

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -176,3 +164,61 @@ func (c *webhookAuthenticatorController) ensureKubeConfigSecret(ctx context.Cont
...
+	}
+
+	cert, ok = certSecret.Data["tls.crt"]
+	if !ok {
+		waitingForCertKeyMsg = pstr("the authenticator's client cert secret is missing the 'tls.crt' key")
+		return nil, nil, nil
+	}
+
+	// stop progressing on the cert/key secret
+	waitingForCertKeyMsg = pstr("")
```
mfojtik, line 218: where is this used?

pkg/controllers/webhookcerts/csr_approver.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+		return err
+	}
+
+	if csr.Spec.Username != "system:serviceaccount:openshift-authentication-operator:authentication-operator" {
+		return fmt.Errorf("CSR %q was created by an unexpected user: %q", csrName, csr.Spec.Username)
+	}
+
+	if csrGroups := sets.NewString(csr.Spec.Groups...); !csrGroups.IsSuperset(saGroupSet) {
+		return fmt.Errorf("CSR %q was created by a user with unexpected groups: %v", csrName, csrGroups.List())
```
sttts, line 71: what is the result of these? We go degraded? Or just silence?

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
stlaz: I'm afraid there's no other way to propagate the status err to the return value :-|

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
stlaz: in the defer

pkg/controllers/webhookcerts/csr_approver.go
stlaz: the `WithSyncDegradedOnError(operatorClient).` line in the creation of this controller makes sure errors in the sync cause degraded state

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
stlaz: This is actually it.

source:
https://github.com/openshift/cluster-authentication-operator/blob/f3ce0832d72f133018b14e6767c59e34c6fb2f35/vendor/github.com/openshift/library-go/pkg/operator/status/condition.go#L68-L71

PR #417, opened 2021-03-04
Bug 1934400: apiserver: specifically allow privilege escalation

stlaz:
An SCC that allows privileged containers but defaults allowing
privilege escalation in containers might nuke the apiservers on restart.

---
Apparently IBM started distributing such an SCC with `priority:1` in their Cloud Paks multi-cluster-management offerings. Such an SCC is very likely to break any container that sets `privileged: true` but does not set `allowPrivilegeEscalation: true` (as one would expect that to be implied by the system).


sttts:
/close

Replaced with https://github.com/openshift/apiserver-library-go/pull/47

PR #422, opened 2021-03-24
oauth-apiserver: add access token inactivity timeout observer

stlaz:
/assign @sttts 
cc @ingvagabund


stlaz:
The operator tests failed because the default test timeout was 10m but the token timeout tests are tremendously time consuming.

I increased the token timeout period and attempted to decrease the time per the test by roughly 15minutes.

stlaz:
/retest
infra

stlaz:
/retest
install failed on e2e-operator

stlaz:
/retest
no gcp resources

test/e2e/tokentimeout_test.go
```diff
@@ -95,7 +95,7 @@ func TestTokenInactivityTimeout(t *testing.T) {
 		testTokenValidity(t, req, http.StatusOK, tokenWithoutTimeout.Name, "token with out timeout should work")
 	}
 
-	configInactivityTimeout := int32(600)
+	configInactivityTimeout := int32(420)
```
sttts, line 99: why still that long? Why not `O(5 minutes)` ?

test/e2e/tokentimeout_test.go
```diff
@@ -299,3 +305,19 @@ func createOAuthClient(t *testing.T, oauthClientClient *oauthclient.OauthV1Clien
...
+	t.Logf("waiting all oauth-apiserver replicas to be updated and ready")
+
+	err := wait.PollImmediate(time.Second, 10*time.Minute, func() (bool, error) {
+		deployment, err := kubeClient.AppsV1().Deployments("openshift-oauth-apiserver").Get(context.Background(), "apiserver", metav1.GetOptions{})
+		if err != nil {
+			t.Logf("failed to retrieve oauth-apiserver's deployment: %v", err)
+			return false, nil
+		}
+
+		return deployment.Status.Replicas == deployment.Status.ReadyReplicas && deployment.Status.UpdatedReplicas == deployment.Status.Replicas, nil
```
sttts, line 319: not spec.replicas?

test/e2e/tokentimeout_test.go
sttts: Or iow: what's the reason for these two numbers?

stlaz: 300s is the minimal, I think the original test was trying to prove that one of the timeouts apply and it's not the other one, so it had different values. I thought 2 minutes should be a sufficient-enough difference.

PR #424, opened 2021-03-29
oauth-apiserver: don't attempt to write to system trust store

stlaz:
We don't use the global trusted CA bundle in this API server, don't attempt to write it.

/assign @p0lyn0mial

### DIFF
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -329,26 +329,19 @@ spec:
         terminationMessagePolicy: FallbackToLogsOnError
         image: ${IMAGE}
         imagePullPolicy: IfNotPresent
-        command: ["/bin/bash", "-ec"]
+        command: ["oauth-apiserver", "start"]
         args:
-          - |
-            if [ -s /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem ]; then
-              echo "Copying system trust bundle"
-              cp -f /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
-            fi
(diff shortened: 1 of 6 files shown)


pkg/operator/assets/bindata.go
```diff
@@ -329,26 +329,19 @@ spec:
         terminationMessagePolicy: FallbackToLogsOnError
         image: ${IMAGE}
         imagePullPolicy: IfNotPresent
-        command: ["/bin/bash", "-ec"]
+        command: ["oauth-apiserver", "start"]
         args:
-          - |
-            if [ -s /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem ]; then
```
p0lyn0mial, line 335: it was taken from `oas` - https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/bindata/v3.11.0/openshift-apiserver/deploy.yaml#L67

should we update there as well?

p0lyn0mial:
> We don't use the global trusted CA bundle in this API server, don't attempt to write it.

we use the CA from the config map, right?

p0lyn0mial:
if this is what we want then we should remove the wiring like https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator/starter.go#L395

pkg/operator/assets/bindata.go
stlaz: no, I think OAS is actually using this, but we're not

stlaz:
> we use the CA from the config map, right?

no, we're the trust roots vendored inside the image

> if this is what we want then we should remove the wiring like https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator/starter.go#L395

we can't, that's a different workload that's actually using it :)

PR #425, opened 2021-03-29
update library-go to fix tight loops

deads2k:
openshift/library-go#1024 and openshift/library-go#1036


stlaz:
/hold cancel
#414 merged, commits
```
update library-go to fix tight loops
react to bump
```
should no longer be needed

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -89,7 +89,7 @@ func NewWellKnownReadyController(kubeInformers v1helpers.KubeInformersForNamespa
 		nsOpenshiftConfigManagedInformers.Core().V1().ConfigMaps().Informer(),
 		routeInformer.Informer(),
 	).
-		ResyncEvery(30*time.Second).
```
stlaz, line 92: in the worst cases scenario, this change might prolong an installation by 30s

PR #426, opened 2021-03-30
expose pod information to oauth-apiserver container via env variables

p0lyn0mial:
the pod name and the namespace are used by the termination code to properly record events.

before the change the termination events were recorded to the default namespace and weren't associated with any pod.
after the change the termination evens are recorded in the openshift-oauth-apiserver namespace and are bound to oauth-apiserver pods.

### DIFF
--- a/bindata/oauth-apiserver/deploy.yaml
+++ b/bindata/oauth-apiserver/deploy.yaml
@@ -75,6 +75,15 @@ spec:
           requests:
             memory: 200Mi
             cpu: 150m
+        env:
+          - name: POD_NAME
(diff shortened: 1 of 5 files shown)


stlaz:
shall we add an e2e test to prove it works?

p0lyn0mial:
> shall we add an e2e test to prove it works?

mhm, we do have e2e that captures more interesting events, these were purely informative so I would say no, wdyt?

p0lyn0mial:
xref: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/440

PR #430, opened 2021-04-06
Custom Certs for OAuth Route

awgreene:


pkg/controllers/configobservation/customroutersecret/observe_custom_router_secret.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/configobservation"
+)
+
+func NewObserveCustomRouterSecret() func(configobserver.Listers, events.Recorder, map[string]interface{}) (map[string]interface{}, []error) {
```
stlaz, line 16: I would expect this observer to configure `observedConfig.oauthServer.servingInfo.namedCertificates` but I don't see that anywhere. What am I missing?

pkg/controllers/configobservation/customroutersecret/observe_custom_router_secret.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// resourceSyncFunc syncs a resource from the source location to the destination location.
+type resourceSyncFunc func(destination, source resourcesynccontroller.ResourceLocation) error
+
+// syncActionRules rules define source resource names indexed by destination resource names.
+// Empty value means to delete the destination.
+type syncActionRules map[string]string
+
+// syncObservedResources copies or deletes resources, sources in GlobalUserSpecifiedConfigNamespace and destinations in OperatorNamespace namespace.
```
stlaz, line 94: this is the only mention about `GlobalUserSpecifiedConfigNamespace` in this repo

pkg/controllers/configobservation/customroutersecret/observe_custom_router_secret.go
```diff
@@ -0,0 +1,135 @@
...
+func (c *customRouterObserver) componentRouteSecret(lister configobservation.Listers, ingress *configv1.Ingress) ([]interface{}, syncActionRules, error) {
+	componentRoutes := []interface{}{}
+
+	rules := syncActionRules{}
+
+	// make sure the output slice of named certs is sorted by domain so that the generated config is deterministic
+	for _, componentRoute := range ingress.Spec.ComponentRoutes {
+		if componentRoute.Name == c.componentRoute.Name &&
+			componentRoute.Namespace == c.componentRoute.Namespace {
+			if _, err := lister.SecretsLister.Secrets("openshift-config").Get(componentRoute.ServingCertKeyPairSecret.Name); err != nil ||
```
stlaz, line 121: what if `componentRoute.ServingCertKeyPairSecret.Name` is empty? The call will fail, but do we want to test it this way?

pkg/controllers/configobservation/customroutersecret/observe_custom_router_secret.go
```diff
@@ -0,0 +1,135 @@
...
+				componentRoute.ServingCertKeyPairSecret.Name == "" {
+				rules[c.destSecret] = ""
+			} else {
+				componentRoutes = append(componentRoutes, map[string]interface{}{
+					"secret":   interface{}(componentRoute.ServingCertKeyPairSecret.Name),
+					"hostname": interface{}(string(componentRoute.Hostname)),
+				})
```
stlaz, line 128: I am not sure what these fields are for :-|

pkg/controllers/configobservation/customroutersecret/observe_custom_router_secret.go
```diff
@@ -0,0 +1,135 @@
...
+		// if 'from' is empty, then it means we want to delete
+		destination := resourcesynccontroller.ResourceLocation{Namespace: "openshift-authentication", Name: to}
+		if err := syncResource(destination, source); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errs
+}
+
+func (c *customRouterObserver) componentRouteSecret(lister configobservation.Listers, ingress *configv1.Ingress) ([]interface{}, syncActionRules, error) {
```
stlaz, line 112: you never appear to use error in this function

pkg/controllers/configobservation/routersecret/observe_router_secret.go
```diff
@@ -53,16 +69,15 @@ func ObserveRouterSecret(genericlisters configobserver.Listers, recorder events.
...
+func routerSecretToSNI(routerSecret *corev1.Secret, certFile string, keyFile string) ([]interface{}, error) {
 	certs := []interface{}{}
 	// make sure the output slice of named certs is sorted by domain so that the generated config is deterministic
 	for _, domain := range sets.StringKeySet(routerSecret.Data).List() {
 		certs = append(certs, map[string]interface{}{
 			"names":    []interface{}{"*." + domain}, // ingress domain is always a wildcard
-			"certFile": interface{}("/var/config/system/secrets/v4-0-config-system-router-certs/" + domain),
-			"keyFile":  interface{}("/var/config/system/secrets/v4-0-config-system-router-certs/" + domain),
+			"certFile": interface{}(certFile + domain),
+			"keyFile":  interface{}(keyFile + domain),
```
stlaz, line 79: the docs to `ServingCertKeyPairSecret` say `// ServingCertKeyPairSecret is a reference to a secret in namespace/openshift-config that is a kubernetes tls secret.` so this is most probably not what the cert/key files are going to look like. Instead, they are likely to be `tls.crt` and `tls.key`

pkg/controllers/deployment/default_deployment.go
```diff
@@ -94,9 +95,45 @@ func getOAuthServerDeployment(operatorConfig *operatorv1.Authentication, proxyCo
...
+					SecretName: "v4-0-config-system-custom-router-certs",
+				},
+			},
+		})
+		container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{
+			Name:      "v4-0-config-system-custom-router-certs",
+			ReadOnly:  true,
+			MountPath: "/var/config/system/secrets/v4-0-config-system-custom-router-certs",
+		})
+	}
```
stlaz, line 112: you may want to try to optionally bind it all the time

stlaz (changes requested):
Does the host of the route get changed anywhere?

awgreene:
> Does the host of the route get changed anywhere?
@stlaz not yet.

pkg/controllers/configobservation/customroutersecret/observe_custom_router_secret.go
awgreene: Left over from a copy/paste, I'll rework the comments.

pkg/controllers/configobservation/customroutersecret/observe_custom_router_secret.go
awgreene: I'll remove the error.

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -0,0 +1,86 @@
+package componentroutesecretsync
```
stlaz, line 1: I would expect the observer disappear with this commit. Actually, the observer does not seem necessary with this commit it?

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+	destination := resourcesynccontroller.ResourceLocation{Namespace: c.destSecretNamespace, Name: c.destSecretName}
+	return c.resourceSyncer.SyncSecret(destination, source)
+}
+
+func (c *componentRouteSecretSyncController) componentRouteSecretSource(ingress *configv1.Ingress) resourcesynccontroller.ResourceLocation {
+	// check if expected componentRoute exists
+	for _, componentRoute := range ingress.Spec.ComponentRoutes {
+		if componentRoute.Name == c.componentRoute.Name &&
+			componentRoute.Namespace == c.componentRoute.Namespace {
```
stlaz, line 106: this is a reappearing pattern, might be worth having a function that does it for you

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -0,0 +1,86 @@
...
+type componentRouteSecretSyncController struct {
+	destSecretNamespace string
+	destSecretName      string
+	componentRoute      types.NamespacedName
+	ingressLister       configlistersv1.IngressLister
+	secretLister        corev1listers.SecretLister
+	resourceSyncer      resourcesynccontroller.ResourceSyncer
+}
+
+func (c *componentRouteSecretSyncController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
stlaz, line 73: also add componentRoute status reporting

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -55,18 +62,41 @@ type componentRouteSecretSyncController struct {
...
 	ingress, err := c.ingressLister.Get("cluster")
 	if err != nil {
-		return err
+		return common.UpdateControllerConditions(c.operatorClient, degradedConditionTypes, []operatorv1.OperatorCondition{
+			{
+				Type:    "IngressComponentRouteControllerDegraded",
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "UnableToRetriveClusterIngressConfig",
+				Message: fmt.Sprintf("Unable to get ingress.config.openshift.io/cluster resource: %s", err.Error()),
+			}})
```
stlaz, line 82: this gets handled by the controller from the controller factory, I don't think this is necessary at this point.

stlaz (changes requested):
there now appears to be both the controller and the observer to sync the secret. Keep the controller, add ingress status reporting about the route.

Route hostname change should still happen, right?

_edit_: I did not notice the observer getting remove in the later commit. Squash please.

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
awgreene: This is not an observer, this is a controller that syncs a secret from the cluster-admin provided location to the known secret location.

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"context"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
+	corev1listers "k8s.io/client-go/listers/core/v1"
```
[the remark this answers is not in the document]
awgreene: Can do!

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+	configinformers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	"k8s.io/apimachinery/pkg/types"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
```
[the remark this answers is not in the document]
awgreene: Can do!

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -0,0 +1,86 @@
...
+		componentRoute: types.NamespacedName{
+			Namespace: componentRouteNamespace,
+			Name:      componentRouteName,
+		},
+		ingressLister:  configInformer.Lister(),
+		secretLister:   secretInformer.Lister(),
+		resourceSyncer: resourceSyncer,
+	}
+	return factory.New().
+		WithInformers(
```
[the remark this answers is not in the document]
awgreene: the configInformer is an ingress informer.

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
awgreene: I was thinking of adding that in a second controller, but I can do that here.

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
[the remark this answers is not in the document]
awgreene: Done!

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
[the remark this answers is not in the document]
awgreene: Updated the name of the variable to improve clarity.

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type componentRouteSecretSyncController struct {
+	destSecretNamespace string
+	destSecretName      string
+	componentRoute      types.NamespacedName
+	ingressLister       configlistersv1.IngressLister
+	secretLister        corev1listers.SecretLister
+	resourceSyncer      resourcesynccontroller.ResourceSyncer
+}
```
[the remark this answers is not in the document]
awgreene: Can do!

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
awgreene: Agreed!

awgreene: Moved this to the `pkg/controller/common/ingress_config.go` file.

pkg/controllers/componentroutestatus/ingress_component_route.go
```diff
@@ -0,0 +1,89 @@
+package componentroutestatus
```
awgreene, line 1: @stlaz asked me offline to merge this into component_route_secret_sync_controller.

pkg/controllers/common/ingress_config.go
```diff
@@ -28,3 +28,36 @@ func GetIngressConfig(ingressLister configv1lister.IngressLister, conditionPrefi
 	}
 	return ingress, nil
 }
+
+// getComponentRouteSpec searches the entries of the ingress.spec.componentRoutes array for a componentRoute with a matching namespace and name.
+// If a componentRoute is found it is returned along with a true boolean
+// If a componentRoute is not found an empty component route is return along with a false.
+func getComponentRouteSpec(ingress configv1.Ingress, namespace string, name string) (configv1.ComponentRouteSpec, bool) {
+	for _, componentRoute := range ingress.DeepCopy().Spec.ComponentRoutes {
```
EmilyM1, line 36: Looks good Alex.
I always go with the short form on parameters ex: `(ingress configv1.Ingress, namespace, name string)` but this is mostly a style preference

pkg/controllers/configobservation/routersecret/observe_router_secret.go
```diff
@@ -53,16 +51,49 @@ func ObserveRouterSecret(genericlisters configobserver.Listers, recorder events.
...
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	}
+
+	if err == nil {
+		ingress, err := listers.IngressLister.Get("cluster")
+		if err != nil {
+			return nil, err
+		}
+		return []interface{}{namedCertificate(common.GetHostname(*ingress, "openshift-auth", "oauth"), "/var/config/system/secrets/v4-0-config-system-custom-router-certs/tls.key", "/var/config/system/secrets/v4-0-config-system-custom-router-certs/tls.crt")}, nil
```
stlaz, line 66: nit: ugly, split to multiple lines

stlaz (changes requested):
- [ ] merge the status and secret sync controllers
- [ ] the componentRoutes in status should have `Conditions` describing their current state (see the enhancement)
- [ ] the componentRoutes in status should have `consumingUsers` set
- [ ] don't pass variables by value unless you've got a good reason to do so, your stack might start hating you for that

stlaz:
I just realized: should we use the "CustomRoute" instead of "ComponentRoute" terminology in the controller name(s)? IMO it makes more sense to call it that

pkg/controllers/common/ingress_config.go
```diff
@@ -28,3 +28,36 @@ func GetIngressConfig(ingressLister configv1lister.IngressLister, conditionPrefi
 	}
 	return ingress, nil
 }
+
+// getComponentRouteSpec searches the entries of the ingress.spec.componentRoutes array for a componentRoute with a matching namespace and name.
+// If a componentRoute is found it is returned along with a true boolean
+// If a componentRoute is not found an empty component route is return along with a false.
+func getComponentRouteSpec(ingress configv1.Ingress, namespace string, name string) (configv1.ComponentRouteSpec, bool) {
```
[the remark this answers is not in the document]
awgreene: Great idea, thanks!

pkg/controllers/componentroutestatus/ingress_component_route.go
```diff
@@ -0,0 +1,89 @@
...
+			componentRoute.Name != c.componentRoute.Name {
+			continue
+		}
+		location = i
+	}
+	componentRoute := configv1.ComponentRouteStatus{
+		Namespace:        c.componentRoute.Namespace,
+		Name:             c.componentRoute.Name,
+		DefaultHostname:  configv1.Hostname(common.GetHostname(*ingress, "openshift-authentication", "oauth")),
+		CurrentHostnames: []configv1.Hostname{"domain"},
```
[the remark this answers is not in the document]
awgreene: Sorry - hardcoded this for test purposes. Will place the actual hostname here.

pkg/controllers/configobservation/routersecret/observe_router_secret.go
```diff
@@ -63,6 +78,18 @@ func routerSecretToSNI(routerSecret *corev1.Secret) ([]interface{}, error) {
 			"keyFile":  interface{}("/var/config/system/secrets/v4-0-config-system-router-certs/" + domain),
 		})
 	}
+	return certs, nil
+}
 
+func routerCustomSecretToSNI(routerSecret *corev1.Secret) ([]interface{}, error) {
+	certs := []interface{}{}
+	// make sure the output slice of named certs is sorted by domain so that the generated config is deterministic
+	for _, domain := range sets.StringKeySet(routerSecret.Data).List() {
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest version of the commit.

awgreene:
> I just realized: should we use the "CustomRoute" instead of "ComponentRoute" terminology in the controller name(s)? IMO it makes more sense to call it that

Done

pkg/controllers/componentroutestatus/ingress_component_route.go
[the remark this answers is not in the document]
awgreene: Addressed in the latest version of the pr.

pkg/controllers/common/ingress_config.go
```diff
@@ -28,3 +28,36 @@ func GetIngressConfig(ingressLister configv1lister.IngressLister, conditionPrefi
...
+		}
+	}
+	return configv1.ComponentRouteSpec{}, false
+}
+
+// GetServingCertKeyPairSecret searches the entries of the ingress.spec.componentRoutes array for a componentRoute with a matching namespace and name
+// and returns the servingCertKeyPairSecret.Name  defined in the entry if found, returns an empty string otherwise.
+func GetServingCertKeyPairSecret(ingress configv1.Ingress, namespace string, name string) string {
+	componentRoute, found := getComponentRouteSpec(ingress, namespace, name)
+	if found {
```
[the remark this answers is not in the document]
awgreene: Done.

awgreene:
>  the componentRoutes in status should have consumingUsers set

@stlaz I do not believe that we need to set this value. Doing so would cause the ingress-controller to generate RBAC for the provided users. This controller already has the rbac necessary to read the admin provided secrets.

pkg/controllers/componentroutestatus/ingress_component_route.go
awgreene: Done.

pkg/controllers/componentroutesecretsync/component_route_secret_sync_controller.go
awgreene: Actually done this time.

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -242,8 +252,8 @@ func getOauthRoute(ingressConfig *configv1.Ingress) *routev1.Route {
 			OwnerReferences: nil, // TODO
 		},
 		Spec: routev1.RouteSpec{
-			Host:      "oauth-openshift." + ingressConfig.Spec.Domain, // mimic the behavior of subdomain
-			Subdomain: "",                                             // TODO once subdomain is functional, remove reliance on ingress config and just set subdomain=targetName
+			Host:      "oauth-openshift." + domain, // mimic the behavior of subdomain
```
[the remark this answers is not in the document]
awgreene: Agreed, should be all set now.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -224,7 +224,8 @@ func getOAuthRouteTLSConfig(cmLister corev1listers.ConfigMapLister, secretLister
 	}
 
 	return &tls.Config{
-		RootCAs: rootCAs,
+		RootCAs:            rootCAs,
+		InsecureSkipVerify: true, // TODO: REMOVE THIS! Don't merge this to OpenShift, hack to support self signed certs.
```
awgreene, line 228: This must be removed before the PR is merged.
/hold

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,132 @@
...
+	ingressConfig, err := c.ingressLister.Get("cluster")
+	if err != nil {
+		return err
+	}
+
+	// Update ingressConfig
+	route, err := c.routeLister.Routes("openshift-authentication").Get("oauth-openshift")
+	if err != nil {
+		return err
+	}
```
stlaz, line 87: `route.Spec.Host` should most likely be set in this controller.

That would mean moving `metadataController.handleRoute` here, and the route object should be moved to bindata.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,132 @@
...
+	// Copy source secret to destination
+	destination := resourcesynccontroller.ResourceLocation{Namespace: c.destSecret.Namespace, Name: c.destSecret.Name}
+	source := c.componentRouteSecretSource(ingressConfigCopy)
+
+	return c.resourceSyncer.SyncSecret(destination, source)
+}
+
+func (c *customRouteController) componentRouteSecretSource(ingress *configv1.Ingress) resourcesynccontroller.ResourceLocation {
+	// check if expected componentRoute exists
+	if secretName := common.GetCustomRouteServingCertKeySecret(ingress, c.componentRoute.Namespace, c.componentRoute.Name); secretName != "" {
```
stlaz, line 101: no validating the secret?

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,132 @@
...
+		if _, err := c.secretLister.Secrets("openshift-config").Get(secretName); err == nil {
+			return resourcesynccontroller.ResourceLocation{Namespace: "openshift-config", Name: secretName}
+		}
+	}
+	return resourcesynccontroller.ResourceLocation{}
+}
+
+func (c *customRouteController) updateIngressConfigStatus(ctx context.Context, currentHostname configv1.Hostname, ingress *configv1.Ingress) error {
+	desiredComponentRoute := configv1.ComponentRouteStatus{
+		Namespace:        c.componentRoute.Namespace,
```
stlaz, line 111: needs conditions

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,132 @@
...
+				Resource:  "routes",
+				Name:      "oauth-openshift",
+				Namespace: "openshift-authentication"},
+		},
+	}
+
+	if existingComponentRoute := common.GetComponentRouteStatus(ingress, c.componentRoute.Namespace, c.componentRoute.Name); existingComponentRoute == nil {
+		ingress.Status.ComponentRoutes = append(ingress.Status.ComponentRoutes, desiredComponentRoute)
+	} else {
+		*existingComponentRoute = desiredComponentRoute
```
stlaz, line 127: I prefer addres assignment to copying:
`existingComponentRoute = &desiredComponentRoute`

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -92,14 +99,13 @@ func (c *oauthsClientsController) getIngressConfig() (*configv1.Ingress, error)
 	return ingress, nil
 }
 
-func (c *oauthsClientsController) getCanonicalRouteHost(ingressConfigDomain string) (string, error) {
+func (c *oauthsClientsController) getCanonicalRouteHost(hostname string) (string, error) {
```
stlaz, line 102: s/hostname/expectedHost

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
stlaz: +1 :slightly_smiling_face:

stlaz (changes requested):
needs an e2e test

stlaz:
> I do not believe that we need to set this value. Doing so would cause the ingress-controller to generate RBAC for the provided users. This controller already has the rbac necessary to read the admin provided secrets.

In the unlikely event we decide to narrow down the RBAC of this operator to only the resources it needs (even though it basically sets up the components that decide who is who), this would still break. Also, this would be helpful if we were to move this controller to library-go.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Given that we are going to move the `handleRoute` method from the `metadataController` I can surface the conditions it generates here.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Can do!

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,307 @@
...
+
+	// Copy source secret to destination
+	if len(foundConditions) == 0 {
+		foundConditions = append(foundConditions, c.syncSecret(ingressConfigCopy)...)
+	}
+
+	return common.UpdateControllerConditions(c.operatorClient, knownConditionNames, foundConditions)
+}
+
+func (c *customRouteController) handleRoute(ctx context.Context, ingress *configv1.Ingress) (conditions []operatorv1.OperatorCondition) {
```
stlaz, line 125: I think we will eventually want @deads2k to bless this PR and I agree with him that the pattern when conditions are returned is making the code hard to read. Maybe just return an error?

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,307 @@
...
+
+	route, err := c.routeClient.Get(ctx, "oauth-openshift", metav1.GetOptions{})
+
+	// Reflect the status of the router in the ingressConfig.status.componentRoute array
+	defer func() {
+		hostname := ""
+		if route != nil {
+			hostname = route.Spec.Host
+		}
+		conditions = append(conditions, c.updateIngressConfigStatus(ctx, hostname, componentRouteConditions(conditions), ingress)...)
```
stlaz, line 136: this should be outside this function

pkg/controllers/customroute/custom_route_controller.go
stlaz: s/handleRoute/applyRoute, pass the expected route as an argument

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	return resourcesynccontroller.ResourceLocation{}
+}
+
+func componentRouteConditions(conditions []operatorv1.OperatorCondition) []metav1.Condition {
+	now := metav1.Now()
+	if len(conditions) == 0 {
+		return []metav1.Condition{
+			{
+				Type:               "Available",
```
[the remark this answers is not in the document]
stlaz: can you think of a scenario where we could go "progressing=true"?

pkg/controllers/metadata/metadata_controller.go
```diff
@@ -78,13 +72,9 @@ func NewMetadataController(kubeInformersForTargetNamespace informers.SharedInfor
 func (c *metadataController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
 	foundConditions := []operatorv1.OperatorCondition{}
 
-	ingress, ingressConditions := common.GetIngressConfig(c.ingressLister, "IngressConfig")
+	_, ingressConditions := common.GetIngressConfig(c.ingressLister, "IngressConfig")
```
stlaz, line 75: we don't need this here anymore, do we?

pkg/operator/assets/bindata.go
```diff
@@ -856,6 +857,41 @@ func oauthOpenshiftOauthServiceYaml() (*asset, error) {
...
 }
 
+var _oauthOpenshiftRouteYaml = []byte(`apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: oauth-openshift
+  namespace: openshift-authentication
+  labels:
+    app: oauth-openshift
+spec:
```
stlaz, line 867: move bindata generation to a separate commit

stlaz (changes requested):

pkg/controllers/customroute/custom_route_controller.go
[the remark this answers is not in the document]
awgreene: > can you think of a scenario where we could go "progressing=true"?

Degraded could be set to true if the secret is invalid.
Progressing could be set to true if the secret is valid and the deployment is rolling out with the new secrets,

pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -46,6 +46,7 @@ func NewConfigObserver(
 		configInformer.Config().V1().Consoles().Informer(),
 		configInformer.Config().V1().Infrastructures().Informer(),
 		configInformer.Config().V1().OAuths().Informer(),
+		configInformer.Config().V1().Ingresses().Informer(),
```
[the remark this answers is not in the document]
awgreene: Done!

pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -82,6 +82,7 @@ func NewConfigObserver(
 		configobservation.Listers{
 			ConfigMapLister: kubeInformersForNamespaces.ConfigMapLister(),
 			SecretsLister:   kubeInformersForNamespaces.SecretLister(),
+			IngressLister:   configInformer.Config().V1().Ingresses().Lister(),
```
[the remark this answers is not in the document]
awgreene: Done.

pkg/controllers/configobservation/routersecret/observe_router_secret_test.go
```diff
@@ -11,9 +11,11 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 
-	"github.com/openshift/library-go/pkg/operator/events"
+	configv1 "github.com/openshift/api/config/v1"
 
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/cluster-authentication-operator/pkg/controllers/configobservation"
+	"github.com/openshift/library-go/pkg/operator/events"
```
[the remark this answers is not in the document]
awgreene: Done.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Added a few basic conditions, still need to add more advanced tests to check service availability.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		Conditions: conditions,
+	}
+
+	if existingComponentRoute := common.GetComponentRouteStatus(ingress, c.componentRoute.Namespace, c.componentRoute.Name); existingComponentRoute == nil {
+		ingress.Status.ComponentRoutes = append(ingress.Status.ComponentRoutes, desiredComponentRoute)
+	} else {
+		existingComponentRoute = &desiredComponentRoute
+	}
+
+	_, err := c.ingressClient.UpdateStatus(ctx, ingress, metav1.UpdateOptions{})
```
awgreene, line 252: I plan to add a "deep equals" check to see if the conditions is the same (minus timestamps).

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+
+func NewCustomRouteController(
+	componentRouteNamespace string,
+	componentRouteName string,
+	destSecretNamespace string,
+	destSecretName string,
+	ingressInformer configinformers.IngressInformer,
+	ingressClient configsetterv1.IngressInterface,
+	routeInformer routeinformer.RouteInformer,
+	routeSetter routeclient.RouteInterface,
```
stlaz, line 60: nit: `routeClient`

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+
+func (c *customRouteController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	ingressConfig, err := c.ingressLister.Get("cluster")
+	if err != nil {
+		return err
+	}
+
+	ingressConfigCopy := ingressConfig.DeepCopy()
+
+	// Create or modify the expected router
```
stlaz, line 105: nit: that's a route you're working with, not a router

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// Sync the secret
+	if err := c.syncSecret(secretName); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getOauthRoute(ingressConfig *configv1.Ingress) *routev1.Route {
```
stlaz, line 131: nit: OAuth

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Sync the secret
+	if err := c.syncSecret(secretName); err != nil {
+		return err
+	}
+
+	return nil
```
stlaz, line 128: `return c.syncSecret(secretName)`

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+					LastTransitionTime: metav1.Now(),
+					Reason:             "InvalidCustomSecret",
+					Message:            fmt.Sprintf("The %s/%s secret is invalid: %v", "openshift-config", secretName, errors),
+				},
+				{
+					Type:               "Progressing",
+					Status:             metav1.ConditionFalse,
+					LastTransitionTime: metav1.Now(),
+					Reason:             "InvalidCustomSecret",
+					Message:            fmt.Sprintf("The %s/%s secret is invalid: %v", "openshift-config", secretName, errors),
```
stlaz, line 218: that's not a reason to go progressing, that's a reason to go degraded

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+			conditions = []metav1.Condition{
+				{
+					Type:               "Degraded",
+					Status:             metav1.ConditionTrue,
+					LastTransitionTime: metav1.Now(),
+					Reason:             "InvalidCustomSecret",
+					Message:            fmt.Sprintf("The %s/%s secret is invalid: %v", "openshift-config", secretName, errors),
+				},
+				{
+					Type:               "Progressing",
```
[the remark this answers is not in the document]
stlaz: Those checks should be outside of the `validateCustomSecret` function, of course

pkg/controllers/customroute/custom_route_controller.go
stlaz: you can have a look at https://github.com/openshift/library-go/blob/7b0fb0ed03f87f3b8db4e71b8ad9c0bf1ec5d3eb/pkg/operator/v1helpers/helpers.go#L157-L185 for inspiration, but yeah, you don't want to do "Updates" if the objects are the same

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -92,13 +101,12 @@ func (c *oauthsClientsController) getIngressConfig() (*configv1.Ingress, error)
 	return ingress, nil
 }
 
-func (c *oauthsClientsController) getCanonicalRouteHost(ingressConfigDomain string) (string, error) {
+func (c *oauthsClientsController) getCanonicalRouteHost(expectedHost string) (string, error) {
```
stlaz, line 104: will this work? IIRC `routeapihelpers.IngressURI` checks for admitted ingresses, will `expectedHost` be among them if it's outside the router's domain?

stlaz (commented):
Mostly just nits. I think we're really getting close now. Still needs an e2e test.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func getConditions(conditions []metav1.Condition) []metav1.Condition {
+	now := metav1.Now()
+	if len(conditions) == 0 {
+		return []metav1.Condition{
+			{
+				Type:               "Degraded",
+				Status:             metav1.ConditionFalse,
+				LastTransitionTime: now,
```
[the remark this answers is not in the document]
awgreene: It is not set by the API Server, as shown below where I attempted to remove it:
```
 CustomRouteController reconciliation failed: Ingress.config.openshift.io "cluster" is invalid: status.componentRoutes.conditions.lastTransitionTime: Required value
 ```

File: test/e2e/custom_route_test.go [hunk 47]
```diff
@@ -0,0 +1,185 @@
...
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	})
+}
+
+func checkRouteHostname(t *testing.T, routeClient *routeclient.Clientset, routeNamespace string, routeName string, hostname string) error {
+	t.Helper()
+	return wait.PollImmediate(time.Second, 10*time.Minute, func() (bool, error) {
```
stlaz, line 187: 10 minutes is a bit too much IMO

test/e2e/custom_route_test.go
```diff
@@ -0,0 +1,185 @@
...
+			return false, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		return route.Spec.Host == hostname, nil
+	})
+}
+
+func removeComponentRoute(t *testing.T, configClient *configclient.Clientset, namespace string, name string) error {
```
stlaz, line 200: why is this done in a wait?

pkg/controllers/common/ingress_config.go
```diff
@@ -72,3 +78,42 @@ func GetCustomRouteHostname(ingress *configv1.Ingress, namespace string, name st
...
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		ingressConfig, err := client.Get(ctx, "cluster", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		existingComponentRoute := GetComponentRouteStatus(ingressConfig, componentRoute.Namespace, componentRoute.Name)
+		if existingComponentRoute != nil {
+			// Ignore timestamps on the existing route
+			setLastTransactionTime(existingComponentRoute, now)
```
stlaz, line 102: you shouldn't change the input, at least make a copy

test/e2e/custom_route_test.go
```diff
@@ -104,6 +104,50 @@ func TestCustomRouterCerts(t *testing.T) {
 
 	err = checkRouteHostname(t, routeClient, "openshift-authentication", "oauth-openshift", "foo.bar.com")
 	require.NoError(t, err)
+
+	err = checkCustomSecretIsMountedOntoOauthDeployment(t, kubeClient)
+	require.NoError(t, err)
+}
+
+func checkCustomSecretIsMountedOntoOauthDeployment(t *testing.T, kubeClient *kubernetes.Clientset) error {
```
stlaz, line 112: you should check that your content is being served on the route instead

stlaz (changes requested):

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	if _, _, err := routeapihelpers.IngressURI(route, expectedRoute.Spec.Host); err != nil {
+		// be careful not to print route.spec as it many contain secrets
+		return err
+	}
+
+	if _, err := c.secretLister.Secrets("openshift-authentication").Get("v4-0-config-system-router-certs"); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
awgreene: This was present when I ported it from the metadata controller, I'll remove it.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	var customRouteErrors []error
+
+	// configure the expected route
+	expectedRoute, secretName, err := c.getOAuthRouteAndSecretName(ingressConfigCopy)
+	if err != nil {
+		customRouteErrors = append(customRouteErrors, err)
+	}
+
+	// create or modify the existing route
+	if err = c.applyRoute(ctx, expectedRoute); err != nil {
```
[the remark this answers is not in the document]
awgreene: @stlaz and I spoke offline and agreed to fail if the provided configuration is invalid.

test/e2e/custom_route_test.go
```diff
@@ -0,0 +1,185 @@
...
+	require.NoError(t, err)
+
+	// generate crypto materials
+	rootCA := e2e.NewCertificateAuthorityCertificate(t, nil)
+	intermediateCA := e2e.NewCertificateAuthorityCertificate(t, rootCA)
+	clientCA := e2e.NewClientCertificate(t, intermediateCA, "foo.bar.com")
+
+	// create tls secret
+	var certificates bytes.Buffer
+	certificates.Write(pem.EncodeToMemory(&pem.Block{Type: cert.CertificateBlockType, Bytes: clientCA.Certificate.Raw}))
```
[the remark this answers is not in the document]
awgreene: Ah, yes.

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
awgreene: Removed in the latest version of the PR.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Thanks and done!

test/e2e/custom_route_test.go
```diff
@@ -0,0 +1,185 @@
...
+	require.NoError(t, err)
+	configClient, err := configclient.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+	routeClient, err := routeclient.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+
+	// generate crypto materials
+	rootCA := e2e.NewCertificateAuthorityCertificate(t, nil)
+	intermediateCA := e2e.NewCertificateAuthorityCertificate(t, rootCA)
+	clientCA := e2e.NewClientCertificate(t, intermediateCA, "foo.bar.com")
```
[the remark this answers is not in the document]
awgreene: This was a misunderstanding on my part. The cert validation code I was using was meant for client certs. I should use my own validation code and it should be used for server certs.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	// Check if a serving secret is provided
+	secretName := common.GetCustomRouteServingCertKeySecret(ingressConfig, OAuthComponentRouteNamespace, OAuthComponentRouteName)
+	err := c.validateCustomTLSSecret(secretName)
+	if err != nil {
+		secretName = ""
+		// Do not use provided hostname if an invalid custom secret is provided
+		route.Spec.Host = "oauth-openshift." + ingressConfig.Spec.Domain // mimic the behavior of subdomain
+	}
```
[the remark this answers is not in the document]
awgreene: Done

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// sync the secret
+	return c.syncSecret(secretName)
+}
+
+func (c *customRouteController) getOAuthRouteAndSecretName(ingressConfig *configv1.Ingress) (*routev1.Route, string, error) {
+	route := resourceread.ReadRouteV1OrDie(assets.MustAsset("oauth-openshift/route.yaml"))
+
+	// Assume that we will use a custom hostname, if no custom hostname is provided use the default.
+	route.Spec.Host = common.GetCustomRouteHostname(ingressConfig, OAuthComponentRouteNamespace, OAuthComponentRouteName)
```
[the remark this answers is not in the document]
awgreene: Done

pkg/controllers/common/ingress_config.go
awgreene: Done.

test/e2e/custom_route_test.go
awgreene: Done.

test/e2e/custom_route_test.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+		}, metav1.CreateOptions{})
+	require.NoError(t, err)
+	defer func() {
+		removeComponentRoute(t, configClient, "openshift-authentication", "oauth-openshift")
+		_ = kubeClient.CoreV1().Secrets(secret.Namespace).Delete(context.TODO(), secret.Name, metav1.DeleteOptions{})
+	}()
+
+	// Set a custom hostname without a secret
+	err = getAndUpdateComponentRoute(t, configClient, &configv1.ComponentRouteSpec{
```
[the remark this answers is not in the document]
awgreene: The operator can enter a degraded state for a few minutes when moving towards custom routes because the expected metadata returned by the well-known endpoint isn't updated quick enough.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,273 @@
...
+	now := metav1.Now()
+	if len(errs) == 0 {
+		return []metav1.Condition{
+			{
+				Type:               "Degraded",
+				Status:             metav1.ConditionFalse,
+				LastTransitionTime: now,
+				Reason:             "RouteIsAvailable",
+				Message:            "Route successfully configured",
+			},
```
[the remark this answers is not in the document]
awgreene: This condition is now reported in the componentRoute conditions.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,273 @@
...
+			{
+				Type:               "Progressing",
+				Status:             metav1.ConditionFalse,
+				LastTransitionTime: now,
+				Reason:             "RouteIsAvailable",
+				Message:            "Route successfully configured",
+			},
```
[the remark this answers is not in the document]
awgreene: This condition is now reported in the componentRoute conditions.

test/e2e/custom_route_test.go
awgreene: Reducing to 1 minute.

awgreene: I reduced all poll functions to a duration of 1 minute other than the `pollForCustomServingCertificates`, which can take a few minutes to take affect.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+				Reason:             "ErrorReachingOutToService",
+				Message:            fmt.Sprintf("unexpected error at %s: %v", route.Spec.Host, err),
+			},
+			{
+				LastTransitionTime: now,
+				Type:               "Progressing",
+				Status:             metav1.ConditionTrue,
+				Reason:             "ErrorReachingOutToService",
+				Message:            fmt.Sprintf("unexpected error at %s: %v", route.Spec.Host, err),
+			},
```
stlaz, line 327: why is this degraded=false,progressing=true?

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+		}
+	}
+
+	return []metav1.Condition{
+		{
+			LastTransitionTime: now,
+			Type:               "Degraded",
+			Status:             metav1.ConditionFalse,
+			Reason:             "RouteIsAvailable",
```
stlaz, line 336: I think we usually go with
```
Reason: AsExpected
Message: All is well
```

https://github.com/openshift/cluster-authentication-operator/blob/4363b9ba94615e3868b9bb3b1b5bb633b8843da2/vendor/github.com/openshift/library-go/pkg/operator/status/condition.go#L65-L75

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+func (c *customRouteController) checkRouteAvailable(host string, ingress *configv1.Ingress) error {
+	url := "https://" + host
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return err
+	}
+
+	reqCtx, cancel := context.WithTimeout(context.TODO(), 10*time.Second) // avoid waiting forever
+	defer cancel()
+	req.WithContext(reqCtx)
```
stlaz, line 358: this does nothing, look at `http.NewRequestWithContext` (I learned this yesterday)

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+
+	secret, isDefault, err := common.GetActiveRouterSecret(c.secretLister, "openshift-authentication", "v4-0-config-system-router-certs", "openshift-authentication", "v4-0-config-system-custom-router-certs")
+	if err != nil {
+		return err
+	}
```
stlaz, line 363: use the secret you retrieved earlier

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+	destSecret     types.NamespacedName
+	componentRoute types.NamespacedName
+	ingressLister  configlistersv1.IngressLister
+	ingressClient  configsetterv1.IngressInterface
+	routeLister    routev1lister.RouteLister
+	routeClient    routeclient.RouteInterface
+	secretLister   corev1listers.SecretLister
+	resourceSyncer resourcesynccontroller.ResourceSyncer
+	operatorClient v1helpers.OperatorClient
+	httpClient     http.Client
```
stlaz, line 60: don't store the client, create it wherever you need it

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+				LastTransitionTime: now,
+				Type:               "Progressing",
+				Status:             metav1.ConditionFalse,
+				Reason:             "RouteNotAdmitted",
+				Message:            fmt.Sprintf("Route not admitted: %v", err),
+			},
+		}
+	}
+
+	if err := c.checkRouteAvailable(route.Spec.Host, ingress); err != nil {
```
stlaz, line 312: I would just push this outside in the sync once you've got a validated secret and then use the secret inside this function

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 403 {
+		return fmt.Errorf("%q returned %d", url, resp.StatusCode)
+	}
+
+	if resp.TLS == nil {
+		return fmt.Errorf("response TLS field is nil")
```
stlaz, line 400: this is not a very user friendly message

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+		source = resourcesynccontroller.ResourceLocation{Namespace: "openshift-config", Name: secretName}
+	}
+
+	destination := resourcesynccontroller.ResourceLocation{Namespace: c.destSecret.Namespace, Name: c.destSecret.Name}
+
+	if err := c.resourceSyncer.SyncSecret(destination, source); err != nil {
+		return err
+	}
+
+	return nil
```
stlaz, line 459: `return c.resourceSyncer.SyncSecret(destination, source)`

pkg/controllers/routercerts/controller.go
```diff
@@ -46,19 +50,24 @@ func NewRouterCertsDomainValidationController(
...
+		ingressLister:          ingressInformer.Lister(),
+		secretLister:           secretInformer.Lister(),
+		configMapLister:        configMapInformer.Lister(),
+		defaultSecretNamespace: defaultSecretNamespace,
+		defaultSecretName:      defaultSecretName,
+		customSecretNamespace:  customSecretNamespace,
+		customSecretName:       customSecretName,
+		routeName:              routeName,
+		systemCertPool:         x509.SystemCertPool,
+		usingDefaultSecret:     false,
```
stlaz, line 70: don't cache what you're newly caching

pkg/controllers/routercerts/controller.go
```diff
@@ -103,21 +126,27 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
+	var secret *corev1.Secret
+	secret, c.usingDefaultSecret, err = common.GetActiveRouterSecret(c.secretLister, c.defaultSecretNamespace, c.defaultSecretName, c.customSecretNamespace, c.customSecretName)
 	if err != nil {
-		return newRouterCertsDegradedf("NoRouterCertSecret", "secret/%v -n %v: could not be retrieved: %v", c.secretName, c.targetNamespace, err)
+		return newRouterCertsDegradedf("NoRouterCertSecret", "secret/%v -n %v: could not be retrieved: %v", c.getActiveSecretName(), c.getActiveSecretNamespace(), err)
+	}
+
+	certKey := ingress.Spec.Domain
+	if !c.usingDefaultSecret {
+		certKey = corev1.TLSCertKey
```
stlaz, line 137: looks more like you need a unified way to retrieve the secret's cert and key

stlaz (changes requested):

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+		return fmt.Errorf("provided hostname (%s) is invalid, must match %s", hostname, hostnameRegex)
+	}
+
+	return nil
+}
+
+func (c *customRouteController) applyRoute(ctx context.Context, expectedRoute *routev1.Route) error {
+	route, err := c.routeClient.Get(ctx, "oauth-openshift", metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		route, err = c.routeClient.Create(ctx, expectedRoute, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
awgreene: There is an `if` statement immediately after this which handles:
* non-IsNotFound errors from the  `c.routeClient.Get` call
* errors from the `c.routeClient.Create` call

pkg/controllers/customroute/custom_route_controller.go
awgreene: I'll use the `http.NewRequestWithContext` as suggested, thanks!

awgreene: [This bit](https://golang.org/pkg/net/http/#Request.WithContext) of documentation for `WithContext` was helpful as well.

pkg/controllers/customroute/custom_route_controller.go
awgreene: How does `fmt.Errorf("Unable to retrieve TLS information from %s", url)` sound?

stlaz: much better :+1:

pkg/controllers/customroute/custom_route_controller.go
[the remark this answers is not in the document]
stlaz: yes, but you don't need to be doing anything after `Create()` because the route has just been created exactly how you needed it, no?

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+	if *modified || !equality.Semantic.DeepEqual(existingCopy.Spec, expectedRoute.Spec) {
+		// be careful not to print route.spec as it many contain secrets
+		existingCopy.Spec = expectedRoute.Spec
+		route, err = c.routeClient.Update(ctx, existingCopy, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
+	_, _, err = routeapihelpers.IngressURI(route, expectedRoute.Spec.Host)
```
stlaz, line 240: I might have asked this before and this might have been answered, but - why is this here?

awgreene: This already existed when I pulled it into the package, I'll remove it.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+				Reason:             "CustomRouteError",
+				Message:            fmt.Sprintf("Error Configuring custom route: %v", errs),
+			},
+			{
+				LastTransitionTime: now,
+				Type:               "Progressing",
+				Status:             metav1.ConditionFalse,
+				Reason:             "CustomRouteError",
+				Message:            fmt.Sprintf("Error Configuring custom route: %v", errs),
+			},
```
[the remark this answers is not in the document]
awgreene: In this instance, the route is not progressing because the configured route was deemed invalid, meaning it is unrecoverable and can't progress.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Good question - I had assumed that the admitted route might eventually become available. Instead, we may want to follow your suggestion [here](https://github.com/openshift/cluster-authentication-operator/pull/430#discussion_r623079557) and report:
* Degraded = false and progressing = true  if less than some time has passed (say 5 minutes)
* Degraded = true and progressing = false  if 5 minutes has passed

pkg/controllers/customroute/custom_route_controller.go
awgreene: Right now this code only validates custom serving secrets, it would not handle cases where the default cert secret is in use.

pkg/controllers/routercerts/controller.go
awgreene: Added a method to the common package

pkg/controllers/routercerts/controller.go
awgreene: Done.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Done.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Updated.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+
+	if _, _, err := routeapihelpers.IngressURI(route, route.Spec.Host); err != nil {
+		return []metav1.Condition{
+			{
+				LastTransitionTime: now,
+				Type:               "Degraded",
+				Status:             metav1.ConditionTrue,
+				Reason:             "RouteNotAdmitted",
+				Message:            fmt.Sprintf("Route not admitted: %v", err),
+			},
```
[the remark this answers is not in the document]
awgreene: Done.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,509 @@
...
+
+func validateCustomHostname(hostname string) error {
+	re, err := regexp.Compile(hostnameRegex)
+	if err != nil {
+		return err
+	}
+
+	if !re.MatchString(hostname) {
+		return fmt.Errorf("provided hostname (%s) is invalid, must match %s", hostname, hostnameRegex)
+	}
```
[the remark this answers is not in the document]
awgreene: Done.

pkg/controllers/common/ingress_config.go
```diff
@@ -28,3 +32,60 @@ func GetIngressConfig(ingressLister configv1lister.IngressLister, conditionPrefi
...
+	if componentRoute := GetComponentRouteSpec(ingress, namespace, name); componentRoute != nil {
+		return string(componentRoute.Hostname)
+	}
+	return ""
+}
+
+// GetActiveRouterSecret returns the secret that contains the serving certificates for the openshift-authentication/oauth-openshift
+// route, a boolean representing if the default openshift-authentication/v4-0-config-system-router-certs secret is being used, and
+// any errors in retrieving the active secret.
+func GetActiveRouterSecret(secretsLister corev1listers.SecretLister, defaultSecretNamespace string, defaultSecretName string, customSecretNamespace string, customSecretName string) (*corev1.Secret, bool, error) {
```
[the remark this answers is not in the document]
awgreene: Done.

pkg/controllers/customroute/custom_route_controller.go
awgreene: Conditions should be good to go now.

test/e2e/custom_route_test.go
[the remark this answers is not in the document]
stlaz: this should now be fixed as https://github.com/openshift/cluster-authentication-operator/pull/442 merged

stlaz (commented):
I think these are mostly cosmetic changes now

pkg/controllers/routercerts/controller.go
```diff
@@ -173,6 +186,11 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates(ingre
 		klog.Infof("system cert pool not available: %v", err)
 		verifyOptions.Roots = x509.NewCertPool()
 	}
+	if !common.IsDefaultSecret(secret) {
+		verifyOptions.Roots.AddCert(certificates[0])
```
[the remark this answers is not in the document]
awgreene: The `populateVerifyOptionsFromCertSlice(opts *x509.VerifyOptions, certs []*x509.Certificate)` function takes the list of certificates and categorizes them as roots, intermediates, and server certificates. I believe that there are cases where the custom certificate may not be a CA, causing it to be categorized as a serverCert.

The `Verify` command is later ran against the certificate pol, which attempts to Verify attempts to verify c by building one or more chains from c to a certificate in opts.Roots, using certificates in opts.Intermediates if needed. Given that the custom routes could be entirely unique, the  `populateVerifyOptionsFromCertSlice` does not meet our needs.

pkg/controllers/customroute/custom_route_conditions.go
```diff
@@ -0,0 +1,218 @@
...
+func findCondition(conditions []metav1.Condition, conditionType string) *metav1.Condition {
+	for i := range conditions {
+		if conditions[i].Type == conditionType {
+			return &conditions[i]
+		}
+	}
+	return nil
+}
+
+func checkCustomRouteConfiguration(errors []error) []metav1.Condition {
```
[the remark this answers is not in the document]
awgreene: I'll rename it to `checkErrorsConfiguringCustomRoute`.

pkg/controllers/customroute/custom_route_conditions.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// Compare the certificates served against those defined in the secret
+	pubKeys, err := parseCertificates(data)
+	if err != nil {
+		return err
+	}
+
+	for _, expectedCert := range resp.TLS.PeerCertificates {
+		found := false
+		for _, pubKey := range pubKeys {
```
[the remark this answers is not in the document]
awgreene: Can you elaborate on this? How can I identify the server certificate here?

pkg/controllers/routercerts/controller.go
```diff
@@ -186,10 +180,9 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
 		klog.Infof("system cert pool not available: %v", err)
 		verifyOptions.Roots = x509.NewCertPool()
 	}
-	if !common.IsDefaultSecret(secret) {
+	if !isDefault {
```
stlaz, line 183: NOTE to me: I still need to think about this.

stlaz (commented):
only a few comments now, we can take the PR over, I don't want to keep Alex from his work

pkg/controllers/routercerts/controller.go
stlaz: yeah, remove this, it's effectively a no-op and so this controller should be a no-op when custom route certificate is supplied. It was designed to check the validity of the router-certs secret, the route's TLS validity is checked in the other endpointcheckers anyway.

stlaz (changes requested):
we need to make the routercerts controller a noop on custom certs

pkg/controllers/routercerts/controller.go
```diff
@@ -139,20 +139,30 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
 
 	// get router certs secret
-	certBytes, _, isDefault, err := common.GetActiveRouterCertKeyBytes(c.secretLister, ingress, c.secretNamespace, c.defaultSecretName, c.customSecretName)
+	secret, err := common.GetActiveRouterSecret(c.secretLister, c.secretNamespace, c.defaultSecretName, c.customSecretName)
 	if err != nil {
 		return newRouterCertsDegradedf("NoRouterCertSecret", "neither the custom secret/%v -n %v or default secret/%v -n %v could be retrieved: %v", c.defaultSecretName, c.secretNamespace, c.customSecretName, c.secretNamespace, err)
 	}
 
+	// Perform a no-op if non-default secret is in use
+	if secret.GetName() != c.defaultSecretName {
```
stlaz, line 148: remove the `c.defaultSecretName` in favour of a constant string

pkg/controllers/routercerts/controller.go
```diff
@@ -139,20 +139,30 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
+			Type:   conditionRouterCertsDegradedType,
+			Status: operatorv1.ConditionFalse,
+			Reason: "AsExpected",
+		}
+	}
+
 	// cert data should exist
-	if len(certBytes) == 0 {
-		return newRouterCertsDegradedf("MissingRouterCertsPEM", "secret/%v -n %v: certificate not found", c.activeSecretName(isDefault), c.secretNamespace)
+	data := secret.Data[ingressDomain]
```
stlaz, line 157: nit: `s/data/certBytes/`

awgreene: I'm going to leave this as data because I reverted the earlier changes of this file based on the decision to perform a no-op when the default secret is not in use.

test/e2e/custom_route_test.go
```diff
@@ -0,0 +1,236 @@
...
+		Namespace: "openshift-authentication",
+		Name:      "oauth-openshift",
+		Hostname:  "foo.bar.com",
+	})
+	require.NoError(t, err)
+
+	// check that the hostname was updated
+	err = checkRouteHostname(t, routeClient, "openshift-authentication", "oauth-openshift", "foo.bar.com")
+
```
slaskawi, line 83: I think we need `require.NoError(t, err)` here.

slaskawi (approved):
One possible (small) issue found. Otherwise LGTM

/lgtm

stlaz:
/lgtm cancel
Unit tests are failing:
```
FAIL: github.com/openshift/cluster-authentication-operator/pkg/controllers/routercerts TestValidateRouterCertificates/CustomSecretMissingRouterCertsPEM 0s
FAIL: github.com/openshift/cluster-authentication-operator/pkg/controllers/routercerts TestValidateRouterCertificates/CustomSecretMalformedRouterCertsPEM 0s
FAIL: github.com/openshift/cluster-authentication-operator/pkg/controllers/routercerts TestValidateRouterCertificates/CustomSecretNoServerCertRouterCerts 0s
FAIL: github.com/openshift/cluster-authentication-operator/pkg/controllers/routercerts TestValidateRouterCertificates/CustomSecretInvalidServerCertRouterCertsBadChain 0s
FAIL: github.com/openshift/cluster-authentication-operator/pkg/controllers/routercerts TestValidateRouterCertificates 19.49s 
```

slaskawi (approved):
/lgtm

pkg/operator/datasync/validation.go
```diff
@@ -78,7 +78,20 @@ func validateClientCert(pem []byte) []error {
...
-func validatePrivateKey(pemKey []byte) []error {
+func ValidateServerCert(pem []byte) []error {
+	errs := []error{}
+
+	certs, certErrs := parseCerts(pem)
+	errs = append(errs, certErrs...)
+
+	if numCerts := len(certs); numCerts != 1 {
+		return append(errs, fmt.Errorf("expected a single server certificate, got %d", numCerts))
+	}
```
stlaz, line 89: The server cert PEM will likely contain the full certificate chain so I would remove this check

stlaz (changes requested):

pkg/operator/datasync/validation.go
awgreene: Will update the check to expect at least one cert.

PR #432, opened 2021-04-09
e2e: fix the token names in tokentimeout test to match the new format

stlaz:


test/library/client.go
```diff
@@ -22,3 +25,14 @@ func NewClientConfigForTest(t *testing.T) *rest.Config {
...
 	return config
 }
+
+// GenerateOAuthTokenPair returns two tokens to use with OpenShift OAuth-based authentication.
+// The first token is a private token meant to be used as a Bearer token to send
+// queries to the API, the second token is a hashed token meant to be stored in
+// the database.
+func GenerateOAuthTokenPair() (privToken, pubToken string) {
+	const sha256Prefix = "sha256~"
+	randomToken := base64.RawURLEncoding.EncodeToString([]byte(fmt.Sprintf("nottoorandom%d", mathrand.Int())))
```
p0lyn0mial, line 35: ok, bearer tokens should be base64 encoded.

stlaz: Funny you noticed, I think this is actually copy-paste where only certain parts were changed in order to make this simpler for tests. I'll make this simpler not to waste that precious CPU time

PR #433, opened 2021-04-14
Bug 1946540: only configure webhook authenticators when oauth-apiservers are ready during upgrade

stlaz:
If we configure the webhook authenticators too early during an upgrade,
we might end up with apiservers that are incapable of handling
TokenReviews. Make sure that the version of the oauth-apiserver payload
in the version getter global to the operator is current if it is set.
If it is unset, this should not be an upgrade.

/assign @deads2k

### DIFF
--- a/pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
+++ b/pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -14,6 +15,7 @@ import (
 	"k8s.io/client-go/informers"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/flowcontrol"
 
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
@@ -22,6 +24,7 @@ import (
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	"github.com/openshift/cluster-authentication-operator/pkg/operator/assets"
@@ -41,6 +44,9 @@ type webhookAuthenticatorController struct {
 
 	operatorConfigClient operatorconfigclient.AuthenticationsGetter
 	operatorClient       v1helpers.OperatorClient
+
+	apiServerVersionWaitEventsLimiter flowcontrol.RateLimiter
+	versionGetter                     status.VersionGetter
 }
 
 func NewWebhookAuthenticatorController(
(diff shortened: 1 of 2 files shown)


pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -74,6 +81,18 @@ func NewWebhookAuthenticatorController(
 }
 
 func (c *webhookAuthenticatorController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	// TODO: remove in 4.8; this is to ensure we don't configure webhook authenticators
```
deads2k, line 84: remove in 4.8?

stlaz: oops, 4.9

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -74,6 +81,18 @@ func NewWebhookAuthenticatorController(
...
+	// before the oauth-apiserver revision that's capable of handling it is ready
+	// during upgrade
+	versions := c.versionGetter.GetVersions()
+	if apiserverVersion, ok := versions["oauth-apiserver"]; ok {
+		// a previous version found means this could be an upgrade, unless the version is already current
+		if expectedVersion := os.Getenv("IMAGE_OAUTH_APISERVER"); apiserverVersion != expectedVersion {
+			klog.V(4).Infof("the oauth-apiserver hasn't reported it version to be %q yet, its current version is %q", expectedVersion, apiserverVersion)
```
deads2k, line 91: create an apimachinery ratelimiter and use the `.Try` method to decide to send an event every minute we go down this path so you can tell it's happening.

deads2k:
/lgtm
/hold

Confirm that you see the correct event in the update job log and link it here before removing the hold please.

stlaz:
/hold cancel
/retest

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-authentication-operator/433/pull-ci-openshift-cluster-authentication-operator-master-e2e-agnostic-upgrade/1382350081788743680/artifacts/e2e-agnostic-upgrade/gather-extra/artifacts/events.json
```json
        {
            "apiVersion": "v1",
            "count": 5,
            "eventTime": null,
            "firstTimestamp": "2021-04-14T18:00:32Z",
            "involvedObject": {
                "apiVersion": "apps/v1",
                "kind": "Deployment",
                "name": "authentication-operator",
                "namespace": "openshift-authentication-operator",
                "uid": "628f51c4-af3f-4e7e-a1f6-fe928fbb41ef"
            },
            "kind": "Event",
            "lastTimestamp": "2021-04-14T18:04:32Z",
            "message": "the oauth-apiserver hasn't reported it version to be \"registry.build01.ci.openshift.org/ci-op-vxjc7gzx/stable@sha256:1ed70b28ade9db93ee65f046ad0629cc01a01c45d7279261733e8b626386bcb7\" yet, its current version is \"4.8.0-0.ci.test-2021-04-14-150848-ci-op-vxjc7gzx\"",
            "metadata": {
                "creationTimestamp": "2021-04-14T18:00:32Z",
                "name": "authentication-operator.1675cb5ad33ba3fc",
                "namespace": "openshift-authentication-operator",
                "resourceVersion": "48178",
                "uid": "76960f76-0b0d-4972-8150-3845e451de06"
            },
            "reason": "OAuthAPIServerWaitForLatest",
            "reportingComponent": "",
            "reportingInstance": "",
            "source": {
                "component": "oauth-apiserver-webhook-authenticator-controller-webhookauthenticatorcontroller"
            },
            "type": "Normal"
        },
```

PR #435, opened 2021-04-20
Bug 1946540: fix configuring webhook authenticators on upgrade

stlaz:
Wrong version source was used for the oauth-apiserver version
and so the webhook authenticators would never be used on
upgraded clusters.

### DIFF
--- a/pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
+++ b/pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
@@ -89,9 +89,9 @@ func (c *webhookAuthenticatorController) sync(ctx context.Context, syncCtx facto
 	versions := c.versionGetter.GetVersions()
 	if apiserverVersion, ok := versions["oauth-apiserver"]; ok {
 		// a previous version found means this could be an upgrade, unless the version is already current
-		if expectedVersion := os.Getenv("IMAGE_OAUTH_APISERVER"); apiserverVersion != expectedVersion {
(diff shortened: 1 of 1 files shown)


stlaz:
I can see that the event is there in the correct form, finally, and the webhook authenticators are configured:

```
- apiVersion: v1
  count: 5
  eventTime: null
  firstTimestamp: "2021-04-20T11:24:08Z"
  involvedObject:
    apiVersion: apps/v1
    kind: Deployment
    name: authentication-operator
    namespace: openshift-authentication-operator
    uid: b60659f9-724e-480e-88ee-79e11ae23a06
  kind: Event 
  lastTimestamp: "2021-04-20T11:28:08Z"
  message: the oauth-apiserver hasn't reported its version to be "4.8.0-0.ci.test-2021-04-20-091746-ci-op-7zhlxkwl"
    yet, its current version is "4.8.0-0.ci.test-2021-04-20-091350-ci-op-7zhlxkwl"
  metadata:
    creationTimestamp: "2021-04-20T11:24:09Z"
    name: authentication-operator.16778d34b937f09e
    namespace: openshift-authentication-operator
    resourceVersion: "49257"
    uid: d23187ff-ee43-4a28-a563-3fb4e8e1da2a
  reason: OAuthAPIServerWaitForLatest
  reportingComponent: ""
  reportingInstance: ""
  source:
    component: oauth-apiserver-webhook-authenticator-controller-webhookauthenticatorcontroller
  type: Normal
```

PR #438, opened 2021-04-22
 Bug 1950379: routersecret: sync only the cert/key pair for the default domain

stlaz:
During e2e testing, the routing team is adding additional router which revealed that we may rollout unnecessarily as we don't particularly care about any other but the default router cert/keys.

/hold
contains fake bump on a PoC in lib-go

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,13 +4,12 @@ go 1.15
 
 require (
 	github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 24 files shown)


stlaz:
/hold cancel
lib-go PR merged

stlaz:
/retest
sdn sandbox creation failure

stlaz:
/retest
`error adding container to network "openshift-sdn": failed to create container veth: failed to make veth pair: operation not permitted`

PR #440, opened 2021-04-28
[4.7] Bug 1941840: endpoints controller: close response bodies

stlaz:


stlaz:
Turns out that the memory consumption is triggered on proxy configuration. While the network checkers not closing the bodies was bad, the proxy checker actually uses custom CAs, and when such client hangs out idly on the heap, so do the certificate bundles. And that can eat a lot of memory.

heap

rhpmali:
Added comment in the https://bugzilla.redhat.com/show_bug.cgi?id=1941840#c25 , Still observing memory increase.

stlaz:
There was a missing commit that I failed to push to this backport :|

rhpmali:
Added comment in BZ : https://bugzilla.redhat.com/show_bug.cgi?id=1941840#c27 

/lgtm

PR #441, opened 2021-04-28
Bug 1954544: endpoints controller: actually use the context-constrained request

stlaz:
/assign @sttts


stlaz:
turns out the same code lives in the proxychecker :|

PR #442, opened 2021-05-03
Bug 1939580: wellknownready: delay going degraded on oauthmetadata changes

stlaz:
/cc @s-urbaniak @sttts

### DIFF
--- a/pkg/controllers/common/conditions.go
+++ b/pkg/controllers/common/conditions.go
@@ -2,12 +2,80 @@ package common
 
 import (
 	"strings"
+	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
+
+var _ error = &ControllerProgressingError{}
+
+type ControllerProgressingError struct {
+	reason string
+	err    error
+	maxAge time.Duration
+}
+
+// NewControllerProgressingError creates a ControllerProgressingError that can be
+// handled by the controllers in a generic manner to go Progressing instead of Degraded.
+// The error allows configuring `maxAge` which determines for how long can an
+// this error appear in the operator's status before it goes Degraded:
+// maxAge:
+//   <= 0 - never go Degraded
+//   > 0  - go Degraded if the controller was progressing with this error longer than `maxAge`
+func NewControllerProgressingError(reason string, err error, maxAge time.Duration) *ControllerProgressingError {
+	return &ControllerProgressingError{
+		reason: reason,
+		err:    err,
+		maxAge: maxAge,
+	}
+}
+
+func (e *ControllerProgressingError) Error() string {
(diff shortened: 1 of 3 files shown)


pkg/controllers/common/conditions.go
```diff
@@ -2,13 +2,67 @@ package common
...
 
+var _ error = &ControllerProgressingError{}
+
+type ControllerProgressingError struct {
+	reason string
+	err    error
+	maxAge time.Duration
+}
+
+func NewControllerProgressingError(reason string, err error, maxAge time.Duration) *ControllerProgressingError {
```
sttts, line 21: godoc, what is this thing doing?

pkg/controllers/common/conditions.go
```diff
@@ -2,13 +2,67 @@ package common
...
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+var _ error = &ControllerProgressingError{}
+
+type ControllerProgressingError struct {
```
s-urbaniak, line 15: mini nit: i guess this could be package private (lower case)

pkg/controllers/common/conditions.go
```diff
@@ -2,13 +2,67 @@ package common
...
+	if lastCondition == nil {
+		return false
+	}
+
+	if lastCondition.Reason != e.reason || lastCondition.Message != e.Error() {
+		return false
+	}
+
+	if lastTransition := lastCondition.LastTransitionTime; !lastTransition.IsZero() && e.maxAge > 0 {
+		return lastTransition.Add(e.maxAge).Before(time.Now())
```
wking, line 57: Looks like you've already got this approach working, but if you wanted something that's currently on the shelf in library-go, you could use [`WithDegradedInertia`][1] to set degraded inertia times for particular condition groups.

[1]: https://github.com/openshift/library-go/pull/632/files#diff-fb5ae39018b57446d1f9b76bb37b9d342df79bb58793ff36aafa362d907f83ccR302-R312

stlaz: I think you might have sent me this once but to be honest, I completely forgot, so thank you for reminding me :+1: 

I believe this PR still has something to offer. I want the errors I am adding here to be specifically turned into a progressing condition as they mark the time during which the controller is waiting for the kube-apiservers to rollout. The inertia seems more like a means to mute the error for a given time, which may be useful, yet the use-case is a little bit different.

I can see how we could handle this error type easily in https://github.com/openshift/library-go/blob/master/pkg/controller/factory/base_controller.go#L218, which, to me, looks like a pretty elegant solution.

pkg/controllers/common/conditions.go
stlaz: I think errors should better be exported in case someone wants to check on a kind of an error, even though such use might be unlikely in this case

PR #444, opened 2021-05-07
Bug 1953041: manifests/deployment: bump cpu and memory resource requests

s-urbaniak:
This is based on the following measurements on an idle 4.8-ci cluster:

CPU:
image

Memory:
image

These values are higher than the current settings, but let me know if this increase is concerning.

/cc @sttts @stlaz


stlaz:
/lgtm
seems more realistic than the original given how the operator grew

PR #458, opened 2021-06-29
Bug 1977054: observe api-audiences for the oauth-apiserver

stlaz:
When the kube-apiserver receives a token validation request,
that request is audience-constrained since we configure the SA
issuer and audiences. This means that kube-apiserver also checks
the audiences returned from a TokenReview during webhook token
authentication.

The oauth-apiserver by default uses the https://kubernetes.default.svc
audiences. If authentication.spec.serviceAccountIssuer is set however,
the kube-apiserver will expect different audiences from the default.
If the oauth-apiserver does not adjust the TokenReview endpoint,
users won't be able to validate their OAuthAccessTokens as the
audience returned by the oauth-apiserver in a TokenReview response
would differ

/assign @s-urbaniak

### DIFF
--- a/pkg/operator/configobservation/authentication/apiaudiences.go
+++ b/pkg/operator/configobservation/authentication/apiaudiences.go
@@ -1,0 +1,104 @@
+package auth
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"k8s.io/klog/v2"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/operator/configobservation"
+)
+
+var (
+	audiencesPath = []string{"apiServerArguments", "api-audiences"}
+)
+
+// ObserveAPIAudiences changes apiServerArguments.api-audiences from
+// the default value if Authentication.Spec.ServiceAccountIssuer specifies a valid
+// non-empty value.
+func ObserveAPIAudiences(
+	genericListers configobserver.Listers,
+	recorder events.Recorder,
+	existingConfig map[string]interface{},
+) (ret map[string]interface{}, errs []error) {
+	defer func() {
+		ret = configobserver.Pruned(ret, audiencesPath)
+	}()
+
+	listers := genericListers.(configobservation.Listers)
+	errs = []error{}
+	var issuerChanged bool
+	var existingAudience, newAudience string
+	// when the issuer will change, indicate that by setting `issuerChanged` to true
+	// to emit the informative event
+	defer func() {
+		if !issuerChanged {
+			return
+		}
+
+		recorder.Eventf(
+			"ObserveAPIAudiences",
+			"service account issuer changed from %s to %s",
+			existingAudience, newAudience,
+		)
+	}()
+
+	existingAudiences, _, err := unstructured.NestedStringSlice(existingConfig, audiencesPath...)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("unable to extract service account issuer from unstructured: %v", err))
+	}
+
+	// we're using the value of authentication.spec.issuer which is always a string, not a slice
+	if len(existingAudiences) > 0 {
+		existingAudience = existingAudiences[0]
+	}
+
+	authConfig, err := listers.AuthConfigLister().Get("cluster")
+	if apierrors.IsNotFound(err) {
+		klog.Warningf("authentications.config.openshift.io/cluster: not found")
+		// No issuer if the auth config is missing
+		authConfig = &configv1.Authentication{}
+	} else if err != nil {
+		return existingConfig, append(errs, err)
+	}
+
+	newAudience = authConfig.Spec.ServiceAccountIssuer
+	if len(newAudience) == 0 {
+		newAudience = "https://kubernetes.default.svc"
+	} else if err := checkIssuer(newAudience); err != nil {
+		return existingConfig, append(errs, err)
+	}
+
+	// set `issuerChanged` to emit an event about config change in defer
+	issuerChanged = existingAudience != newAudience
+
+	return map[string]interface{}{
+		"apiServerArguments": map[string]interface{}{
+			"api-audiences": []interface{}{
+				newAudience,
+			},
+		},
+	}, errs
+}
+
+// checkIssuer validates the issuer in the same way that it will be validated by
+// kube-apiserver
+func checkIssuer(issuer string) error {
+	if !strings.Contains(issuer, ":") {
+		return nil
+	}
+	// If containing a colon, must parse without error as a url
+	_, err := url.Parse(issuer)
+	if err != nil {
+		return fmt.Errorf("service-account issuer contained a ':' but was not a valid URL: %v", err)
+	}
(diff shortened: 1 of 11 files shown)


pkg/operator/configobservation/authentication/apiaudiences.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ObserveAPIAudiences changes apiServerArguments.api-audiences from
+// the default value if Authentication.Spec.ServiceAccountIssuer specifies a valid
+// non-empty value.
+func ObserveAPIAudiences(
+	genericListers configobserver.Listers,
+	recorder events.Recorder,
+	existingConfig map[string]interface{},
+) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
```
s-urbaniak, line 32: is this safe?

pkg/operator/configobservation/authentication/apiaudiences.go
```diff
@@ -0,0 +1,107 @@
...
+	// when the issuer will change, indicate that by setting `issuerChanged` to true
+	// to emit the informative event
+	defer func() {
+		if issuerChanged {
+			recorder.Eventf(
+				"ObserveAPIAudiences",
+				"service account issuer changed from %s to %s",
+				existingAudience, newAudience,
+			)
+		}
```
s-urbaniak, line 46: readability nit:
```
if !issuerChanged {
  return
}

recorder.Eventf...
```

pkg/operator/configobservation/authentication/apiaudiences.go
stlaz: yes, the programmer is responsible for it to be correct, it's used as such in a few other spots as well

pkg/operator/configobservation/authentication/apiaudiences.go
```diff
@@ -0,0 +1,107 @@
...
+			"api-audiences": []interface{}{
+				"https://kubernetes.default.svc",
+			},
+		},
+	}, errs
+}
+
+// checkIssuer validates the issuer in the same way that it will be validated by
+// kube-apiserver
+func checkIssuer(issuer string) error {
```
s-urbaniak, line 97: is this something we use elsewhere, potentially reusable logic?

stlaz: I think this is the only spot

pkg/operator/configobservation/authentication/apiaudiences.go
```diff
@@ -0,0 +1,112 @@
...
+		)
+	}()
+
+	existingAudiences, _, err := unstructured.NestedStringSlice(existingConfig, audiencesPath...)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("unable to extract service account issuer from unstructured: %v", err))
+	}
+
+	// we're using the value of authentication.spec.issuer which is always a string, not a slice
+	if len(existingAudiences) > 0 {
```
s-urbaniak, line 60: it's confusing that the configuration field is plural but only single-value in logic because it correlates with single value service account issuer.

does it make sense to emit some "quirky configuration" back to the user, i.e. setting a Status field with InvalidConfiguration=True?

s-urbaniak: i think we have the same logic in KAS-o though.

s-urbaniak: otoh this config should never be set by the user (sorry for thinking while commenting).

stlaz: This is an internal configuration that's not exposed to the user so I wouldn't worry about it too much. There would only be multiple values here if someone or something was touching something they aren't supposed to, and it would be rewritten right away.

pkg/operator/configobservation/authentication/apiaudiences.go
```diff
@@ -0,0 +1,112 @@
...
+		return map[string]interface{}{
+			"apiServerArguments": map[string]interface{}{
+				"api-audiences": []interface{}{
+					newAudience,
+				},
+			},
+		}, errs
+	}
+
+	issuerChanged = existingAudience != newAudience
```
s-urbaniak, line 90: nit: this can go before the `len(newAudience)` block as it is asserted in both cases.

pkg/operator/configobservation/authentication/apiaudiences.go
```diff
@@ -0,0 +1,112 @@
...
+		return existingConfig, append(errs, err)
+	}
+
+	newAudience = authConfig.Spec.ServiceAccountIssuer
+	if err := checkIssuer(newAudience); err != nil {
+		return existingConfig, append(errs, err)
+	}
+
+	if len(newAudience) != 0 {
+		issuerChanged = existingAudience != newAudience
```
s-urbaniak, line 79: this can be removed if it is moved before the if block, as it is asserted for both cases.

pkg/operator/configobservation/authentication/apiaudiences.go
```diff
@@ -0,0 +1,111 @@
...
+		}, errs
+	}
+
+	return map[string]interface{}{
+		"apiServerArguments": map[string]interface{}{
+			"api-audiences": []interface{}{
+				"https://kubernetes.default.svc",
+			},
+		},
+	}, errs
```
s-urbaniak, line 96: suggestion, for readability:
```
issuerChanged = existingAudience != newAudience
audience := "https://kubernetes.default.svc"
if len(newAudience) != 0 {
  audience = newAudience
}
return map[string]interface{}{
	"apiServerArguments": map[string]interface{}{
		"api-audiences": []interface{}{
			audience,
		},
	},
}, errs
```

stlaz: lgtm, I think I adhered to the form of the copy-paste but this definitely makes much more sense

stlaz:
/hold
too many events

stlaz:
/hold cancel
fixed

marun:
Is there a reason we need to support the audience not always being `[issuer]`? Our apiserver doesn't support setting the audiences, and the upstream default is setting the audiences to include only the issuer. I see code in ckao and now for the auth operator that has to worry about divergence between audience and issuer. Maybe our apiservers should not be configured with audiences explicitly and instead assume that the issuer and audience should match?

s-urbaniak:
> Maybe our apiservers should not be configured with audiences explicitly and instead assume that the issuer and audience should match?
good point. i guess this is to prevent also carry-patches, at least for kube-apiserver and here just for symmetry?

stlaz:
> Maybe our apiservers should not be configured with audiences explicitly and instead assume that the issuer and audience should match?

True, but that's very likely to only be the case for the CKAO, and that could work. We are actually keeping the audience==issuer but by setting the audience explicitly in KAS. We probably don't have to since that's the default behavior.

 The oauth-apiserver, however, does not know who the expected issuer of the tokens is. I _think_ there were trouble when the audiences in the TokenReview response were not returned from the oauth-apiserver, but we can investigate what happens when we don't set them at all - for both the cases of KAS and oauth-apiserver.

marun:
>The oauth-apiserver, however, does not know who the expected issuer of the tokens is. I think there were trouble when the audiences in the TokenReview response were not returned from the oauth-apiserver, but we can investigate what happens when we don't set them at all - for both the cases of KAS and oauth-apiserver.

I think I understand the kube-apiserver case, and I believe it would be preferable to stop having ckao explicitly set the audience to get the behavior we want without any code to have to maintain/test/get wrong (i.e. no code > code).

I'm confused, though, by the oauth-apiserver case. How does it use/consume/etc bound tokens? Doc pointers welcome if you have them.

PR #464, opened 2021-07-14
distribute oauth-server trust via a openshift-config-managed configmap

stlaz:
Since the oauth-server's serving certificates can now be customized, it's important to publish the serving certificate of the server. This certificate should be used by applications like the oauth-proxy and openshift-console in order for them to be capable of successfully complete the OAuth2 authorization process as these clients eventually get to talk directly to the oauth-server and therefore need to trust it.

/assign @slaskawi 
/cc @deads2k @jhadvig @spadgett

### DIFF
--- a/bindata/oauth-openshift/trust_distribution_rolebinding.yaml
+++ b/bindata/oauth-openshift/trust_distribution_rolebinding.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: system:openshift:oauth-servercert-trust
+  namespace: openshift-config-managed
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: system:openshift:oauth-servercert-trust
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+  name: system:authenticated
--- a/pkg/controllers/trustdistribution/trustdistribution_controller.go
+++ b/pkg/controllers/trustdistribution/trustdistribution_controller.go
@@ -1,0 +1,101 @@
+package trustdistribution
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+
+	configinformers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/common"
+)
+
+type trustDistributionController struct {
+	configMaps    corev1client.ConfigMapsGetter
+	secretsLister corev1listers.SecretLister
+
+	ingressLister configlistersv1.IngressLister
+}
+
+func NewTrustDistributionController(
+	cmClient corev1client.ConfigMapsGetter,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	ingressInformer configinformers.IngressInformer,
+	eventsRecorder events.Recorder,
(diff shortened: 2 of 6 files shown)


manifests/04_oauthserver_trust_role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
+  name: system:openshift:oauth-servercert-trust
```
slaskawi, line 6: I see the `system:openshift:operator:authentication` already binds to `cluster-admin`. So do we need this spacial file here? We already should have necessary permissions.

slaskawi (commented):
LGTM in general. Just one thing to clarify with the regards to RBAC.

manifests/04_oauthserver_trust_role.yaml
stlaz: this role is not for the authentication operator, but for everyone (see the rolebinding below)

slaskawi: For me it's fine. But please let someone more experienced have a look at this piece as well.

slaskawi (approved):
/lgtm

stlaz:
/hold
for further review

stlaz:
Added a more verbose error in the tests to indicate what's happening.

manifests/04_oauthserver_trust_role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
```
deads2k, line 5: This should not be needed. We should not honor a "false" here, so it doesn't matter.

manifests/04_oauthserver_trust_role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
```
deads2k, line 3: missing namespace?  The namespace here is known.

manifests/04_oauthserver_trust_role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
deads2k, line 1: location on this file is incorrect.  This is about access to something managed *by* the operator, not something managed *for* the operator.  This should be created by the operator, not by the CVO.  Please move to bindata and use the staticresourcecontroller to create and maintain this.

pkg/controllers/common/ingress_config.go
```diff
@@ -71,24 +71,29 @@ func GetCustomRouteHostname(ingress *configv1.Ingress, namespace string, name st
 // GetActiveRouterCertKeyBytes returns a byte array containing the server certificates, a byte array containing the private key,
 // a boolean representing if the default openshift-authentication/v4-0-config-system-router-certs secret is being used, and
 // any errors retrieving the active router secret.
-func GetActiveRouterCertKeyBytes(secretLister corev1listers.SecretLister, ingressConfig *configv1.Ingress, namespace string, defaultSecretName string, customSecretName string) ([]byte, []byte, bool, error) {
+func GetActiveRouterCertKeyBytes(
+	secretLister corev1listers.SecretLister,
+	ingressConfig *configv1.Ingress,
+	namespace string,
+	defaultSecretName string, customSecretName string,
+) (cert []byte, key []byte, isDefaultSecret bool, err error) {
```
deads2k, line 79: no named return values.

pkg/controllers/trustdistribution/trustdistribution_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	var serverCert *x509.Certificate
+	var errs []error
+	for block, rest := pem.Decode(certBundle); block != nil; block, rest = pem.Decode(rest) {
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		if !cert.IsCA {
+			// we found our server cert, get outta here
```
deads2k, line 79: seems like the entire chain of serving cert to root is valid for inclusion.

manifests/04_oauthserver_trust_role.yaml
stlaz: well spotted

bindata/oauth-openshift/trust_distribution_rolebinding.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: system:openshift:oauth-servercert-trust
+  namespace: openshift-config-managed
+  annotations:
```
deads2k, line 6: unnecessary now

stlaz:
/hold cancel
reapplying lgtm

PR #465, opened 2021-07-15
manifests: drop unnecessary histogram buckets

simonpasquier:
This change drops a few buckets from the
`apiserver_request_duration_seconds` histogram for the authentication
operator and the OAuth service. The dropped values are consistent with
what is already done for the OAuth API server. In practice we go from 38
buckets down to 13.

Signed-off-by: Simon Pasquier <spasquie@redhat.com>


stlaz:
/assign @s-urbaniak
/remove-lifecycle stale
Sergi, could you please have a look?

PR #469, opened 2021-07-28
Bug 1986829: metrics: use client cert auth for metrics scraping

stlaz:
/assign @s-urbaniak

### DIFF
--- a/manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
+++ b/manifests/0000_90_cluster-authentication-operator_02_servicemonitor.yaml
@@ -23,6 +23,8 @@ spec:
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
       serverName: metrics.openshift-authentication-operator.svc
+      certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+      keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
   jobLabel: component
   selector:
     matchLabels:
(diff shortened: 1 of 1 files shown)


stlaz:
> /lgtm
> @stlaz for the curious mind in me: do you mind to point me to the place where the client certs are actually respected in openshift-oauth-apiserver, oauth-openshift, and authentication-operator?

@s-urbaniak as for the authentication operator, the generic lib-go controller builder does it as such:
1. we build the server config ...
https://github.com/openshift/library-go/blob/14aa5b749d8ddbae509ea97c988228934dbbd3db/pkg/controller/controllercmd/builder.go#L250

2. and add the authn/z delegation options
https://github.com/openshift/library-go/blob/14aa5b749d8ddbae509ea97c988228934dbbd3db/pkg/config/serving/server.go#L39-L40

3. and then we go with the in-cluster config:
https://github.com/openshift/library-go/blob/14aa5b749d8ddbae509ea97c988228934dbbd3db/vendor/k8s.io/apiserver/pkg/server/options/authentication.go#L306-L320


---

As for the oauth-server and oauth-apiserver, the path is going to be quite similar for both, and actually resembles the above:
1. At first the generic API server config is constructed, and the authentication options are inserted in a generic kube manner:
https://github.com/openshift/oauth-server/blob/a51e18174ac32f0e44dfa8946e812d25ff38a101/pkg/cmd/oauth-server/server.go#L67-L69
2. which once again uses the in-cluster config:
https://github.com/openshift/oauth-server/blob/a51e18174ac32f0e44dfa8946e812d25ff38a101/vendor/k8s.io/apiserver/pkg/server/options/authentication.go#L306-L320
3. ... which we know is in-cluster since we're not in fact setting the ClientCA for the oauth-server explicitly:
https://github.com/enj/cluster-authentication-operator/blob/19e0cf0d50fd1d753a6023bee483cca5bbbd876b/pkg/controllers/payload/payload_config_controller.go#L170

PR #470, opened 2021-08-03
Bug 1988576: pkg/operator: Add deprecated stale status

s-urbaniak:
/cc @sttts


pkg/operator/starter.go
```diff
@@ -258,8 +258,11 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 		[]string{
 			// condition types removed in 4.8
 			"OAuthRouteCheckEndpointAccessibleControllerDegraded",
```
s-urbaniak, line 260: side note: i believe this can be removed due to https://github.com/openshift/cluster-authentication-operator/blob/release-4.7/pkg/controllers/endpointaccessible/endpoint_accessible_controller.go#L40, but better be on the safe side.

PR #478, opened 2021-09-01
Switch to CustomRouteController from library-go

slaskawi:
https://issues.redhat.com/browse/AUTH-14

This is a proof for https://github.com/openshift/library-go/pull/1138


slaskawi:
On my box, the `TestCustomRouterCerts` test is fine. Let's give them a few more tries...

/retest

pkg/operator/starter.go
```diff
@@ -9,6 +9,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/customroute"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
```
stlaz, line 13: order

stlaz (changes requested):

pkg/operator/starter.go
slaskawi: Fixed.

PR #479, opened 2021-09-03
Bug 1948089: bump(library-go): add shutdown detection during apiservice discovery

s-urbaniak:
This pulls in https://github.com/openshift/library-go/pull/1189
to prevent false positives of apiservice discovery during operator shutdown.

### DIFF
--- a/bindata/oauth-apiserver/oauth-apiserver-pdb.yaml
+++ b/bindata/oauth-apiserver/oauth-apiserver-pdb.yaml
@@ -1,0 +1,11 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: oauth-apiserver-pdb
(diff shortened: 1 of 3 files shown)


vendor/github.com/openshift/library-go/pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -120,6 +120,19 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
 
 	err = c.syncAPIServices(ctx, apiServices, syncCtx.Recorder())
 
+	if err != nil {
+		// a closed context indicates that the process has been requested to shutdown
```
p0lyn0mial, line 124: this is what we want :)

PR #481, opened 2021-09-03
Bug 1948080: bump(library-go): add shutdown detection during apiservice discovery

s-urbaniak:
This pulls in openshift/library-go#1189
to prevent false positives of apiservice discovery during operator shutdown.


s-urbaniak:
This depends on https://github.com/openshift/cluster-authentication-operator/pull/480 to land first

PR #476, opened 2021-08-31
bug  1998031: Deploy PDB to prevent more than one replica going unavailable

ingvagabund:
In case master nodes are drained to quickly in a row there's no
waiting until instances go back to available. So it's possible
for two or more replicas to go unavailable which may lead
to all instances going unavailable.

Based on https://github.com/openshift/library-go/pull/1056/commits/8efd1883d406cc389eb25e2a257c4451dfbd668c

PR #488, opened 2021-09-14
[4.8] bug 2003946: Deploy PDB to prevent more than one replica going unavailable

ingvagabund:
In case master nodes are drained to quickly in a row there's no
waiting until instances go back to available. So it's possible
for two or more replicas to go unavailable which may lead
to all instances going unavailable.

Backporting https://github.com/openshift/cluster-authentication-operator/pull/476


hardys:
[patch-manager] Approving for cherry-pick with a score of 0.50

PR #491, opened 2021-09-20
Bug 1958198: Avoid zero or low resync intervals

slaskawi:
https://bugzilla.redhat.com/show_bug.cgi?id=1958198


derekhiggins:
FYI e2e-agnostic-ipv6 is failing because the cloud providers credentials need to be updated, see https://github.com/openshift/release/pull/25749

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
```diff
@@ -51,7 +51,7 @@ func NewIngressNodesAvailableController(
 			nodeInformer.Informer(),
 		).
 		WithSync(controller.sync).
-		ResyncEvery(1*time.Minute).
+		ResyncEvery(time.Minute).
```
p0lyn0mial, line 54: would it make sense to provide some jitter? so that not all controllers try to reach the API at the same time?

slaskawi: Done.

PR #505, opened 2021-11-09
[WIP] enable audit

ibihim:
Looking into what happens when audit is enabled on `oauth-server`.

### DIFF
--- a/bindata/oauth-openshift/deployment.yaml
+++ b/bindata/oauth-openshift/deployment.yaml
@@ -102,6 +102,11 @@ spec:
             - name: v4-0-config-system-trusted-ca-bundle
               readOnly: true
               mountPath: /var/config/system/configmaps/v4-0-config-system-trusted-ca-bundle
+            - name: auditconfig
+              readOnly: true
+              mountPath: /var/config/system/configmaps/auditconfig
+            - name: tmpauditlog
+              mountPath: /tmp/audit
           readinessProbe:
             httpGet:
               path: /healthz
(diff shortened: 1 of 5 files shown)


bindata/oauth-openshift/deployment.yaml
```diff
@@ -102,6 +102,11 @@ spec:
             - name: v4-0-config-system-trusted-ca-bundle
               readOnly: true
               mountPath: /var/config/system/configmaps/v4-0-config-system-trusted-ca-bundle
+            - name: auditconfig
+              readOnly: true
+              mountPath: /var/config/system/configmaps/auditconfig
+            - name: tmpauditlog
```
EmilyM1, line 108: Do we want this in /tmp?

EmilyM1: There is this in must-gather that has a set location I just noticed
https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_audit_logs#L31

s-urbaniak: Currently, `gather_audit_logs` leverages the fact that logs are available in a host path. I suggest we simply keep that also for oauth-server.

I suggest keeping the same semantics as i.e. oauth-apiserver (just because it is a deployment as well) and do:
```
      - hostPath:
          path: /var/log/oauth-server
        name: audit-dir
...
      - mountPath: /var/log/oauth-server
        name: audit-dir
...
        exec oauth-server osinserver \
              --audit-log-path=/var/log/oauth-server/audit.log \
              --audit-log-format=json \
              --audit-log-maxsize=100 \
              --audit-log-maxbackup=10 \
```

ibihim: hahaha, oh my.. this is something that I wanted to suggest... I should've read your comment it would've eased up a day of thinking :smile:

PR #506, opened 2021-11-16
Bug 1982726: encryption-controller: sync secrets conditionally

dgrisonnet:
Update the version of library-go to include the latest changes that
allow the encryption controller to synchronize the encryption-config
secret only when the encryption controller should be run. This prevents
the secrets from being unnecessarily deleted in the apiserver namespace
whenever it doesn't exist in the config-managed namespace.

Follow-up to openshift/library-go#1243


p0lyn0mial:
/assing @stlaz 

for approval

p0lyn0mial:
we have similar PRs for kas-o (https://github.com/openshift/cluster-kube-apiserver-operator/pull/1255) and oas-o

s-urbaniak:
e2e failures seem unrelated

PR #508, opened 2021-11-18
test/lib: Set `PROXY_ADDRESS_FORWARDING` for keycloak

JAORMX:
As described in the docs [1] and the kubernetes example [2], the
aforementioned environment variable is needed for keycloak to properly
function behind a proxy (in our case, ingress).

[1] https://hub.docker.com/r/jboss/keycloak/
[2] https://raw.githubusercontent.com/keycloak/keycloak-quickstarts/latest/kubernetes-examples/keycloak.yaml

Signed-off-by: Juan Antonio Osorio Robles <jaosorior@redhat.com>


slaskawi (approved):
Makes perfect sense.

/lgtm

stlaz:
/approve
it would be quite nice if the docker docs also expained _why_ the variable needs to be set

PR #511, opened 2021-11-18
[WIP] AUTH-89: Audit logs as opts, yet another PR

ibihim:
## What

Start `oauth-server` with audit options.

## Why

Enable audit logging for `oauth-server` and authentication events.

### DIFF
--- a/bindata/oauth-openshift/deployment.yaml
+++ b/bindata/oauth-openshift/deployment.yaml
@@ -63,15 +63,21 @@ spec:
               fi
               exec oauth-server osinserver \
               --config=/var/config/system/configmaps/v4-0-config-system-cliconfig/v4-0-config-system-cliconfig \
-              --v=${LOG_LEVEL}
+              --v=5 \
+              ${AUDIT_OPTS}
           ports:
             - name: https
               containerPort: 6443
               protocol: TCP
           securityContext:
+            privileged: true
             readOnlyRootFilesystem: false # because of the `cp` in args
             runAsUser: 0 # because /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem is only writable by root
           volumeMounts:
+            - mountPath: /var/run/configmaps/audit
+              name: audit-policies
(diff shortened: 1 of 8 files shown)


ibihim:
@stlaz, do you have a clue, why the `audit-policy.yaml` doesn't create a `configmap`?

EmilyM1:
It may be possible that CAO is no longer creating static policies, but only creating dynamic policies

bindata/oauth-openshift/deployment.yaml
```diff
@@ -63,15 +63,25 @@ spec:
...
           securityContext:
+            privileged: true
             readOnlyRootFilesystem: false # because of the `cp` in args
             runAsUser: 0 # because /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem is only writable by root
           volumeMounts:
+            - mountPath: /var/run/configmaps/audit
+              name: audit-policies
+            - mountPath: /var/log/oauth-server
+              name: audit-dir
             - name: v4-0-config-system-session
```
EmilyM1, line 85: These are both names similar but are different audit logs:
https://github.com/openshift/cluster-authentication-operator/blob/master/bindata/oauth-apiserver/deploy.yaml#L101
Wonder if unique name would be best?

ibihim: This is done on purpose as we are writing audit logs to the node or am I wrong here?

pkg/operator/starter.go
```diff
@@ -260,6 +260,7 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 		"OpenshiftAuthenticationStaticResources",
 		assets.Asset,
 		[]string{
+			"oauth-openshift/audit-policy.yaml",
```
ibihim, line 263: I am not sure if I am missing something here, but it seems like the configmap is not there, when I push the CAO image into the cluster. Am I missing something?

@emilym1 mentioned that there are no static configmaps anymore?

PR #514, opened 2021-11-26
bug 2039670: Create the PDB dynamically so it does not get created on the SNO topology

ingvagabund:
Builds on top of https://github.com/openshift/library-go/pull/1246
Also need WithIsSNOCheck from https://github.com/openshift/library-go/pull/1238

/hold
- needs https://github.com/openshift/library-go/pull/1238 to merge
- needs https://github.com/openshift/library-go/pull/1266 to merge

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -53,6 +53,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/library-go/pkg/operator/revisioncontroller"
 	"github.com/openshift/library-go/pkg/operator/staleconditions"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/guard"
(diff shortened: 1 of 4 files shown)


pkg/operator/starter.go
```diff
@@ -547,15 +550,38 @@ func prepareOauthAPIServerOperator(ctx context.Context, controllerContext *contr
...
+				ShouldCreateFn: func() bool {
+					isSno, err := guard.IsSNOCheckFnc(infraLister)()
+					if err != nil {
+						return false
+					}
+					return !isSno
+				},
+				ShouldDeleteFn: func() bool {
+					isSno, err := guard.IsSNOCheckFnc(infraLister)()
+					if err != nil {
```
deads2k, line 577: I think you'll have a nicer time if you make the `IsSNOCheckFnc` only return a bool

PR #526, opened 2022-01-03
golang 1.16 bump: Updating ose-cluster-authentication-operator images to be consistent with ART

s-urbaniak:
supersedes https://github.com/openshift/cluster-authentication-operator/pull/455


stlaz:
added the labels the old PR had, the changes on top of the original are only specific to the CI machinery and do not affect the code itself.

PR #528, opened 2022-01-03
Bug 2030961: endpoints checker: check only the custom hostname if configured

stlaz:
The endpoints checker would use hostname from the first admitted ingress but it seems that the older ingresses may still appear in the route's status.

/assign @s-urbaniak 

cc @openshift/openshift-team-auth-maintainers

### DIFF
--- a/pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
+++ b/pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"strconv"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


s-urbaniak:
@stlaz can you add some tests which proofs the fixed behavior?

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -152,25 +154,49 @@ func listOAuthServices(serviceLister corev1listers.ServiceLister) ([]string, err
 	return toHealthzURL(results), nil
 }
 
-func listOAuthRoutes(routeLister routev1listers.RouteLister) ([]string, error) {
+func listOAuthRoutes(ingressLister configv1lister.IngressLister, routeLister routev1listers.RouteLister) ([]string, error) {
 	var results []string
+	ingressConfig, err := ingressLister.Get("cluster")
```
s-urbaniak, line 159: nit: s/ingressLister/ingressConfigLister

PR #533, opened 2022-01-14
bug 2040521: routeName used as customSecretName

ibihim:

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -293,8 +293,8 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 		operatorCtx.kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().ConfigMaps(),
 		"openshift-authentication",
 		"v4-0-config-system-router-certs",
-		"oauth-openshift",
 		"v4-0-config-system-custom-router-certs",
+		"oauth-openshift",
 	)
 
 	ingressStateController := ingressstate.NewIngressStateController(


ibihim:
I will take a look if I can add some tests on Monday.

slaskawi (approved):

s-urbaniak:
@ibihim can you please add some more context what this concretely fixes. I am wondering if this can be proven here with a unit test as well?

ibihim:
I will check if there is a reasonable way to provide tests for this as mentioned on Friday already.

It might be, that it is hard to protect against it to happen again from a testing perspective. The issue is that we have 4 string attributes, which are not the same logical type. Ideally we would introduce different types for `namespace`, `secretName` and `routeName`.

ibihim:
@s-urbaniak @stlaz, the only thing that I could improve here would be giving those strings a type. I will check the tests in origin, if I can fix it there.
If we don't want to do the type solution, can we merge this PR or do we need to have the tests in origin created / changed first?

ibihim:
> Cannot create more than 100 public IP addresses for this subscription in this region.

:sweat_smile:

ibihim:
1.  If the Bugzilla associated with the PR has the "FastFix" keyword, the subjective assessment on the issue has already been done and a customer is impacted. These PRs should be prioritized for merge.
    - [x] verified
    - [ ] does not apply
    - [ ] verified
    - [x] does not apply (workaround exists)
    - [ ] The bug is a critical fix, no reasonable workaround exists, and a recommendation for upgrade has been ruled out, or 
    - [ ] The bug is a security related bug
    - [x] Branch not in maintenance mode yet (current release + previous release for 90 days after current GA; everything older is in maintenance)
    - [x] verified
5. The PR was created with the cherry-pick bot OR the PR’s description is well formed with user-focused release notes that state the bug number, impact, cause, and resolution. Where appropriate, it should also contain information about how a user can identify whether a particular cluster is affected.
    - [x] verified

PR #535, opened 2022-01-17
Bug 2037274: starter.go: add invalidCertsController

s-urbaniak:
This includes a controller detecting invalid provider certificates as described in https://github.com/openshift/enhancements/pull/980.


s-urbaniak:
1.  If the Bugzilla associated with the PR has the "FastFix" keyword, the subjective assessment on the issue has already been done and a customer is impacted. These PRs should be prioritized for merge.
    - [x] verified
    - [ ] does not apply
    - [x] verified
    - [ ] does not apply
    - [x] The bug is a critical fix, no reasonable workaround exists, and a recommendation for upgrade has been ruled out, or 
    - [ ] The bug is a security related bug
    - [ ] Branch not in maintenance mode yet (current release + previous release for 90 days after current GA; everything older is in maintenance)
    - [x] verified
5. The PR was created with the cherry-pick bot OR the PR’s description is well formed with user-focused release notes that state the bug number, impact, cause, and resolution. Where appropriate, it should also contain information about how a user can identify whether a particular cluster is affected.
    - [x] verified

pkg/operator/starter.go
```diff
@@ -223,6 +224,17 @@ func prepareOauthOperator(controllerContext *controllercmd.ControllerContext, op
 
 	oauthInformers := oauthinformers.NewSharedInformerFactory(oauthClient, resync)
 
+	invalidCertsController := metricscontroller.NewMetricsController(
+		"InvalidProviderCertsController",
```
stlaz, line 228: Given that the only condition message we'll get is: `<controllerName>::InvalidCertsDetected: x server certificates without SAN detected`, should we maybe call this something else?

s-urbaniak: let's call it `InvalidCertsController`, similar to cluster-kube-apiserver-operator we might have more metrics in here.

s-urbaniak: the discussion results are manifested in https://github.com/openshift/library-go/pull/1289

PR #545, opened 2022-02-22
Bug 2052467: Custom route HTTPS certificate SAN validation

pierreprinetti:
This patch adds a new validation check for custom route HTTPS
certificates: certificates without SAN fields will prevent upgrades.

This change puts in force the long-standing deprecation of the CN field
as a provider for names.

After this patch, custom routes secured with legacy HTTPS certificates
will prevent the upgrade to OCP v4.10. OCP v4.10 is compiled with Go
v1.17 and is incompatible with such legacy certificates; this
deprecation prevents unexpected failures to occur on upgrade.

Addresses: Bug 2052467

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 all: build
 .PHONY: all
 
+export GODEBUG := x509ignoreCN=0
+
 # Include the library makefile
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
--- a/pkg/controllers/customroute/custom_route_conditions.go
+++ b/pkg/controllers/customroute/custom_route_conditions.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -13,6 +14,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/cluster-authentication-operator/pkg/controllers/common"
+	"github.com/openshift/cluster-authentication-operator/pkg/operator/datasync"
 	"github.com/openshift/library-go/pkg/route/routeapihelpers"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
@@ -31,6 +33,18 @@ func ensureDefaultConditions(conditions []metav1.Condition) []metav1.Condition {
 			})
 		}
 	}
+	for _, conditionType := range []string{"Upgradeable"} {
+		condition := findCondition(conditions, conditionType)
+		if condition == nil {
+			conditions = append(conditions, metav1.Condition{
+				LastTransitionTime: metav1.Now(),
+				Type:               conditionType,
+				Status:             metav1.ConditionTrue,
+				Reason:             "AsExpected",
+				Message:            "All is well",
+			})
+		}
+	}
 	return conditions
 }
 
@@ -43,27 +57,49 @@ func findCondition(conditions []metav1.Condition, conditionType string) *metav1.
 	return nil
 }
 
-func checkErrorsConfiguringCustomRoute(errors []error) []metav1.Condition {
-	if len(errors) != 0 {
-		now := metav1.Now()
-		return []metav1.Condition{
-			{
-				LastTransitionTime: now,
-				Type:               "Degraded",
-				Status:             metav1.ConditionTrue,
-				Reason:             "CustomRouteError",
-				Message:            fmt.Sprintf("Error Configuring custom route: %v", errors),
-			},
+func checkErrorsConfiguringCustomRoute(errs []error) []metav1.Condition {
+	now := metav1.Now()
+	var conditions []metav1.Condition
+	var degradedErrors, upgradeableErrors []error
+	for _, err := range errs {
+		var errNoSAN datasync.ErrNoSAN
+		if errors.As(err, &errNoSAN) {
+			upgradeableErrors = append(upgradeableErrors, err)
+		} else {
+			degradedErrors = append(degradedErrors, err)
+		}
+	}
+
+	if len(errs) != 0 {
+		conditions = []metav1.Condition{
 			{
 				LastTransitionTime: now,
 				Type:               "Progressing",
 				Status:             metav1.ConditionFalse,
 				Reason:             "CustomRouteError",
-				Message:            fmt.Sprintf("Error Configuring custom route: %v", errors),
+				Message:            fmt.Sprintf("Error Configuring custom route: %v", errs),
 			},
 		}
 	}
-	return nil
+	if len(degradedErrors) != 0 {
+		conditions = append(conditions, metav1.Condition{
+			LastTransitionTime: now,
+			Type:               "Degraded",
+			Status:             metav1.ConditionTrue,
+			Reason:             "CustomRouteError",
+			Message:            fmt.Sprintf("Error Configuring custom route: %v", degradedErrors),
+		})
+	}
+	if len(upgradeableErrors) != 0 {
+		conditions = append(conditions, metav1.Condition{
+			LastTransitionTime: now,
+			Type:               "Upgradeable",
+			Status:             metav1.ConditionFalse,
+			Reason:             "CustomRouteError",
+			Message:            fmt.Sprintf("Error Configuring custom route: %v", upgradeableErrors),
+		})
+	}
+	return conditions
 }
 
 func checkIngressURI(ingressConfig *configv1.Ingress, route *routev1.Route) []metav1.Condition {
--- a/pkg/operator/datasync/validation.go
+++ b/pkg/operator/datasync/validation.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/client-go/util/keyutil"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
 )
 
 var validators = map[string]func(data []byte) []error{
@@ -88,6 +89,11 @@ func ValidateServerCert(pem []byte) []error {
 		return append(errs, fmt.Errorf("expected at least one server certificate"))
 	}
 
+	for _, cert := range certs {
+		if !crypto.CertHasSAN(cert) {
+			errs = append(errs, newErrNoSAN(cert))
+		}
+	}
 	return errs
 }
 
--- a/pkg/operator/datasync/errors.go
+++ b/pkg/operator/datasync/errors.go
@@ -1,0 +1,22 @@
+package datasync
+
+import (
+	"crypto/x509"
+	"fmt"
+)
+
+type ErrNoSAN struct {
+	certSerialNumber string
+	certIssuer       string
+}
+
+func (err ErrNoSAN) Error() string {
+	return fmt.Sprintf("certificate relies on legacy Common Name field, use SANs instead:\n\tsn=%s;\n\tiss=%s", err.certSerialNumber, err.certIssuer)
+}
+
+func newErrNoSAN(cert *x509.Certificate) ErrNoSAN {
+	return ErrNoSAN{
+		certSerialNumber: cert.SerialNumber.String(),
+		certIssuer:       cert.Issuer.String(),
+	}
+}
--- a/pkg/operator/datasync/validation_test.go
+++ b/pkg/operator/datasync/validation_test.go
@@ -2,6 +2,7 @@ package datasync
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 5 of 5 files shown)


s-urbaniak:
@pierreprinetti thank you for the PR 👍 did you test this on a cluster-bot cluster? i.e. by configuring a faulty cert and then resetting it to a valid one?

pkg/controllers/routercerts/controller.go
```diff
@@ -207,6 +207,13 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
 		return newRouterCertsDegradedf("InvalidServerCertRouterCerts", "secret/%v.spec.data[%v] -n %v: certificate could not validate route hostname %v: %v", c.defaultSecretName, ingressDomain, c.secretNamespace, verifyOptions.DNSName, err)
 	}
 
+	// check the server certificates for reliance on the deprecated CN field
```
stlaz, line 210: This is a redundant change, `verifyWithAnyCertificate` should already return errors.

You should be looking at https://github.com/openshift/cluster-authentication-operator/blob/55b274c54c4a404799dea0e9e41f1b62e9f84c55/pkg/controllers/customroute/custom_route_controller.go

stlaz (changes requested):

pkg/controllers/routercerts/controller.go
pierreprinetti: Until Go v1.17, the `GODEBUG=x509ignoreCN=0` flag (which is embedded in all base images) will prevent `verifyWithAnyCertificate`, or any other certificate parsing, from failing due to missing SAN fields. This patch is targeted against OCP v4.9 which is compiled with Go v1.16.

pierreprinetti: That being said, `custom_route_controller` seems like a good tip. Do you think that the check should be implemented in both places? @stlaz

pierreprinetti: Oh and thank you for the review!

pierreprinetti: I have added the check in `ValidateServerCert`, which is the validation function called in the custom route controller loop. Thank you!

@stlaz PTAL

pierreprinetti:
> @pierreprinetti thank you for the PR +1 did you test this on a cluster-bot cluster? i.e. by configuring a faulty cert and then resetting it to a valid one?

Tested on AWS with cluster-bot, following the BZ steps. The certificate is rejected:

```
$ oc get ingress.config cluster -o yaml
[...]
status:
  componentRoutes:
  - conditions:
    - lastTransitionTime: "2022-02-23T18:22:02Z"
      message: "Error Configuring custom route: [error validating secret openshift-config/custom-auth-component:
        [certificate relies on legacy Common Name field, use SANs instead:\n\tsub=CN=auth-openshift-custom.jkfd.de;\n\tiss=CN=xxia_test_ca]]"
      reason: CustomRouteError
      status: "True"
      type: Degraded
    - lastTransitionTime: "2022-02-23T18:22:02Z"
      message: "Error Configuring custom route: [error validating secret openshift-config/custom-auth-component:
        [certificate relies on legacy Common Name field, use SANs instead:\n\tsub=CN=auth-openshift-custom.jkfd.de;\n\tiss=CN=xxia_test_ca]]"
      reason: CustomRouteError
      status: "False"
      type: Progressing
[...]
```

pkg/controllers/routercerts/controller.go
stlaz: I suppose we could have this check even here, although the description should say: "open a bug to the Routing component". cc @openshift/openshift-team-network-edge

pierreprinetti:
> https://test-route-e2e-test-authentication-operator-w98tj.apps.ci-op-bpmism0k-2959a.*********************************/auth/realms/master: 503 Service Unavailable

Still unsure if it's related.

/test e2e-operator

pkg/controllers/routercerts/controller.go
s-urbaniak: @stlaz i believe we had a discussion round with the router team. their assertion is that they don't provision the cert per se, so it is an exercise on the corresponding operators to do the validation.

pkg/controllers/routercerts/controller.go
```diff
@@ -207,6 +207,13 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
+	for _, serverCert := range serverCerts {
+		if !crypto.CertHasSAN(serverCert) {
+			return newRouterCertsDegradedf("InvalidServerCertRouterCerts", "secret/%v.spec.data[%v] -n %v (%v): certificate relies on legacy Common Name field, use SANs instead", c.defaultSecretName, ingressDomain, c.secretNamespace, serverCert.Subject.String())
```
stlaz, line 213: log the cert serial number instead of subject

pkg/operator/datasync/validation.go
```diff
@@ -88,6 +89,11 @@ func ValidateServerCert(pem []byte) []error {
 		return append(errs, fmt.Errorf("expected at least one server certificate"))
 	}
 
+	for _, cert := range certs {
+		if !crypto.CertHasSAN(cert) {
+			errs = append(errs, fmt.Errorf("certificate relies on legacy Common Name field, use SANs instead:\n\tsub=%s;\n\tiss=%s", cert.Subject, cert.Issuer))
```
stlaz, line 94: log cert serial number instead of subject

stlaz (changes requested):
Subject logging has long been disregarded as it might contain PII. Log certificate serial numbers instead where possible.

pkg/controllers/routercerts/controller.go
pierreprinetti: Thank you. I understand that this means that we keep both checks in place.

s-urbaniak: > we keep both checks in place.

agreed, hence SAN-less certs can happen both in the central route cert and in custom route certs and for both cases we should verify.

pierreprinetti:
I might need some help understanding if the test results are legit.
/retest-required

pierreprinetti:
/retest-required

ok let’s try one more time

pkg/controllers/routercerts/controller.go
stlaz: > @stlaz i believe we had a discussion round with the router team. their assertion is that they don't provision the cert per se, so it is an exercise on the corresponding operators to do the validation.

I believe they eventually decided to introduce the checks. If they did not check their inputs, our operators would go degraded.

pierreprinetti:
It looks like the CI fails the same tests on a pristine `release-4.9` branch: https://github.com/openshift/cluster-authentication-operator/pull/547

pierreprinetti:
For tests to be successful, #551 needs to be backported

pierreprinetti:
#555 merged! Tests have a chance of passing now.

/retest

pkg/controllers/routercerts/controller.go
```diff
@@ -207,6 +208,13 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
+	// check the server certificates for reliance on the deprecated CN field
+	for _, serverCert := range serverCerts {
+		if !crypto.CertHasSAN(serverCert) {
+			return newRouterCertsNoUpgradef("InvalidServerCertRouterCerts", "secret/%v.spec.data[%v] -n %v (%v): certificate relies on legacy Common Name field, use SANs instead", c.defaultSecretName, ingressDomain, c.secretNamespace, serverCert.SerialNumber)
+		}
+	}
+
 	// we made it this far without a problem
 	return operatorv1.OperatorCondition{
 		Type:   conditionRouterCertsDegradedType,
```
s-urbaniak, line 220: i guess this has to be `conditionRouterCertsNoUpgradeType`

pierreprinetti: OK that's going to be more complicated than that then. The `sync` loop currently assumes that we only operate one condition type in this function; operating `NoUpgrade` will require a little refactoring.

pierreprinetti: /hold

pierreprinetti:
Unit tests require https://github.com/openshift/release/pull/27417 to merge

pkg/controllers/routercerts/controller.go
```diff
@@ -194,26 +198,60 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
+	{
+		// check the server certificates for reliance on the deprecated CN field
+		var invalidCertificates []*x509.Certificate
+		for _, serverCert := range serverCerts {
+			if !crypto.CertHasSAN(serverCert) {
+				invalidCertificates = append(invalidCertificates, serverCert)
+			}
+		}
+		if len(invalidCertificates) > 0 {
+			message := fmt.Sprintf("the following certificates in 'secret/%v.spec.data[%v] -n %v' rely on legacy Common Name field, use SANs instead:", c.defaultSecretName, ingressDomain, c.secretNamespace)
```
stlaz, line 231: "the following are serial numbers of certificates in... that rely on..."

Makefile
```diff
@@ -1,6 +1,8 @@
 all: build
 .PHONY: all
 
+export GODEBUG := x509ignoreCN=0
```
stlaz, line 4: I would like to leave this as distribution-specific and have our upstream free of it

pkg/controllers/routercerts/controller_test.go
```diff
@@ -342,6 +391,20 @@ func withServer(host string, options ...func([]*cryptoMaterials) []*cryptoMateri
 	}
 }
 
+func withServerCN(host string, options ...func([]*cryptoMaterials) []*cryptoMaterials) func([]*cryptoMaterials) []*cryptoMaterials {
```
stlaz, line 394: `withServerNoSAN` would be a better name

stlaz (commented):
Remind me please, what happened to the custom route certificate validation?

I believe the router team in the end admitted they need to validate the certs this PR validates now, isn't that right?

pkg/controllers/routercerts/controller.go
```diff
@@ -194,26 +198,60 @@ func (c *routerCertsDomainValidationController) validateRouterCertificates() ope
...
 	// we made it this far without a problem
-	return operatorv1.OperatorCondition{
+	degradedCondition := operatorv1.OperatorCondition{
 		Type:   conditionRouterCertsDegradedType,
 		Status: operatorv1.ConditionFalse,
 		Reason: "AsExpected",
 	}
 
+	var upgradeCondition operatorv1.OperatorCondition
+	{
```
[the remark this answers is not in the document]
pierreprinetti: In my quest for readability, I isolate the code that determines the value to be assigned to `upgradeCondition`.

The variable `invalidCertificates`, for example, which is declared inside the block, is only instrumental to compute the value of `upgradeCondition`. With the scope block, I communicate the boundaries of `invalidCertificates`'s reason to exist.

Let me know if you prefer the code block to go away.

pkg/controllers/routercerts/controller.go
pierreprinetti: ok

pkg/controllers/routercerts/controller_test.go
pierreprinetti: ok; PTAL

pkg/controllers/routercerts/controller.go
pierreprinetti: ok; PTAL

pkg/controllers/routercerts/controller_test.go
```diff
@@ -39,21 +39,41 @@ func TestValidateRouterCertificates(t *testing.T) {
...
+			for _, condition := range status.Conditions {
+				if condition.Type == expectedType {
+					conditions = append(conditions, condition)
+				}
+			}
+
+			require.Len(t, conditions, 1)
+			condition := conditions[0]
+			require.Equal(t, expectedReason, condition.Reason, mergepatch.ToYAMLOrError(status))
+			require.Equal(t, expectedStatus, condition.Status, mergepatch.ToYAMLOrError(status))
```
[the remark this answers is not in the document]
pierreprinetti: ok! PTAL

pierreprinetti:
@stlaz 

>Remind me please, what happened to the custom route certificate validation?
>I believe the router team in the end admitted they need to validate the certs this PR validates now, isn't that right?

Apologies, please help me understand. Shall I put validation back in `ValidateServerCert` (`pkg/operator/datasync/validation.go`)? If so, should it raise Degraded or NoUpgrade?

pierreprinetti:
Changed the scope of this PR to only validate custom route server certificates.

@stlaz PTAL

pierreprinetti:
@stlaz ?

Makefile
stlaz: The art-team sets it in their builds.

pkg/controllers/customroute/custom_route_conditions.go
```diff
@@ -43,27 +57,46 @@ func findCondition(conditions []metav1.Condition, conditionType string) *metav1.
...
-				Status:             metav1.ConditionTrue,
-				Reason:             "CustomRouteError",
-				Message:            fmt.Sprintf("Error Configuring custom route: %v", errors),
-			},
+func checkErrorsConfiguringCustomRoute(errs []error) []metav1.Condition {
+	now := metav1.Now()
+	var conditions []metav1.Condition
+	var degradedErrors, upgradeableErrors []error
+	for _, err := range errs {
+		conditions = []metav1.Condition{
```
stlaz, line 65: you're rewriting this in each loop? I assume you might have intended to do something else?

stlaz (changes requested):

Makefile
pierreprinetti: The flag is [embedded in OCP base images since v4.6](https://github.com/openshift/images/commit/fbf868763eaefbea0654c27c82cd958147b19942). As a consequence, our code in this repository expects to run under it. Local runs (triggered with make) should run with the same flag.

If we don't add the flag, the unit tests will fail. And this is a good thing.

So either we add the flag in the Makefile, or at runtime in the test files. I think that here in the Makefile is better because it sets it for every test, thus making the tests behave a little more closely to what they'd do in production.

pkg/controllers/customroute/custom_route_conditions.go
pierreprinetti: that was a copy-paste mistake, thanks for catching.

pierreprinetti:
>The art-team sets it in their builds.

The flag is [embedded in OCP base images since v4.6](https://github.com/openshift/images/commit/fbf868763eaefbea0654c27c82cd958147b19942). As a consequence, our code in this repository expects to run under it. Local runs (triggered with make) should run with the same flag.

If we don't add the flag, the unit tests will fail. And this is a good thing.

So either we add the flag in the Makefile, or at runtime in the test files. I think that here in the Makefile is better because it sets it for every test, thus making the tests behave a little more closely to what they'd do in production.

pierreprinetti:
@stlaz Are you the one who can put the `backport-risk-assessed` label?

pierreprinetti:
1. If the Bugzilla associated with the PR has the "FastFix" keyword, the subjective assessment on the issue has already been done and a customer is impacted. These PRs should be prioritized for merge.
   
   * [ ]  verified
   * [x]  does not apply

   * [x]  verified
   * [ ]  does not apply

   * [x]  The bug is a critical fix, no reasonable workaround exists, and a recommendation for upgrade has been ruled out, or
   * [ ]  The bug is a security related bug

   * [x]  verified

5. The PR is merged in the next newer release branch and the bug for that newer release is VERIFIED by QE. The PR’s description is well formed with user-focused release notes that state the bug number, impact, cause, and resolution. Where appropriate, it should also contain information about how a user can identify whether a particular cluster is affected.
   
   * [ ]  verified
   * [x]  does not apply

PR #551, opened 2022-03-01
Bug 2059515: e2e: Pin the Keycloak version used in tests

pierreprinetti:
Before this patch, the e2e tests pulled the `latest` tag of
`quay.io/keycloak/keycloak`. The upgrade of the `latest` image to
Keycloack v17.0.0 made the e2e test suite fail.

More specifically, in the same conditions the 17.0.0 image [does not
start automatically][1] any more when the container is created.

This patch pins the Keycloak version to the "legacy" variant of v17.0.0,
which exposes the same behaviour as what the tests expect.

[1]: https://github.com/keycloak/keycloak-containers/commit/41baf7c224ca6206edb44d73252b44bb72ee31cc

### DIFF
--- a/test/library/keycloakidp.go
+++ b/test/library/keycloakidp.go
@@ -38,7 +38,7 @@ func AddKeycloakIDP(
 
 	nsName, keycloakHost, cleanup := deployPod(t, kubeClients, routeClient,
 		"keycloak",
-		"quay.io/keycloak/keycloak:latest",
+		"quay.io/keycloak/keycloak:17.0.0-legacy",
 		[]corev1.EnvVar{
 			// configure password for GitLab root user
 			{Name: "KEYCLOAK_USER", Value: "admin"},


pierreprinetti:
More specifically, the `17.0.0` image does not start automatically when the container is created.

Here I propose to pin the Keycloak version anyway and backport the change down to v4.7 to unblock any other patches. In a separate patch, we can instrument 4.11 to actually use Keycloak v17 (by passing the `start` or `start-dev` command).

One alternative can be to slap `start-dev` here and in the backports.

/retitle Bug 2059515: e2e: Pin the Keycloak version used in tests
/bugzilla refresh
/cc s-urbaniak

stlaz:
/hold
share:
1) issue in a Keycloak issue tracker that shows that the developers are aware of the problem. https://github.com/keycloak/keycloak-quickstarts/issues/300 is not enough because nobody has looked into it for 14 days. Contact the developers personally (we have the choice in RH, fortunately), to check they are aware of the issue and see when they are going to fix it if https://github.com/keycloak/keycloak-quickstarts/issues/300 is the only tracker for it

2) open a release-blocking BZ to move back to the latest image tag. We want to be testing with the latest Keycloak releases to make sure our functionality does not break with their new release

pierreprinetti:
@stlaz
Thank you for the swift review.

As mentioned [in the BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2059515#c2), I have already contacted the Keycloak devs. They mentioned that v17 introduced an API change and there's no intention to come back to the old behaviour (see the internal conversation [in the chat](https://chat.google.com/room/AAAAAJc6I9I/XYUJ_pbupRg)).

Since you don't want to pin the test dependency, I am going to change this PR to apply the alternative solution, which is: change the tests to use the new (v17, Quarkus-based) API.

My understanding is that you expect the API change to also be backported; that's what I am now going for.

pierreprinetti:
Based on our out-of-band conversation, I am reverting this patch to pin Keycloak to `17.0.0-legacy`. This patch will be backported.

I will also prepare a patch to bump the Keycloak image in `master` only (v4.11), and possibly unpin it again.

pierreprinetti:
/close
Superseded by #554

PR #554, opened 2022-03-03
Bug 2059515: e2e: Use Keycloak v17.0.0

pierreprinetti:
In v17, the Keycloak container image introduced several breaking changes
to the API[1], and most notably:
* a keycloak container won't start by default; it needs a `start` or
  `start-dev` command
* the `auth` part of the API path has been removed
* server configuration has changed

This patch fixes the e2e suite to test against the new Quarkus-based
Keycloak v17.0.0.

Before this patch, and because this repository is not pinning the
Keycloak version, e2e tests fail with:

    FAIL: TestKeycloakAsOIDCPasswordGrantCheckAndGroupSync

[1]: https://www.keycloak.org/migration/migrating-to-quarkus


test/library/keycloakidp.go
```diff
@@ -211,10 +219,18 @@ func (kc *keycloakClient) AuthenticatePassword(clientID, clientSecret, name, pas
 		return err
 	}
 
-	accessToken := authResp["access_token"].(string)
-	if len(accessToken) == 0 {
-		return fmt.Errorf("failed to retrieve an access token: %q - %q", authResp["error"], authResp["error_description"])
+	accessToken, ok := authResp["access_token"].(string)
+	if !ok || accessToken == "" {
```
stlaz, line 223: int comparison >> string comparison

keep `len(accessToken) == 0`

pierreprinetti: I find the string comparison more readable.

As of the compiled instructions, [there is no difference at all in Go 1.17](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:go,source:'package+main%0A%0Afunc+isEmpty1(s+string)+bool+%7B%0A++++return+s+%3D%3D+%22%22%0A%7D%0Afunc+isEmpty2(s+string)+bool+%7B%0A++++return+len(s)+%3D%3D+0%0A%7D%0A%0Afunc+main()+%7B%0A%7D%0A'),l:'5',n:'0',o:'Go+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:gl1170,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:go,libs:!(),options:'',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'amd64+gc+1.17+(Go,+Editor+%231,+Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

stlaz: I don't particularly care about the syntactic sugar and would rather stick with the explicit integer comparison, which is used all over this repository.

stlaz (changes requested):

PR #555, opened 2022-03-03
Bug 2060473: e2e: Pin Keycloack to the legacy variant

pierreprinetti:
In v17, the Keycloak container image introduced several breaking changes
to the API[1], and most notably:
* a keycloak container won't start by default; it needs a `start` or
  `start-dev` command
* the `auth` part of the API path has been removed
* server configuration has changed

This patch pins the test suite to the "legacy" variant of the Keycloak
17.0.0 image, which is backwards-compatible.

Before this patch, and because this repository was not pinning the
Keycloak version, e2e tests fail with:

    FAIL: TestKeycloakAsOIDCPasswordGrantCheck

[1]: https://www.keycloak.org/migration/migrating-to-quarkus

### DIFF
--- a/test/library/keycloakidp.go
+++ b/test/library/keycloakidp.go
@@ -38,7 +38,7 @@ func AddKeycloakIDP(
 
 	nsName, keycloakHost, cleanup := deployPod(t, kubeClients, routeClient,
 		"keycloak",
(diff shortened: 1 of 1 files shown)


stlaz:
this is only tests, I am going to override all the labels.

PR #561, opened 2022-03-25
make apiserver readiness probe honor readyz

deads2k:
This also uses a startup probe instead of a long initial delay seconds for livez and instead of excessive failure thresholds for readiness.  The failure thresholds were so long that the readiness check never failed.

/assign @tkashem @s-urbaniak

### DIFF
--- a/bindata/oauth-apiserver/deploy.yaml
+++ b/bindata/oauth-apiserver/deploy.yaml
@@ -71,7 +71,7 @@ spec:
               --etcd-cafile=/var/run/configmaps/etcd-serving-ca/ca-bundle.crt \
               --etcd-keyfile=/var/run/secrets/etcd-client/tls.key \
               --etcd-certfile=/var/run/secrets/etcd-client/tls.crt \
-              --shutdown-delay-duration=10s \
+              --shutdown-delay-duration=15s \
               --tls-private-key-file=/var/run/secrets/serving-cert/tls.key \
               --tls-cert-file=/var/run/secrets/serving-cert/tls.crt \
               ${FLAGS}
@@ -101,18 +101,36 @@ spec:
         - mountPath: /var/log/oauth-apiserver
           name: audit-dir
         livenessProbe:
-          initialDelaySeconds: 30
           httpGet:
             scheme: HTTPS
             port: 8443
             path: healthz
+          initialDelaySeconds: 0
+          periodSeconds: 10
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
         readinessProbe:
-          failureThreshold: 10
           httpGet:
             scheme: HTTPS
             port: 8443
             path: readyz
-      terminationGracePeriodSeconds: 70 # a bit more than the 60 seconds timeout of non-long-running requests
+          initialDelaySeconds: 0
+          periodSeconds: 5
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 1
+        startupProbe:
+          httpGet:
+            scheme: HTTPS
+            port: 8443
+            path: healthz
+          initialDelaySeconds: 0
+          periodSeconds: 5
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 30
+      terminationGracePeriodSeconds: 90 # a bit more than the 60 seconds timeout of non-long-running requests + the shutdown delay
       volumes:
       - name: audit-policies
         configMap:
--- a/pkg/operator/assets/bindata.go
+++ b/pkg/operator/assets/bindata.go
@@ -270,7 +270,7 @@ spec:
               --etcd-cafile=/var/run/configmaps/etcd-serving-ca/ca-bundle.crt \
               --etcd-keyfile=/var/run/secrets/etcd-client/tls.key \
               --etcd-certfile=/var/run/secrets/etcd-client/tls.crt \
-              --shutdown-delay-duration=10s \
+              --shutdown-delay-duration=15s \
               --tls-private-key-file=/var/run/secrets/serving-cert/tls.key \
               --tls-cert-file=/var/run/secrets/serving-cert/tls.crt \
               ${FLAGS}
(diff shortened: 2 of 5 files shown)


tkashem:
LGTM 

similar to https://github.com/kubernetes/kubernetes/issues/107727, and will give @s-urbaniak an opportunity to review as well.

bindata/oauth-apiserver/deploy.yaml
```diff
@@ -101,18 +101,36 @@ spec:
...
-          initialDelaySeconds: 30
           httpGet:
             scheme: HTTPS
             port: 8443
             path: healthz
+          initialDelaySeconds: 0
+          periodSeconds: 10
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
```
s-urbaniak, line 112: just confirm, these are the defaults as per https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes.

What is the rationale of going from 30s initial delay to 0?

bindata/oauth-apiserver/deploy.yaml
```diff
@@ -101,18 +101,36 @@ spec:
...
           httpGet:
             scheme: HTTPS
             port: 8443
             path: readyz
-      terminationGracePeriodSeconds: 70 # a bit more than the 60 seconds timeout of non-long-running requests
+          initialDelaySeconds: 0
+          periodSeconds: 5
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 1
```
s-urbaniak, line 122: Why is readiness probed for double as often (every 5s vs. 10s) compared to livenessProbe?

In the same realm, why are we just trying once (failureThreshold) before marking Unready vs. three times compared to liveness?

bindata/oauth-apiserver/deploy.yaml
```diff
@@ -101,18 +101,36 @@ spec:
...
+          periodSeconds: 10
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
         readinessProbe:
-          failureThreshold: 10
           httpGet:
             scheme: HTTPS
             port: 8443
             path: readyz
```
s-urbaniak, line 117: The PR title says "make apiserver readiness probe honor readyz", but it does honor readyz already, no? or did you mean to say "make apiserver startupProbe honor readyz"?

bindata/oauth-apiserver/deploy.yaml
```diff
@@ -101,18 +101,36 @@ spec:
...
+          httpGet:
+            scheme: HTTPS
+            port: 8443
+            path: healthz
+          initialDelaySeconds: 0
+          periodSeconds: 5
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 30
+      terminationGracePeriodSeconds: 90 # a bit more than the 60 seconds timeout of non-long-running requests + the shutdown delay
```
s-urbaniak, line 133: where does the referenced 60 seconds come from?

bindata/oauth-apiserver/deploy.yaml
```diff
@@ -101,18 +101,36 @@ spec:
...
+        startupProbe:
+          httpGet:
+            scheme: HTTPS
+            port: 8443
+            path: healthz
+          initialDelaySeconds: 0
+          periodSeconds: 5
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 30
```
s-urbaniak, line 132: same questions as above: how did we come up with the custom thresholds for `periodSeconds` and `failureThreshold`?

s-urbaniak: so, effectively this means: "tolerate 30 failures during startup, and up to 3 failures once started before killing"

bindata/oauth-apiserver/deploy.yaml
deads2k: > What is the rationale of going from 30s initial delay to 0?

We're using a startup probe now, which allows for a different set of settings before passing the startup, then using a tight value after startup is complete.

deads2k: > just confirm, these are the defaults as per https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes.

Many are.  Abu thought it was easier to reason about with all the values listed.  I think he was right.

bindata/oauth-apiserver/deploy.yaml
deads2k: > The PR title says "make apiserver readiness probe honor readyz", but it does honor readyz already, no? or did you mean to say "make apiserver startupProbe honor readyz"?

It did not honor it during shutdown because 10 failures times 10 seconds between attempts, means that ready would not be set to false before the graceful termination was already over.  So the pod disappeared, it didn't go ready=false.  This tightens it considerably.

bindata/oauth-apiserver/deploy.yaml
deads2k: > Why is readiness probed for double as often (every 5s vs. 10s) compared to livenessProbe?

liveness should basically never go false, but readyz should go false every time a shutdown starts.  So we don't really need to check liveness often, but readiness we want to responsive to.

> In the same realm, why are we just trying once (failureThreshold) before marking Unready vs. three times compared to liveness?

if for some reason this pod cannot answer requests (ready=false), we want it out of the service LB quickly and it can re-enter.  But if liveness fails, we're going to kill the container, so we want to give it a little more grace.   The stakes are higher.

bindata/oauth-apiserver/deploy.yaml
deads2k: > so, effectively this means: "tolerate 30 failures during startup

Yes, this allows a long startup.  Chosen based on a comment somewhere from sttts saying it took 10 seconds per API and 100 seconds total.

> up to 3 failures once started before killing"

Once started, it uses liveness checks

> same questions as above: how did we come up with the custom thresholds for periodSeconds and failureThreshold?

it roughly matches the long timeouts from before that were slowing down responsiveness overall.

bindata/oauth-apiserver/deploy.yaml
deads2k: > where does the referenced 60 seconds come from?

hardcoded (I think) into the apiserver for shutting down long running requests .

s-urbaniak:
@deads2k thank you for the detailed answers, this clears a lot of questions up! 🙇

PR #571, opened 2022-05-13
Updating ose-cluster-authentication-operator images to be consistent with ART

s-urbaniak:
supersedes #543

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.17-openshift-4.10
+  tag: rhel-8-release-golang-1.18-openshift-4.11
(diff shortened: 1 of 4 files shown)


_output/named-golang-versions
```diff
@@ -0,0 +1,3 @@
+.ci-operator.yaml: 1.18
```
stlaz, line 1: eh, remove the `_output`

stlaz (changes requested):
fix https://github.com/openshift/cluster-authentication-operator/blob/master/.ci-operator.yaml

edit: I see it's fixed, I accidentally submitted my review too early

_output/named-golang-versions
s-urbaniak: lol, good catch

s-urbaniak: done

PR #573, opened 2022-06-29
Bug 2078287: only ever include certificates in the oauth-serving-cert CM

stlaz:
/assign @ibihim


stlaz:
as advised by @jmguzik, adding the labels for the inaccessible BZ manually

PR #576, opened 2022-07-06
Bug 2111842: v/o/library-go - version-bump

ibihim:
## What

Version bump library-go.

## Why

library-go contains changes that avoid a state where an unexpected state
ends up in P=f, A=t and D=f. It stays no in P=t.

## Notes

IsNOCheckFnc got moved into a different package and renamed.

## Reference

[Flaky Request Header IDP tests](https://bugzilla.redhat.com/show_bug.cgi?id=2065028)
[Clone BZ for tracking](https://bugzilla.redhat.com/show_bug.cgi?id=2111842)


pkg/operator/starter.go
```diff
@@ -53,7 +53,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/library-go/pkg/operator/revisioncontroller"
 	"github.com/openshift/library-go/pkg/operator/staleconditions"
-	"github.com/openshift/library-go/pkg/operator/staticpod/controller/guard"
+	staticpodcommon "github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
```
stlaz, line 56: move non-generated changes into a separate commit

stlaz (changes requested):

pkg/operator/starter.go
ibihim: The version bump broke the system and I needed to fix it. So I should create a commit that contains the update (even though it isn't working in itself) and a fix for the update in a separate commit?

stlaz: yes

PR #578, opened 2022-07-14
[4.9] Bug 2107027: only ever include certificates in the oauth-serving-cert CM

stlaz:


y4sht:
Steps to verify: https://bugzilla.redhat.com/show_bug.cgi?id=2107027#c2

PR #581, opened 2022-08-15
e2e: address PSa failures

stlaz:
PSa now attempts to autosync enforce labels but since we're directly applying pod with admin kubeconfig, the synchronization has no way of telling that this is allowed.

Specifically mark the IDP namespace as privileged.

/assign @ibihim


ibihim (approved):
/lgtm

PR #589, opened 2022-11-09
OCPBUGS-3440: Restart authentication operator if console capability is enabled

bparees:


manifests/07_deployment.yaml
```diff
@@ -35,7 +35,8 @@ spec:
               echo "Copying system trust bundle"
               cp -f /var/run/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
-          exec authentication-operator operator --config=/var/run/configmaps/config/operator-config.yaml --v=2 --terminate-on-files=/var/run/configmaps/trusted-ca-bundle/ca-bundle.crt
+          touch /tmp/terminate
```
stlaz, line 38: The observer should be capable of watching even just a creation of the file, we shouldn't need to create it.

The termination would then happen simply by performing some sort of "touch" from Go.

pkg/controllers/configobservation/clusterversion/observe_clustercapabilities.go
```diff
@@ -0,0 +1,43 @@
...
+	clusterVersionConfig, err := listers.ClusterVersionLister.Get("version")
+	if err != nil {
+		return existingConfig, append(errs, err)
+	}
+
+	for _, capability := range clusterVersionConfig.Status.Capabilities.EnabledCapabilities {
+		if capability == configv1.ClusterVersionCapabilityConsole {
+			recorder.Eventf("ObserveClusterCapabilities", "Console capability enabled, restarting cluster authentication operator")
+			klog.Infof("Console capability enabled, restarting cluster authentication operator")
+			f, err := os.OpenFile("/tmp/terminate", os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
```
stlaz, line 29: That's an interesting pattern. Basically we're working around being able to `cancel()` the global context from the inside of the operator :thinking:

bparees: yeah, the global context/cancel function created in the library isn't exposed to the operator itself so i couldn't just cancel it, at least i didn't see a way...i'd certainly have preferred to have just done that.  But i also didn't want to go refactoring openshift-library to make it possible.

pkg/controllers/configobservation/clusterversion/observe_clustercapabilities.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/klog/v2"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/configobservation"
+)
+
+func ObserveClusterCapabilities(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, _ []error) {
```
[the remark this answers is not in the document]
bparees: redone and pushed

pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -94,11 +93,12 @@ func NewConfigObserver(
 		PreRunCachesSynced:   preRunCacheSynced,
 	}
 
-	// Check if the Console capability is enabled on the cluster and sync and add its informer and lister.
+	// Check if the Console capability is enabled on the cluster and sync and add its informer, lister, and config observer
 	if enabledClusterCapabilities.Has("Console") {
 		listers.PreRunCachesSynced = append(listers.PreRunCachesSynced, configInformer.Config().V1().Consoles().Informer().HasSynced)
 		informers = append(informers, configInformer.Config().V1().Consoles().Informer())
 		listers.ConsoleLister = configInformer.Config().V1().Consoles().Lister()
+		oauthServerObservers = append(oauthServerObservers, configobserver.WithPrefix(console.ObserveConsoleURL, configobservation.OAuthServerConfigPrefix))
```
bparees, line 101: this wasn't strictly needed but there's no reason to run this observer as long as the console isn't enabled.  (the observer itself no-ops if the console cap is disabled)

pkg/controllers/termination/termination_controller.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+type terminationController struct {
+	clusterVersionLister configlistersv1.ClusterVersionLister
+	recorder             events.Recorder
+}
+
+func NewTerminationController(configInformer configinformers.SharedInformerFactory, recorder events.Recorder) factory.Controller {
```
stlaz, line 23: godoc please

manifests/07_deployment.yaml
bparees: done

pkg/controllers/configobservation/console/observe_consoleurl_test.go
```diff
@@ -120,7 +126,7 @@ func TestObserveConsoleURL(t *testing.T) {
 			}
 
 			for i := range errs {
-				if strings.Contains(tt.expectedErrs[i], errs[i].Error()) {
+				if !strings.Contains(errs[i].Error(), tt.expectedErrs[i]) {
```
bparees, line 129: this logic was wrong in two ways.

as the tests are written, the expected error string is a subset of the actual error string.

and of course the test should fail if an error occurs that was not in the expected list, not fail when an expected error occurs.

PR #599, opened 2023-03-06
OCPBUGS-8372: manifests: create target namespace earlier

vrutkovs:
In order to speed up initial install target namespace definitions need to be included in the payload, as these need to have SCCs assigned and its benefitial to do this earlier on startup


manifests/0000_10_cluster-authentication-operator_01_operand-namespace.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-oauth-apiserver
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
```
stlaz, line 10: /hold
CVO does not maintain this NS. This should be "create-only" at best

stlaz (changes requested):

PR #608, opened 2023-04-20
AUTH-363: make proxy config check less obtrusive

liouk:
This PR contains the following changes:
- increase check interval to 60 min
- perform the actual endpoint calls with/without proxy only when they are required to deduce the check outcome
- filter configmap events to only the ones used in the context of the proxy config checker

### DIFF
--- a/pkg/controllers/proxyconfig/proxyconfig_controller.go
+++ b/pkg/controllers/proxyconfig/proxyconfig_controller.go
@@ -56,8 +56,9 @@ func NewProxyConfigChecker(
 		ResyncEvery(60 * time.Minute).
 		WithSyncDegradedOnError(operatorClient)
 
-	for ns := range caConfigMaps {
-		c.WithInformers(
+	for ns, configMapNames := range caConfigMaps {
+		c.WithFilteredEventsInformers(
+			factory.NamesFilter(configMapNames...),
 			configMapInformers.InformersFor(ns).Core().V1().ConfigMaps().Informer(),
 		)
 	}
--- a/pkg/controllers/proxyconfig/proxyconfig_controller.go
+++ b/pkg/controllers/proxyconfig/proxyconfig_controller.go
@@ -52,7 +52,7 @@ func NewProxyConfigChecker(
 		WithInformers(
 			routeInformer.Informer(),
 		).
-		ResyncEvery(5 * time.Minute).
+		ResyncEvery(60 * time.Minute).
 		WithSyncDegradedOnError(operatorClient)
 
 	for ns := range caConfigMaps {
--- a/pkg/controllers/proxyconfig/proxyconfig_controller.go
+++ b/pkg/controllers/proxyconfig/proxyconfig_controller.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"sync"
 	"time"
 
 	"golang.org/x/net/http/httpproxy"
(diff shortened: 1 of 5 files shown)


pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -52,7 +52,7 @@ func NewProxyConfigChecker(
 		WithInformers(
 			routeInformer.Informer(),
 		).
-		ResyncEvery(5 * time.Minute).
+		ResyncEvery(60 * time.Minute).
 		WithSyncDegradedOnError(operatorClient)
 
 	for ns := range caConfigMaps {
```
stlaz, line 58: This `for` could be quite problematic as it registers event handlers for any CM in a namespace, and so even with a more lean resync, you might fire the sync loop much more often.

Use `c.WithFilteredEventsInformers()` and base the filter off of the CM names that you store in `caConfigMaps` already

stlaz (changes requested):

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -94,18 +94,24 @@ func (p *proxyConfigChecker) sync(ctx context.Context, _ factory.SyncContext) er
 // checkProxyConfig determines any mis-configuration in proxy settings by attempting
 // to connect to endpoint directly and via proxy and comparing the results with expectations.
 func checkProxyConfig(ctx context.Context, endpointURL *url.URL, noProxy string, clientWithProxy, clientWithoutProxy *http.Client) error {
-	withProxyErr := isEndpointReachable(ctx, endpointURL.String(), clientWithProxy)
-	withoutProxyErr := isEndpointReachable(ctx, endpointURL.String(), clientWithoutProxy)
```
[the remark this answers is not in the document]
liouk: This wouldn't be enough though; we're using these values more than once so we would anyway have to call the wrapper func and store the return value as is done now with `isEndpointReachable`, so we'd end up with similar code anyway.

What we could do is use the closures to store the result and only call `isEndpointReachable` once, for example:

```golang
var withProxyErr *error
withProxy := func() *error {
  if withProxyErr == nil {
    err := isEndpointReachable(ctx, endpointURL.String(), clientWithProxy)
    withProxyErr = &err
  }

  return withProxyErr
}
```

Thus whenever we use `withProxy()` it'll either call `isEndpointReachable` or return the existing value in `withProxyErr`. This becomes less readable though IMHO :/ Thoughts?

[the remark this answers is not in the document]
stlaz: ah, I was about to propose a lazy evaluator originally but I thought we only need each of these once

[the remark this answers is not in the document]
liouk: added a proper implementation of the lazy evaluation of the endpoint checks

pkg/controllers/proxyconfig/proxyconfig_controller.go
liouk: This makes sense, thanks for the suggestion. I'm not very familiar with informers and event handlers yet -- is this what you meant? :point_right: https://github.com/openshift/cluster-authentication-operator/pull/608/files#diff-067c8061203aef85d6b9fcc6819fda8a9cc8a490a46f38973e0d7e40e0b1c5c7R60

On a side note, I noticed that there's a [TODO](https://github.com/liouk/cluster-authentication-operator/blob/proxyconfig-check-freq/pkg/controllers/common/config.go#L34) on `common.NamesFilter` that it should be in library-go; and in fact it exists in [`library-go/pkg/controller/factory/eventfilters.go`](https://github.com/openshift/library-go/blob/master/pkg/controller/factory/eventfilters.go#L17) -- should I use that one instead?

If yes, there are a couple of other usages of `common.NamesFilter` that we could replace with `factory.NamesFilter` ([here](https://github.com/liouk/cluster-authentication-operator/blob/proxyconfig-check-freq/pkg/controllers/routercerts/controller.go#L82) and [here](https://github.com/liouk/cluster-authentication-operator/blob/proxyconfig-check-freq/pkg/controllers/oauthclientscontroller/oauthclientscontroller.go#L62-L69)), what do you think?

stlaz: > This makes sense, thanks for the suggestion. I'm not very familiar with informers and event handlers yet -- is this what you meant? point_right https://github.com/openshift/cluster-authentication-operator/pull/608/files#diff-067c8061203aef85d6b9fcc6819fda8a9cc8a490a46f38973e0d7e40e0b1c5c7R60

Yes, that looks like it could be used here, too.

> On a side note, I noticed that there's a [TODO](https://github.com/liouk/cluster-authentication-operator/blob/proxyconfig-check-freq/pkg/controllers/common/config.go#L34) on common.NamesFilter that it should be in library-go; and in fact it exists in [library-go/pkg/controller/factory/eventfilters.go](https://github.com/openshift/library-go/blob/master/pkg/controller/factory/eventfilters.go#L17) -- should I use that one instead?

yes, use that instead, feel free to replace all the other use of the names filter in this repo with it

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -52,11 +54,12 @@ func NewProxyConfigChecker(
 		WithInformers(
 			routeInformer.Informer(),
 		).
-		ResyncEvery(5 * time.Minute).
+		ResyncEvery(60 * time.Minute).
 		WithSyncDegradedOnError(operatorClient)
 
-	for ns := range caConfigMaps {
-		c.WithInformers(
```
stlaz, line 59: separate these changes to the first commit, the lazy checking goes to a second one

stlaz (changes requested):

pkg/controllers/proxyconfig/proxyconfig_controller.go
liouk: done in a separate commit for the other references

pkg/controllers/proxyconfig/proxyconfig_controller.go
liouk: good idea, done

dgoodwin:
/label jira/valid-bug

TRT picking this up from the merge queue.

PR #611, opened 2023-05-17
OCPBUGS-12907: pkg/operator/starter: add InvalidProviderInvalidCertsUpgradeable to stale conditions

s-urbaniak:
This removes the `InvalidProviderInvalidCertsUpgradeable` condition present in 4.9. This condition is not present in 4.10 any more as the controller setting it is removed.

The underlying reason is that invalid certs were tolerated in 4.9, however in 4.10 they won't work at all any more because of the Go version being used, see https://bugzilla.redhat.com/show_bug.cgi?id=2037274#c23.

xref https://github.com/openshift/cluster-authentication-operator/pull/610


s-urbaniak:
closing out as fixing this can be done manually and the criticality is not high enough.

PR #613, opened 2023-06-06
OCPBUGS-2765: Library go bump

ibihim:
# What

Bumping library-go.
Bumping library-go's necessary deps.

# Why

Contains updated audit policy.


stlaz:
/hold
explain what the bump's for
go.sum update does not belong to "fixes after bump"
split the golang bump to 1.20 to a separate commit

PR #603, opened 2023-03-20
OCPBUGS-10568: migrate to using lease objects for leader election

tkashem:

PR #615, opened 2023-06-19
OCPBUGS-10568: migrate to using lease objects for leader election

dgrisonnet:
Wrapping up the work @tkashem started in https://github.com/openshift/cluster-authentication-operator/pull/603.


dgrisonnet:
This ended up being fixed as a side effect of https://github.com/openshift/cluster-authentication-operator/pull/613.

/close

PR #629, opened 2023-08-25
OCPBUGS-17987: generate only >=0 number of random bits

liouk:
This PR changes the input type of the `randomBits()` func to `uint` so that it won't accept any negative numbers as input (which cause a panic).


stlaz:
/approve
/hold cancel
the automation should be capable of merging this when the critical-fixes-only gate is lifted.

PR #631, opened 2023-08-25
[poc] only issue endpoint GETs if time has expired or connection info has changed

deads2k:
/assign illias

for  illustration only, I don't plan to fix up tests


deads2k:
this was a POC for something.  I think @liouk picked it up.

PR #646, opened 2024-01-11
WRKLDS-1004: use AlwaysAllow UnhealthyPodEvictionPolicy in PDBs

atiratree:
allow eviction of unhealthy (not ready) pods even if there are no disruptions allowed on a PodDisruptionBudget. This can help to drain/maintain a node and recover without a manual intervention when multiple instances of nodes or pods are misbehaving. Use this with caution, as this option can disrupt perspective pods that have not yet had a chance to become healthy.


atiratree:
similar to what was done for kube-apiserver: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1579

PR #656, opened 2024-02-20
AUTH-482: set required-scc for openshift workloads

liouk:


liouk:
Interestingly, the `oauth-openshift` pods were getting admitted with the `node-exporter` SCC (belongs to the [cluster-monitoring-operator](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/node-exporter/security-context-constraints.yaml)) exactly because of the problem that AUTH-482 describes.

liouk:
/hold
need to verify SCC

manifests/07_deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
         app: authentication-operator
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: anyiud
```
ibihim, line 25: `anyuid` is sufficient? Interesting. I've seen that in the cluster as well.

ibihim:
Could be a lgtm, but the tests look failed... 🤔

stlaz (changes requested):

pkg/operator/assets/bindata.go
```diff
@@ -230,6 +230,7 @@ spec:
         apiserver: "true"
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: privileged
```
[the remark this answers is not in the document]
liouk: Done in 76183e3db5ef956316fff6280e0c2c69e1f0a5b9

bindata/assets.go
```diff
@@ -0,0 +1,24 @@
+package bindata
+
+import (
+	"embed"
+)
+
+//go:embed oauth-apiserver/* oauth-apiserver/RBAC/* oauth-openshift/*
```
stlaz, line 7: //go:embed oauth-apiserver/* oauth-openshift/*

stlaz:
https://github.com/openshift/cluster-authentication-operator/blob/6214f1b9fc57d14d242eb4af8c31da727a61582a/pkg/dependencymagnet/dependencymagnet.go#L9
this dep magnet also seems to be ready to be removed

liouk:
debug commits dropped
/hold cancel

PR #657, opened 2024-02-28
NO-ISSUE: oauth-apiserver: enable shutdown-send-retry-after 

p0lyn0mial:
if true the HTTP Server will continue listening until all non long running request(s) in flight have been drained,
during this window all incoming requests will be rejected with a status code 429 and a 'Retry-After' response header,
in addition 'Connection: close' response header is set in order to tear down the TCP connection when idle.

without this flag requests that arrive after the shutdown delay duration will get a network error.

which might explain failures seen in: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-ci-4.15-e2e-azure-sdn-upgrade/1759779765541146624

The other issue is that `TerminationMinimalShutdownDurationFinished` says `The minimal shutdown duration of 0s finished` which means that the server can be immediately terminated when all in-flight requests are completed.

### DIFF
--- a/bindata/oauth-apiserver/deploy.yaml
+++ b/bindata/oauth-apiserver/deploy.yaml
@@ -72,6 +72,7 @@ spec:
               --etcd-keyfile=/var/run/secrets/etcd-client/tls.key \
               --etcd-certfile=/var/run/secrets/etcd-client/tls.crt \
               --shutdown-delay-duration=15s \
+              --shutdown-send-retry-after=true \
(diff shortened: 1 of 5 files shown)


p0lyn0mial:
https://github.com/openshift/oauth-apiserver/pull/105 - addresses the second issue.

p0lyn0mial:
I have just realised that without https://github.com/openshift/oauth-apiserver/pull/105 the `shutdown-send-retry-after` flag won't be applied as well.

PR #664, opened 2024-04-09
OCPBUGS-32089: wellknown-readiness: perform several attempts to connect before going unavailable

stlaz:
cc @openshift/openshift-team-auth

### DIFF
--- a/pkg/controllers/readiness/wellknown_ready_controller.go
+++ b/pkg/controllers/readiness/wellknown_ready_controller.go
@@ -16,7 +16,6 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
-	routev1 "github.com/openshift/api/route/v1"
 	configinformer "github.com/openshift/client-go/config/informers/externalversions"
 	configv1lister "github.com/openshift/client-go/config/listers/config/v1"
 	routeinformer "github.com/openshift/client-go/route/informers/externalversions/route/v1"
@@ -115,7 +114,8 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
 		}
 	}()
 
-	route, err := c.routeLister.Routes("openshift-authentication").Get("oauth-openshift")
+	// the well-known endpoint cannot be ready until we know the oauth-server's hostname
+	_, err = c.routeLister.Routes("openshift-authentication").Get("oauth-openshift")
 	if apierrors.IsNotFound(err) {
 		statusUpdates = append(statusUpdates, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
 			Type:    "WellKnownAvailable",
@@ -128,7 +128,7 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
 		return err
 	}
 
-	if err := c.isWellknownEndpointsReady(operatorSpec, operatorStatus, authConfig, route, infraConfig); err != nil {
+	if err := c.isWellknownEndpointsReady(ctx, operatorSpec, operatorStatus, authConfig, infraConfig); err != nil {
 		statusUpdates = append(statusUpdates, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
 			Type:    "WellKnownAvailable",
 			Status:  operatorv1.ConditionFalse,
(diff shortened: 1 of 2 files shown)


wking:
Checking [the update run][1], the intervals chart shows auth blipping `Available=False` late in the update during the machine-config node-reboot phase, when the incoming operator with your patch should be running:

image

The associated test-case has:

```
: [bz-apiserver-auth] clusteroperator/authentication should not change condition/Available expand_less
Run #0: Failed expand_less	1h31m45s
{  0 unexpected clusteroperator state transitions during e2e test run, as desired.
6 unwelcome but acceptable clusteroperator state transitions during e2e test run.  These should not happen, but because they are tied to exceptions, the fact that they did happen is not sufficient to cause this test-case to fail:

Apr 09 13:38:29.711 E clusteroperator/authentication condition/Available reason/APIServices_Error status/False APIServicesAvailable: apiservices.apiregistration.k8s.io/v1.user.openshift.io: not available: endpoints for service/api in "openshift-oauth-apiserver" have no addresses with port name "https" (exception: https://issues.redhat.com/browse/OCPBUGS-20056)
Apr 09 13:38:29.711 - 4s    E clusteroperator/authentication condition/Available reason/APIServices_Error status/False APIServicesAvailable: apiservices.apiregistration.k8s.io/v1.user.openshift.io: not available: endpoints for service/api in "openshift-oauth-apiserver" have no addresses with port name "https" (exception: https://issues.redhat.com/browse/OCPBUGS-20056)
Apr 09 13:38:33.922 W clusteroperator/authentication condition/Available reason/AsExpected status/True All is well (exception: Available=True is the happy case)
Apr 09 14:27:45.436 E clusteroperator/authentication condition/Available reason/APIServices_PreconditionNotReady status/False APIServicesAvailable: PreconditionNotReady (exception: https://issues.redhat.com/browse/OCPBUGS-20056)
Apr 09 14:27:45.436 - 77ms  E clusteroperator/authentication condition/Available reason/APIServices_PreconditionNotReady status/False APIServicesAvailable: PreconditionNotReady (exception: https://issues.redhat.com/browse/OCPBUGS-20056)
Apr 09 14:27:45.514 W clusteroperator/authentication condition/Available reason/AsExpected status/True All is well (exception: Available=True is the happy case)
}
```

Is that not the check you are tweaking?  Feel free to shard into multiple bugs if you want to take several incremental steps, but you probably want to either save the OCPBUGS-20056 name for the final step or follow any sharding up with a bump to the origin exception list to track your later tickets.

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/664/pull-ci-openshift-cluster-authentication-operator-master-e2e-agnostic-upgrade/1777676323997618176

stlaz:
@wking  This patch is focusing on the problem reported in OCPBUGS-20056, which is the check of the kube-apiserver's wellknown oauth endpoint. The issue from your report appears to be misattributing the bug to other problems. If you want those addressed, please open a different bug report.

wking:
> This patch is focusing on the problem reported in OCPBUGS-20056, which is the check of the kube-apiserver's wellknown oauth endpoint.

That's all I had room for in the title, but the `Description` has (since 2023-11-28) included:

> And simplifying by looking only at reason:
>
> ```
> curl -s 'https://search.ci.openshift.org/search?maxAge=48h&type=junit&name=4.15.*upgrade&context=0&search=clusteroperator/authentication.*condition/Available.*status/False' | jq -r 'to_entries[].value | to_entries[].value[].context[]' | sed 's|.*clusteroperator/\([^ ]*\) condition/Available reason/\([^ ]*\) status/False.*|\1 \2|' | sort | uniq -c | sort -n
>      1 authentication APIServerDeployment_PreconditionNotFulfilled
>      6 authentication OAuthServerDeployment_NoDeployment
>      8 authentication APIServerDeployment_NoDeployment
>     10 authentication APIServerDeployment_NoPod
>     11 authentication WellKnown_NotReady
>     36 authentication OAuthServerRouteEndpointAccessibleController_EndpointUnavailable
>     43 authentication APIServices_PreconditionNotReady
>     66 authentication OAuthServerServiceEndpointAccessibleController_EndpointUnavailable
>     95 authentication APIServices_Error
> ```
>
> ## Expected results:
> Authentication goes `Available=False` if and only if immediate admin intervention is appropriate.

Which covered both of the `reason` picked up in [this pull's CI run][1].  If you want me to do the sharing, were you looking for one bug for each `reason`?  Or would you like me to split things up differently?

[1]: https://github.com/openshift/cluster-authentication-operator/pull/664#issuecomment-2046281937

stlaz:
I see. I've only considered the fix of the WellKnown condition in the bug given it's description. That's about the amount of work that we can commit to for 4.16.

@wking If I might ask you to open another bug that tracks all the conditions and the name is a bit more generic (e.g. `single short-lived operand blip shouldn't cause authentication operator available=false`), that'd be great.

wking:
I've cloned OCPBUGS-32089 off of OCPBUGS-20056.  I retitled OCPBUGS-20056 to make that the generic one, because re-titling this pull to link the new ticket is easier than updating [the test-suite exception bug reference][1].

/retitle OCPBUGS-32089: wellknown-readiness: perform several attempts to connect before going unavailable

[1]: https://github.com/openshift/origin/blob/a42236a41a0037fea701052790e9129375398a1d/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go#L62-L63

PR #665, opened 2024-04-15
WRKLDS-1015: tolerate node-role.kubernetes.io/control-plane:NoExecute 

knelasevero:
Required to be able to advise customers to follow workflow described in https://github.com/openshift/enhancements/pull/1583. If they want to taint control-plane nodes to avoid any other thing to run there (while enforcing it further with validating admission policies), control plane workloads need to tolerate that taint.


stlaz:
/hold
the enhancement is unfinished

ingvagabund:
The enhancement touching the new toleration is finished.
/hold cancel

stlaz:
/hold
still valid

knelasevero:
> /hold still valid

Hey @stlaz so, the EP now is more like a formality. We needed the EP before because the original idea was to actually develop something (a new admision plugin/controller). Now, after review and feedback from @deads2k the workflow we are suggesting customers to follow are available out of the box (well, at least tech preview with ValidatingAdmissionPolicies with CEL if they enable it). So we are keeping the EP up mostly to keep history and link that in the PRs adding the tolarations that we have to open to multiple projects.

stlaz:
Formality or not, an enhancement being unmerged means that it's a subject to change according to our process.
Please also fix the unit tests.

stlaz:
Please squash the commits and fix the commit message. The "add annotation" commit does not add an annotation, but adds a toleration. Which toleration? Why?  Include all that.

PR #668, opened 2024-04-30
NO-JIRA: remove go-jose not to get flagged with its CVEs

stlaz:
go-jose is only used in tests, probably by a mistake anyway.

/assign @ibihim


liouk:
/retitle NO-JIRA: remove go-jose not to get flagged with its CVEs

PR #677, opened 2024-06-18
NO-JIRA: OWNERS: add ibihim, liouk, rm old

ibihim:
## What

Add ibihim as approver, add liouk as reviewer.

## Why

Both of us have contributed to the repository. I had the pleasure to refactor and work with the observed config to deliver an audit policy to the oauth-server.

Liouk worked on the proxy config.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,7 +1,6 @@
 reviewers:
-  - stlaz
+  - ibihim
+  - liouk
 approvers:
-  - stlaz
-  - sttts
   - deads2k
 components: apiserver-auth


deads2k:
/lgtm

it's important to understand the architectural choices about unified operator generation to know when levels have been achieved and why its important.  Otherwise,  it's easy to fall into "simplifications" that result in indeterminant behavior.  Especially when the behavior involves multiple caches of arbitrary latency.  the structural aspects of configuration observation prevent it when using that flow. But do not prevent it when bypassing that flow using things like configmaps.

vrutkovs:
/retitle NO-JIRA: OWNERS: add ibihim, liouk, rm old

PR #683, opened 2024-07-30
[release-4.16] NO-JIRA: Fixes for Keycloack IDP test and unit tests

liouk:
This is a manual cherry-pick of the following PRs:
- #679
- #680

Both are needed at the same time as both fix required tests (#679 fixes `e2e-operator`, #680 fixes `unit`).

### DIFF
--- a/pkg/libs/endpointaccessible/endpoint_accessible_controller_test.go
+++ b/pkg/libs/endpointaccessible/endpoint_accessible_controller_test.go
@@ -34,7 +34,7 @@ func Test_endpointAccessibleController_sync(t *testing.T) {
 		{
 			name: "non working endpoints",
 			endpointListFn: func() ([]string, error) {
-				return []string{"https://google.com", "https://nonexistenturl.com"}, nil
+				return []string{"https://httpbin.org/status/500"}, nil
 			},
 			wantErr: true,
 		},
--- a/test/library/idpdeployment.go
+++ b/test/library/idpdeployment.go
@@ -8,15 +8,18 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/ptr"
 
 	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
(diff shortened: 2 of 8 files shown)


test/library/idpdeployment.go
```diff
@@ -71,19 +81,75 @@ func deployPod(
 		metav1.CreateOptions{},
 	)
 
+	saName := name
```
ibihim, line 84: **nit:** if you want to create a copy of the name for the deployment and use it deliberately as the name of the ServiceAccount, I would do it above the creation of the ServiceAccount and use it in there:

```go
	saName := name
	_, err = clients.CoreV1().ServiceAccounts(namespace).Create(
		testContext,
		&corev1.ServiceAccount{
			ObjectMeta: metav1.ObjectMeta{
				Name: saName,
			},
		},
		metav1.CreateOptions{},
	)
```

test/library/idpdeployment.go
```diff
@@ -71,19 +81,75 @@ func deployPod(
...
 	require.NoError(t, err)
 
 	_, err = clients.CoreV1().Services(namespace).Create(testContext, svcTemplate(httpPort, httpsPort), metav1.CreateOptions{})
 	require.NoError(t, err)
 
+	timeLimitedCtx, cancel := context.WithTimeout(testContext, 5*time.Minute)
+	defer cancel()
+	_, err = watchtools.UntilWithSync(timeLimitedCtx,
+		cache.NewListWatchFromClient(
+			clients.AppsV1().RESTClient(), "deployments", namespace, fields.OneTermEqualSelector("metadata.name", deployment.Name)),
```
ibihim, line 143: Cool.

pkg/libs/endpointaccessible/endpoint_accessible_controller_test.go
```diff
@@ -34,7 +34,7 @@ func Test_endpointAccessibleController_sync(t *testing.T) {
 		{
 			name: "non working endpoints",
 			endpointListFn: func() ([]string, error) {
-				return []string{"https://google.com", "https://nonexistenturl.com"}, nil
+				return []string{"https://httpbin.org/status/500"}, nil
```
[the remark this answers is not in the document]
liouk: It was, anyway -- some domain sales service has picked up `https://nonexistenturl.com` and now it's working, when we expected it wouldn't :slightly_smiling_face:

test/library/idpdeployment.go
liouk: Good point, we missed this in the original PR -- but for the sake of consistency with the original (#679) let's not change it in the backport, WDYT?

pkg/libs/endpointaccessible/endpoint_accessible_controller_test.go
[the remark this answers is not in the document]
vrutkovs: We rely on external service to return an error - HTTP 500 if its successful or 403/DNS error if its broken

PR #684, opened 2024-07-30
AUTH-546: Direct external OIDC e2e test

liouk:
This PR adds an e2e test for the direct external OIDC authentication type (BYO-OIDC). It requires the `ExternalOIDC` feature gate to be enabled.

Currently, the e2e works as follows:
- deploy keycloak in the cluster, to use as an OIDC provider
- configure the OIDC as a direct provider in the KAS
   - update the authentication CR with the oidc provider configuration
   - sync the oidc provider's CA, if necessary, to the KAS pods static resources
   - patch the cluster proxy and the KAS CLI args to provide the OIDC configuration
   - wait for the changes to get rolled out
- run some basic keycloak sanity checks
- run some baseline authentication checks via the KAS

The goal of this test is to incrementally add more test cases, as development of the external OIDC feature progresses. Some upcoming TODOs:
- prove that oauth tokens won't work after the change
- ~~adjust user permissions so that the actions tested using an OIDC bearer token can succeed (currently authentication succeeds but authorization fails)~~ (this is done via a self subject review now)

### DIFF
--- a/test/e2e/external_oidc_test.go
+++ b/test/e2e/external_oidc_test.go
@@ -225,12 +225,12 @@ func (tc *testClient) setupExternalOIDCWithKeycloak(ctx context.Context) (kcClie
 	require.NoError(tc.t, err)
 
 	// sync service-ca signing certificate to the KAS nodes as a static resource so that it can be used with --oidc-ca-file
-	c, err = tc.syncServingCA(ctx, authConfig)
+	c, err = tc.syncServingCA(ctx)
 	cleanups = append(cleanups, c...)
 	require.NoError(tc.t, err)
 
 	// setup kube-apiserver to access the external OIDC directly by modifying its args via UnsupportedConfigOverrides
-	kasOrigRev, c, err := tc.updateKASArgsForOIDC(ctx, authConfig)
+	kasOrigRev, c, err := tc.updateKASArgsForOIDC(ctx, kcClient.IssuerURL())
 	cleanups = append(cleanups, c...)
 	require.NoError(tc.t, err)
 
@@ -291,10 +291,9 @@ func (tc *testClient) updateProxyForIngressCert(ctx context.Context) (cleanups [
 	return
 }
 
-func (tc *testClient) syncServingCA(ctx context.Context, authConfig *configv1.Authentication) (cleanups []func(), err error) {
-	oidcCAConfigMapName := authConfig.Spec.OIDCProviders[0].Issuer.CertificateAuthority.Name
-	if len(oidcCAConfigMapName) == 0 {
-		// no oidc CA present; ignore
+func (tc *testClient) syncServingCA(ctx context.Context) (cleanups []func(), err error) {
+	if len(oidcCABundleConfigMap) == 0 || len(oidcCABundleConfigMapNamespace) == 0 {
+		tc.t.Log("no oidc CA defined; will use system CA")
 		return nil, nil
 	}
 
@@ -331,11 +330,11 @@ func (tc *testClient) syncServingCA(ctx context.Context, authConfig *configv1.Au
 	return
 }
 
-func (tc *testClient) updateKASArgsForOIDC(ctx context.Context, authConfig *configv1.Authentication) (origRevision int32, cleanups []func(), err error) {
+func (tc *testClient) updateKASArgsForOIDC(ctx context.Context, idpURL string) (origRevision int32, cleanups []func(), err error) {
 	tc.t.Log("will update KAS args for OIDC")
 
 	oidcCAFile := "/etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt"
-	if len(authConfig.Spec.OIDCProviders[0].Issuer.CertificateAuthority.Name) > 0 {
+	if len(oidcCABundleConfigMap) > 0 {
 		oidcCAFile = "/etc/kubernetes/static-pod-resources/configmaps/oidc-serving-ca/ca-bundle.crt"
 	}
 
@@ -348,7 +347,7 @@ func (tc *testClient) updateKASArgsForOIDC(ctx context.Context, authConfig *conf
 			"oidc-username-claim": ["%s"],
 			"oidc-username-prefix":["-"]
 		}
-	}`, oidcCAFile, oidcClientId, authConfig.Spec.OIDCProviders[0].Issuer.URL, oidcGroupsClaim, oidcUsernameClaim)
+	}`, oidcCAFile, oidcClientId, idpURL, oidcGroupsClaim, oidcUsernameClaim)
 
 	kas, err := tc.operatorConfigClient.OperatorV1().KubeAPIServers().Get(ctx, "cluster", metav1.GetOptions{})
 	if err != nil {
--- a/test/e2e/external_oidc_test.go
+++ b/test/e2e/external_oidc_test.go
@@ -1,0 +1,450 @@
+package e2e
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	features "github.com/openshift/api/features"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	operatorversionedclient "github.com/openshift/client-go/operator/clientset/versioned"
+	test "github.com/openshift/cluster-authentication-operator/test/library"
+	"github.com/stretchr/testify/require"
+
+	authenticationv1 "k8s.io/api/authentication/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/storage/names"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	oidcClientId      = "admin-cli"
+	oidcAudience      = "openshift-aud"
+	oidcGroupsClaim   = "groups"
+	oidcUsernameClaim = "email"
+
+	kasNamespace = "openshift-kube-apiserver"
+
+	// set if a specific CA bundle is needed for the OIDC provider
+	oidcCABundleConfigMap          = ""
+	oidcCABundleConfigMapNamespace = ""
+)
+
+type testClient struct {
+	t *testing.T
+
+	kubeConfig           *rest.Config
+	kubeClient           *kubernetes.Clientset
+	configClient         *configclient.Clientset
+	operatorConfigClient *operatorversionedclient.Clientset
+}
+
+type oidcAuthResponse struct {
+	AccessToken      string `json:"access_token"`
+	ExpiresIn        int    `json:"expires_in"`
+	RefreshToken     string `json:"refresh_token"`
+	RefreshExpiresIn int    `json:"refresh_expires_in"`
+	TokenType        string `json:"token_type"`
+	IdToken          string `json:"id_token"`
+	NotBeforePolicy  int    `json:"not_before_policy"`
+	SessionState     string `json:"session_state"`
+	Scope            string `json:"scope"`
+}
+
+func TestExternalOIDCWithKeycloak(t *testing.T) {
+	testCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	tc, err := newTestClient(t)
+	require.NoError(t, err)
+
+	// ===============================
+	// Check ExternalOIDC feature gate
+	// ===============================
+
+	featureGates, err := tc.configClient.ConfigV1().FeatureGates().Get(testCtx, "cluster", metav1.GetOptions{})
+	require.NoError(tc.t, err)
+
+	if len(featureGates.Status.FeatureGates) != 1 {
+		// fail test if there are multiple feature gate versions (i.e. ongoing upgrade)
+		tc.t.Fatalf("multiple feature gate versions detected")
+	} else {
+		for _, fgDisabled := range featureGates.Status.FeatureGates[0].Disabled {
+			if fgDisabled.Name == features.FeatureGateExternalOIDC {
+				tc.t.Skipf("feature gate '%s' disabled", features.FeatureGateExternalOIDC)
+			}
+		}
+	}
+
+	// ==============
+	// Setup Keycloak
+	// ==============
+
+	var kcClient *test.KeycloakClient
+	if keycloakURL := os.Getenv("E2E_KEYCLOAK_URL"); len(keycloakURL) > 0 {
+		t.Logf("will use existing keycloak deployment at URL: %s", keycloakURL)
+		kcClient = tc.setupKeycloakClient(testCtx, keycloakURL)
+
+	} else {
+		t.Logf("no existing keycloak deployment found; will create new")
+		var cleanups []func()
+		kcClient, cleanups = tc.setupExternalOIDCWithKeycloak(testCtx)
+		defer test.IDPCleanupWrapper(func() {
+			for _, c := range cleanups {
+				c()
+			}
+		})()
+		t.Logf("keycloak Admin URL: %s", kcClient.AdminURL())
+	}
+
+	group := names.SimpleNameGenerator.GenerateName("e2e-keycloak-group-")
+	err = kcClient.CreateGroup(group)
+	require.NoError(t, err)
+
+	user := names.SimpleNameGenerator.GenerateName("e2e-keycloak-user-")
+	email := fmt.Sprintf("%s@test.dev", user)
+	password := "password"
+	err = kcClient.CreateUser(user, email, password, []string{group})
+	require.NoError(t, err)
+
+	httpClient := &http.Client{Transport: &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}}
+
+	formData := url.Values{
+		"grant_type": []string{"password"},
+		"client_id":  []string{oidcClientId},
+		"scope":      []string{"openid"},
+		"username":   []string{user},
+		"password":   []string{password},
+	}
+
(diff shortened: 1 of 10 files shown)


test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,471 @@
...
+		kas.Spec.UnsupportedConfigOverrides = origUnsupportedConfigOverides
+		kas, err = tc.operatorConfigClient.OperatorV1().KubeAPIServers().Update(ctx, kas, metav1.UpdateOptions{})
+		if err != nil {
+			tc.t.Logf("cleanup failed for kube-apiserver '%s': %v", kas.Name, err)
+			return
+		}
+
+		err = test.WaitForNewKASRollout(tc.t, ctx, tc.operatorConfigClient.OperatorV1().KubeAPIServers(), origRevision)
+		if err != nil {
+			tc.t.Logf("cleanup failed for kube-apiserver '%s': %v", kas.Name, err)
```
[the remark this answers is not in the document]
liouk: That's a great suggestion :+1:

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,471 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/apiserver/pkg/storage/names"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	watchtools "k8s.io/client-go/tools/watch"
+)
+
+const (
+	externalOIDCFeatureGate = "ExternalOIDC"
```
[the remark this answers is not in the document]
liouk: I'm not very familiar with featuregates yet, thanks for pointing this out :+1:

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,471 @@
...
+	cleanups = append(cleanups, func() {
+		auth.Spec = *origSpec
+		_, err = tc.configClient.ConfigV1().Authentications().Update(ctx, auth, metav1.UpdateOptions{})
+		if err != nil {
+			tc.t.Logf("cleanup failed for authentication '%s': %v", auth.Name, err)
+		}
+	})
+
+	// wait for auth CR to get patched
+	origGen := auth.Generation
```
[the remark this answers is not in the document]
liouk: You're right, good catch -- this was leftover actually; in the beginning there was code patching the featuregates in a script, and was waiting for the new CR to get rolled out. Removed this completely.

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,471 @@
...
+	require.NoError(t, err)
+
+	// kubernetes uses the id_token to identify the user (instead of the access_token)
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", authResponse.IdToken))
+	resp, err = httpClient.Do(req)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	// user is authenticated (request is forbidden due to insufficient permissions)
+	require.Equal(t, http.StatusForbidden, resp.StatusCode)
```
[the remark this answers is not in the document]
liouk: Great idea; it was on my todo list to figure out a better way to assert this :+1:

liouk:
@benluddy I've added fixes as per your comments, thanks for the great review. They're fixups now for clarity; I'll squash them once you've reviewed the changes.

Holding until fixups are squashed.
/hold

test/e2e/external_oidc_test.go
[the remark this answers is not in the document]
liouk: @benluddy this requires a bump in openshift/api, which then causes a series of other dep bumps, eventually leading to some breaking changes in client-go. Would it be OK if I handle this at a separate PR that bumps all deps and fixes incompatibilities, in order to not drag this around in this PR?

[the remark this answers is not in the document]
benluddy: OK. I see you at least made this a typed constant, that's good. Could you also add a comment here to mention that this duplicates the constant declared in openshift/api until the dependency issue is resolved?

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,449 @@
...
+	require.NoError(t, err)
+	require.NotEmpty(t, authResponse.AccessToken)
+	require.NotEmpty(t, authResponse.IdToken)
+
+	// ==========================================
+	// Test authentication via the kube-apiserver
+	// ==========================================
+
+	// use the OIDC id_token to do a self subject access review
+	kasURL := fmt.Sprintf("%s/apis/authorization.k8s.io/v1/selfsubjectaccessreviews", tc.kubeConfig.Host)
```
benluddy, line 159: https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/self-subject-review-v1/ actually returns the username/groups of the authenticated user making the request. If I'm understanding the test, we want to verify that the provided token is authenticated properly. If the test is only checking that the requester is _authorized_ to make a certain request, that's more likely to give a false-positive result.

test/e2e/external_oidc_test.go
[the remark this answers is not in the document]
liouk: Done, thanks :+1:

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,461 @@
...
+	// ===============================
+	// Check ExternalOIDC feature gate
+	// ===============================
+
+	featureGates, err := tc.configClient.ConfigV1().FeatureGates().Get(testCtx, "cluster", metav1.GetOptions{})
+	require.NoError(tc.t, err)
+
+	if len(featureGates.Status.FeatureGates) != 1 {
+		// skip test if there are multiple feature gate versions (i.e. ongoing upgrade)
+		tc.t.Skipf("multiple feature gate versions detected")
```
benluddy, line 87: I'd fail instead of skip if the test isn't prepared to handle multiple versions. The risk of skipping is that if this test is later added to an upgrade job config, no one will notice that this functionality is not being tested. Alternatively, check for the feature gate in the latest version in the ClusterVersionStatus?

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,461 @@
...
+type oidcAuthResponse struct {
+	AccessToken      string `json:"access_token"`
+	ExpiresIn        int    `json:"expires_in"`
+	RefreshToken     string `json:"refresh_token"`
+	RefreshExpiresIn int    `json:"refresh_expires_in"`
+	TokenType        string `json:"token_type"`
+	IdToken          string `json:"id_token"`
+	NotBeforePolicy  int    `json:"not_before_policy"`
+	SessionState     string `json:"session_state"`
+	Scope            string `json:"scope"`
```
xingxingxia, line 68: Nit :) , why does the struct type use first-character lower case while the struct member uses first-character upper case?

test/e2e/external_oidc_test.go
xingxingxia: Instead of `Get` featuregate/cluster's `.status.featureGates` array, how about checking below configmap which always reflects the feature gates in `latest version`?
In an env which enables it:
```
$ oc get cm config -n openshift-kube-apiserver -o=jsonpath='{.data.config\.yaml}' \
    | jq '.apiServerArguments."feature-gates"'
...
  "ExternalOIDC=true",
...
```
In an env which does not enable it:
```
$ oc get cm config -n openshift-kube-apiserver -o=jsonpath='{.data.config\.yaml}' \
    | jq '.apiServerArguments."feature-gates"'
...
  "ExternalOIDC=false",
...
```

test/e2e/external_oidc_test.go
liouk: The type is lowercase as there's no need to export it to other packages. The members are exported (first-character upper case) because this is needed in order to be able to deserialize JSON data into the struct using `json.Unmarshal()`.

test/library/idpdeployment.go
```diff
@@ -136,7 +136,7 @@ func deployPod(
 	_, err = clients.CoreV1().Services(namespace).Create(testContext, svcTemplate(httpPort, httpsPort), metav1.CreateOptions{})
 	require.NoError(t, err)
 
-	timeLimitedCtx, cancel := context.WithTimeout(testContext, 5*time.Minute)
+	timeLimitedCtx, cancel := context.WithTimeout(testContext, 10*time.Minute)
```
liouk, line 139: @benluddy FYI since your last review, I decided to double this because I was noticing several flakes timing out on this.

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,461 @@
...
+	defer resp.Body.Close()
+	require.Equal(t, http.StatusCreated, resp.StatusCode)
+
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+	var ssr authenticationv1.SelfSubjectReview
+	err = json.Unmarshal(body, &ssr)
+	require.NoError(t, err)
+	require.Equal(t, email, ssr.Status.UserInfo.Username)
+	require.Contains(t, ssr.Status.UserInfo.Groups, "system:authenticated")
```
xingxingxia, line 182: Need it also check the keycloak group above?

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,461 @@
...
+		err := tc.kubeClient.CoreV1().ConfigMaps(kasNamespace).Delete(ctx, oidcServingCA.Name, metav1.DeleteOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			tc.t.Fatalf("cleanup failed for secret '%s/%s': %v", oidcServingCA.Namespace, oidcServingCA.Name, err)
+		}
+	})
+	if err != nil {
+		return
+	}
+
+	return
```
xingxingxia, line 347: Seems only creating a configmap is not enough. We need also make the "oidc-serving-ca" configmap mounted in the kube-apiserver pods, right? I'm not sure currently whether there is a way to achieve the mounting [1] : )
[1] I tried mounting as below, but the kube-apiserver-asahay02417-m2j57-master-0 never appears:
```
On one control plane node master-0, modify /etc/kubernetes/manifests/kube-apiserver-pod.yaml to append below volumeMounts and volumes:
{"spec":{"containers":[{"name":"kube-apiserver","volumeMounts":[...,{"mountPath":"/etc/kubernetes/static-pod-resources/configmaps/oidc-serving-ca","name":"oidc-ca"}]}],"volumes":[...,{"configMap":{"defaultMode":420,"name":"oidc-serving-ca","items":[{"key":"ca-bundle.crt","path":"ca-bundle.crt"}]},"name":"oidc-ca"}]}}

The kube-apiserver-asahay02417-m2j57-master-0 never appears:
$ oc get po -n openshift-kube-apiserver -l apiserver
NAME                                        READY   STATUS    RESTARTS   AGE
kube-apiserver-asahay02417-m2j57-master-1   5/5     Running   0          3h28m
kube-apiserver-asahay02417-m2j57-master-2   5/5     Running   0          3h34m
```

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,461 @@
...
+
+func (tc *testClient) syncServingCA(ctx context.Context) (cleanups []func(), err error) {
+	if len(oidcCABundleConfigMap) == 0 || len(oidcCABundleConfigMapNamespace) == 0 {
+		tc.t.Log("no oidc CA defined; will use system CA")
+		return nil, nil
+	}
+
+	tc.t.Log("will sync OIDC ca-bundle to KAS static pod resources")
+
+	signingKey, err := tc.kubeClient.CoreV1().Secrets(oidcCABundleConfigMapNamespace).Get(ctx, oidcCABundleConfigMap, metav1.GetOptions{})
```
xingxingxia, line 321: One nit, the variable **name** `oidcCABundleConfigMap` indicates the resource is a "ConfigMap", but the API function **name** `CoreV1().Secrets(...)` intends a "Secret" : )

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,461 @@
...
+func (tc *testClient) setupExternalOIDCWithKeycloak(ctx context.Context) (kcClient *test.KeycloakClient, cleanups []func()) {
+	kcClient, idpName, c := test.AddKeycloakIDP(tc.t, tc.kubeConfig, true)
+	cleanups = append(cleanups, c...)
+
+	// update the authentication CR with the external OIDC configuration
+	authConfig, c, err := tc.updateAuthForOIDC(ctx, kcClient.IssuerURL(), idpName)
+	cleanups = append(cleanups, c...)
+	require.NoError(tc.t, err)
+	require.NotNil(tc.t, authConfig)
+	require.NotEmpty(tc.t, authConfig.Spec.OIDCProviders)
```
xingxingxia, line 237: One stupid question : ): in case one `require.XXX` assertion fails, will the `cleanups` still be executed in a later time?

test/library/waits.go
```diff
@@ -138,3 +139,32 @@ func WaitForHTTPStatus(t *testing.T, waitDuration time.Duration, client *http.Cl
 		return false, nil
 	})
 }
+
+func WaitForNewKASRollout(t *testing.T, ctx context.Context, kasClient operatorv1.KubeAPIServerInterface, origRevision int32) error {
+	err := wait.PollUntilContextTimeout(ctx, 10*time.Second, 30*time.Minute, true, func(ctx context.Context) (bool, error) {
+		kas, err := kasClient.Get(ctx, "cluster", metav1.GetOptions{})
```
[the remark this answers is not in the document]
liouk: I think it's a better practice to pass the context into `WaitForNewKASRollout`, so that we can control its execution with the same context; i.e. if that context gets cancelled (or its parent), then the wait would also be cancelled. I am not sure why `TODO()` is used in the other tests instead of an actual test context.

test/e2e/external_oidc_test.go
liouk: You are right @xingxingxia -- we will need something like this: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1720

For now this functionality isn't actually needed for the test, it's more of a placeholder for when we further develop this feature. I'll add a comment to the code with a link to that PR.

test/e2e/external_oidc_test.go
liouk: That's actually not a nit, indeed it's a config map :) I will update the API call. Nice catch :+1:

test/e2e/external_oidc_test.go
liouk: `require` uses `t.FailNow()` under the hood, which in turn calls `runtime.Goexit()` -- deferred funcs will still be executed after these are called.

From https://pkg.go.dev/runtime#Goexit:
> Goexit runs all deferred calls before terminating the goroutine.

test/e2e/external_oidc_test.go
liouk: As far as I understand, that group does not appear in the id/access token that Keycloak returns because I haven't created any group mapper in Keycloak to include that claim into the tokens. I'll see how this can be done and will try to incorporate it.

This means that currently, `system:authenticated` is the only group returned by the SSR -- this should be enough for now for the current purposes of this test.

test/e2e/external_oidc_test.go
liouk: Fixed in 0e62f4e60d8277f43443976f4fefcc98909f8af5

test/e2e/external_oidc_test.go
liouk: Fixed in 0e62f4e60d8277f43443976f4fefcc98909f8af5

test/library/waits.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	operatorv1 "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
```
benluddy, line 17: Nit:
```suggestion
	operatorv1client "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
```

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,468 @@
...
+	// setup kube-apiserver to access the external OIDC directly by modifying its args via UnsupportedConfigOverrides
+	kasOrigRev, c, err := tc.updateKASArgsForOIDC(ctx, kcClient.IssuerURL())
+	cleanups = append(cleanups, c...)
+	require.NoError(tc.t, err)
+
+	// wait for serving CA and KAS args overrides to get rolled out
+	tc.t.Log("will wait for KAS rollout")
+	err = test.WaitForNewKASRollout(tc.t, ctx, tc.operatorConfigClient.OperatorV1().KubeAPIServers(), kasOrigRev)
+	if err != nil {
+		return
```
benluddy, line 258: Is the test supposed to fail in this case?

test/library/idpdeployment.go
benluddy: Is there some misconfiguration in the keycloak deployment that is causing it to take so long? I don't know if it needs to be addressed immediately, but it would be a shame to needlessly pay an extra ~five minutes of test runtime (and test infrastructure costs) forever.

test/e2e/external_oidc_test.go
```diff
@@ -0,0 +1,468 @@
...
+
+	var authResponse oidcAuthResponse
+	err = json.Unmarshal(data, &authResponse)
+	require.NoError(t, err)
+	require.NotEmpty(t, authResponse.AccessToken)
+	require.NotEmpty(t, authResponse.IdToken)
+
+	// ==========================================
+	// Test authentication via the kube-apiserver
+	// ==========================================
```
benluddy, line 160: Was there a particular reason to avoid using a Kubernetes client for this part?

test/e2e/external_oidc_test.go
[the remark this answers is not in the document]
liouk: Now using `github.com/openshift/api/features/FeatureNameExternalOIDC` in d89947d01557fc41b98fbbf551ea8a8ca0e196bc

test/e2e/external_oidc_test.go
liouk: Great idea, fixed in d89947d01557fc41b98fbbf551ea8a8ca0e196bc

test/e2e/external_oidc_test.go
benluddy: Could this assert on the username instead? I imagine we don't want this test to pass if the request user is authenticated by some means other than the bearer token.

liouk:
fixups squashed; cancelling hold

/hold cancel

PR #687, opened 2024-08-28
OKD-226: Dockerfile.rhel7: Add ARG TAGS=ocp

Prashanth684:
Allow this ARG to be overriden by builds so that OKD builds may have the correct branding by default. OKD/SCOS builds pass the build-arg TAGS=scos so pkg/controllers/configobservation/oauth/brand_okd.go which has //go:build !ocp will be built instead of pkg/controllers/configobservation/oauth/brand_ocp.go which has //go:build ocp. This change should not affect the OCP build of the operator image.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,8 +1,10 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.18 AS builder
+# This pattern is documented in https://github.com/openshift/enhancements/pull/1672. Use ARG to build images based on platform.
+ARG TAGS=ocp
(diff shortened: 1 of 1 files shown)


Dockerfile.rhel7
```diff
@@ -1,8 +1,9 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
+ARG TAGS=ocp
```
deads2k, line 2: please add a comment to the dockerfile linking to a doc that lists valid values of TAGS. openshift/enhancement preferred.

And any other "standard" build-args.

ausil: AFAIK any value other than ocp means OKD

PR #692, opened 2024-09-10
OCPBUGS-41586: increase oauth-apiserver failureThreshold

liouk:
This PR improves the liveness/readiness checks of the oauth-apiserver.

In particular:
- set `etcd-healthcheck-timeout` and `etcd-readycheck-timeout` to 1sec less than the respective probe timeouts
- replace liveness probe path `healthz` with `livez`, and exclude `etcd` from the probe
- increase the readiness probe failure threshold to 3
- set `shutdown-delay-duration` to 50s; the kubelet won't wait for another period if the previous probe fails (see note below), so if `timeoutSeconds > periodSeconds` it'll fire the subsequent probes immediately, therefore worst case is `5s + 3*10s + 15s (buffer)`
- set `terminationGracePeriodSeconds` to 120s: non-long-running requests timeout (60s) + shutdown delay (50s) + 10s buffer

Note (thanks @benluddy): the probes are implemented using a ticker, which internally has a buffered channel of length 1; if a tick fires and the buffered object hasn't been read yet (i.e. the previous probe is still running), that tick is dropped. See https://github.com/kubernetes/kubernetes/blob/5639b4b29cbf1ce7acab1e307ec2c0bdab1a65f5/pkg/kubelet/prober/worker.go#L175-L184

### DIFF
--- a/bindata/oauth-apiserver/deploy.yaml
+++ b/bindata/oauth-apiserver/deploy.yaml
@@ -72,7 +72,9 @@ spec:
               --etcd-cafile=/var/run/configmaps/etcd-serving-ca/ca-bundle.crt \
               --etcd-keyfile=/var/run/secrets/etcd-client/tls.key \
               --etcd-certfile=/var/run/secrets/etcd-client/tls.crt \
-              --shutdown-delay-duration=15s \
+              --etcd-healthcheck-timeout=9s \
+              --etcd-readycheck-timeout=9s \
+              --shutdown-delay-duration=50s \
               --shutdown-send-retry-after=true \
               --tls-private-key-file=/var/run/secrets/serving-cert/tls.key \
               --tls-cert-file=/var/run/secrets/serving-cert/tls.crt \
@@ -118,7 +120,7 @@ spec:
           httpGet:
             scheme: HTTPS
             port: 8443
-            path: healthz
+            path: livez?exclude=etcd
           initialDelaySeconds: 0
           periodSeconds: 10
           timeoutSeconds: 10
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
/lgtm

let's wait until https://github.com/openshift/cluster-openshift-apiserver-operator/pull/588 merges
/hold

p0lyn0mial:
I've checked https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/692/pull-ci-openshift-cluster-authentication-operator-master-e2e-operator/1833818860428464128 for `TerminationMinimalShutdownDurationFinished` event and it says: `The minimal shutdown duration of 50s finished`

bindata/oauth-apiserver/deploy.yaml
```diff
@@ -133,18 +135,18 @@ spec:
...
           timeoutSeconds: 10
           successThreshold: 1
-          failureThreshold: 1
+          failureThreshold: 3
         startupProbe:
           httpGet:
             scheme: HTTPS
             port: 8443
-            path: healthz
+            path: livez
```
benluddy, line 143: Exclude etcd?

bindata/oauth-apiserver/deploy.yaml
```diff
@@ -133,18 +135,18 @@ spec:
...
           timeoutSeconds: 10
           successThreshold: 1
-          failureThreshold: 1
+          failureThreshold: 3
         startupProbe:
           httpGet:
             scheme: HTTPS
             port: 8443
-            path: healthz
+            path: livez?exclude=etcd
```
deads2k, line 143: blocking comment: as indicated in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/588#discussion_r1755137655, it's a good idea to ensure that we have etcd working during *startup*, but it's a good idea to exclude it from ongoing checks.  This means that the startup problem should not exclude etcd and the liveness probe should exclude etcd

p0lyn0mial:
/hold cancel
/lgtm

for approval
/assign @deads2k

PR #697, opened 2024-09-18
API-1835: operator client update

deads2k:
Updates us to the new client that starts using SSA.

/assign @p0lyn0mial @liouk


p0lyn0mial:
LGTM

I will defer to @liouk for tagging it.

PR #698, opened 2024-09-26
API-1835: Migrate a few existing controllers to ApplyOperatorStatus

bertinatto:
In this PR I'm focusing on converting `UpdateOperatorStatus` to `ApplyOperatorSTatus` and measuring the overhead. There are several usages of `Update` and `UpdateStatus` that needs to be looked into, but I'll likely leave that for a second PR.

Verb usage beween `e2e-operator` job from this PR and the same job from https://github.com/openshift/cluster-authentication-operator/pull/704:

verb_usage_comparison_auth


pkg/controllers/common/conditions.go
```diff
@@ -104,3 +105,30 @@ func UpdateControllerConditions(ctx context.Context, operatorClient v1helpers.Op
 
 	return nil
 }
+
+func ApplyControllerConditions(ctx context.Context, operatorClient v1helpers.OperatorClient, fieldManager string, allConditionNames sets.String, updatedConditions []operatorv1.OperatorCondition) error {
```
deads2k, line 109: dead?  I think I do prefer the "normal" style over this.

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -117,47 +120,49 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
...
 	}
 
 	if err := c.isWellknownEndpointsReady(ctx, operatorSpec, operatorStatus, authConfig, infraConfig); err != nil {
-		statusUpdates = append(statusUpdates, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
-			Type:    "WellKnownAvailable",
-			Status:  operatorv1.ConditionFalse,
-			Reason:  "NotReady",
-			Message: fmt.Sprintf("The well-known endpoint is not yet available: %s", err.Error()),
-		}))
+		status = status.WithConditions(applyoperatorv1.OperatorCondition().
```
deads2k, line 134: `WithConditions` appends, so you can have two `WellKnownAvailable`.

Perhaps it's better to have separate method variables for each condition we're going to set and then put them together in defer?

bertinatto: Good idea, created individual variables for conditions and added them to the status in defer.

pkg/controllers/common/conditions.go
bertinatto: Sorry you reviewed the PR at the point :). This was a copy-paste from the function above in order to allow me to gradually migrate controllers. This is gone now.

pkg/controllers/readiness/wellknown_ready_controller.go
bertinatto: I removed the changes for this controller and will handle them in a separate PR (https://github.com/openshift/cluster-authentication-operator/pull/705).

bertinatto:
/test e2e-operator

```
: TestKeycloakAsOIDCPasswordGrantCheckAndGroupSync expand_less 	10m0s
{=== RUN   TestKeycloakAsOIDCPasswordGrantCheckAndG ...  === RUN   TestKeycloakAsOIDCPasswordGrantCheckAndG ...}
```

Looks like it's happening in 4.17 too: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/703/pull-ci-openshift-cluster-authentication-operator-release-4.17-e2e-operator/1841111144366870528

pkg/controllers/common/conditions.go
```diff
@@ -79,8 +80,8 @@ func ControllerProgressingConditionName(controllerName string) string {
 	return controllerName + "Progressing"
 }
 
-func UpdateControllerConditions(ctx context.Context, operatorClient v1helpers.OperatorClient, allConditionNames sets.String, updatedConditions []operatorv1.OperatorCondition) error {
-	updateConditionFuncs := []v1helpers.UpdateStatusFunc{}
+func UpdateControllerConditions(ctx context.Context, operatorClient v1helpers.OperatorClient, fieldManager string, allConditionNames sets.String, updatedConditions []operatorv1.OperatorCondition) error {
```
p0lyn0mial, line 83: nit: wouldn't be better to rename the function to `ApplyControllerConditions` ?

pkg/controllers/common/conditions.go
```diff
@@ -95,12 +96,13 @@ func UpdateControllerConditions(ctx context.Context, operatorClient v1helpers.Op
...
-		return err
+		status = status.WithConditions(applyoperatorv1.OperatorCondition().
+			WithType(newCondition.Type).
+			WithStatus(newCondition.Status).
+			WithReason(newCondition.Reason).
+			WithMessage(newCondition.Message))
 	}
 
-	return nil
+	return operatorClient.ApplyOperatorStatus(ctx, fieldManager, status)
```
p0lyn0mial, line 107: should we skip the call when `allConditionNames` is empty ?

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
```diff
@@ -53,7 +56,7 @@ func NewIngressNodesAvailableController(
 		).
 		WithSync(controller.sync).
 		ResyncEvery(wait.Jitter(time.Minute, 1.0)).
-		ToController("IngressNodesAvailableController", eventRecorder)
+		ToController(instanceName, eventRecorder)
```
p0lyn0mial, line 59: I think that in library-go we are passing `c.controllerInstanceName` (applies to the other changes in this PR)

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
```diff
@@ -83,7 +87,7 @@ func NewWebhookAuthenticatorController(
 	).ResyncEvery(wait.Jitter(time.Minute, 1.0)).
 		WithSync(c.sync).
 		WithSyncDegradedOnError(operatorClient).
-		ToController("WebhookAuthenticatorController", recorder.WithComponentSuffix("webhook-authenticator-controller"))
+		ToController(instanceName, recorder.WithComponentSuffix("webhook-authenticator-controller"))
```
p0lyn0mial, line 90: I think that we cannot rename this controller without removing existing condition added by the `WithSyncDegradedOnError`.

bertinatto: Actually we aren't renaming the controller, it's got the same name here:  https://github.com/openshift/cluster-authentication-operator/pull/698/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R720

p0lyn0mial: ok, then please add a comment where you set the name to not change it without clearing the condition.

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
p0lyn0mial: i'm fine either way but we should be consistent.

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
p0lyn0mial: for example: https://github.com/openshift/library-go/blob/ac65cd59c78f04a82617f7aecd6a0f1d16f985cb/pkg/operator/apiserver/controller/auditpolicy/auditpolicy_controller.go#L62

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
p0lyn0mial: given that for some controller we cannot change the name then maybe we should use `instanceName` ?

bertinatto: That's my understanding as well.

pkg/controllers/webhookauthenticator/webhookauthenticator_controller.go
bertinatto: Good idea; done

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
p0lyn0mial: Actually I take it back. @deads2k created this m****s and was not willing to take into account feedback given in https://github.com/openshift/library-go/pull/1794 :(.

The `instanceName` is the name of the operator, for example, `openshift-apiserver` or `openshift-apiserver-operator`, and the `controllerInstanceName` is the `instanceName` combined with the name of the controller.

Given this, it doesn't make sense to use the `instanceName` as the controller's name, as this would result in multiple controllers sharing the same name.

For the `IngressNodesAvailableController`, we can change its name because the previous name was not used to generate a condition. However, for controllers created with `WithSyncDegradedOnError`, it is not safe to change their name.

Does that make sense ?

p0lyn0mial: [NewCSIDriverNodeServiceController](https://github.com/openshift/library-go/pull/1794/files/a86ed6242ce1d886cd79fbb1785829208d4d4f35#diff-9e9f771b56ad2865cbcbaf9907510ed027520a166736511b6364e68b11da543e) from the library-go PR mentioned earlier use `WithSyncDegradedOnError`, thus its name could not be changed. 

Given how this controller is [created](https://github.com/search?q=org%3Aopenshift%20NewCSIDriverNodeServiceController&type=code), the `instanceName` will be`NFSDriverNodeServiceController` and the `controllerInstanceName` will be `NFSDriverNodeServiceController-CSIDriverNodeService`

bertinatto: Thanks for the explanation, @p0lyn0mial.
I'm worried that some variables, like `instanceName`, have different meanings in different places. For example, in this operator it would be `oauth-apiserver`, but in [storage](https://github.com/openshift/library-go/pull/1794/files/a86ed6242ce1d886cd79fbb1785829208d4d4f35#diff-9e9f771b56ad2865cbcbaf9907510ed027520a166736511b6364e68b11da543e)  it would be `NFSDriverNodeServiceController`, as you pointed out.

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
```diff
@@ -54,7 +57,9 @@ func NewEndpointAccessibleController(
 		WithSync(c.sync).
 		ResyncEvery(wait.Jitter(time.Minute, 1.0)).
 		WithSyncDegradedOnError(operatorClient).
-		ToController(controllerName, recorder.WithComponentSuffix(name+"endpoint-accessible-controller"))
+		ToController(
+			controllerName, // Don't change what is passed here unless you also remove the old FooDegraded condition
```
bertinatto, line 61: This one is a bit strange. Degraded is prefixed by `controllerName`, but Available is prefixed by `name`.

go.mod
```diff
@@ -6,111 +6,113 @@ require (
...
-	k8s.io/klog/v2 v2.120.1
-	k8s.io/kube-aggregator v0.30.1
+	k8s.io/api v0.31.1
+	k8s.io/apimachinery v0.31.1
+	k8s.io/apiserver v0.31.1
+	k8s.io/client-go v0.31.1
+	k8s.io/component-base v0.31.1
+	k8s.io/klog/v2 v2.130.1
+	k8s.io/kube-aggregator v0.31.1
 	k8s.io/pod-security-admission v0.29.0
```
p0lyn0mial, line 25: no 1.31.1 version ?

pkg/libs/endpointaccessible/endpoint_accessible_controller.go
p0lyn0mial: Yes, but we cannot change these names I think.

pkg/operator/starter.go
```diff
@@ -717,6 +717,7 @@ func prepareOauthAPIServerOperator(ctx context.Context, controllerContext *contr
 	)
 
 	webhookAuthController := webhookauthenticator.NewWebhookAuthenticatorController(
+		"openshift-authentication",
```
p0lyn0mial, line 720: I like this name. Thanks.

p0lyn0mial (commented):
LGTM

go.mod
bertinatto: This one is a direct dependency of this module, so it needs to be updated manually.

If you prefer I can add a commit to update it to `v0.31.1`, but from what I can see, there are no changes, just the version.

p0lyn0mial: ok, then let's leave it on 0.29.0

PR #702, opened 2024-09-27
OCPBUGS-42535: when no type is specified, don't make illegal condition

deads2k:


pkg/controllers/routercerts/controller.go
```diff
@@ -100,6 +100,11 @@ func (c *routerCertsDomainValidationController) sync(ctx context.Context, syncCt
 	// set the condition anywhere in sync() to update the controller's degraded condition
 	var condition operatorv1.OperatorCondition
 	defer func() {
+		if len(condition.Type) == 0 {
+			// no change is desired.  This happens when the SyncPartialSecret fails
```
p0lyn0mial, line 104: yeah, in line `123`

pkg/controllers/routercerts/controller.go
```diff
@@ -100,6 +100,11 @@ func (c *routerCertsDomainValidationController) sync(ctx context.Context, syncCt
 	// set the condition anywhere in sync() to update the controller's degraded condition
 	var condition operatorv1.OperatorCondition
 	defer func() {
+		if len(condition.Type) == 0 {
```
p0lyn0mial, line 103: how about adding a helper function for checking if a condition is empty ?

deads2k: to save a length check?  No.

PR #705, opened 2024-10-01
API-1835: Migrate WellKnownReadyController to SSA

bertinatto:
Compared **e2e-operator** job from https://github.com/openshift/cluster-authentication-operator/pull/709 with the same job from this PR:

verb_usage_comparison_auth

### DIFF
--- a/pkg/controllers/readiness/wellknown_ready_controller.go
+++ b/pkg/controllers/readiness/wellknown_ready_controller.go
@@ -18,6 +18,7 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configinformer "github.com/openshift/client-go/config/informers/externalversions"
 	configv1lister "github.com/openshift/client-go/config/listers/config/v1"
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
 	routeinformer "github.com/openshift/client-go/route/informers/externalversions/route/v1"
 	routev1lister "github.com/openshift/client-go/route/listers/route/v1"
 	"github.com/openshift/cluster-authentication-operator/pkg/controllers/common"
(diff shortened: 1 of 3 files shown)


bertinatto:
/hold
for https://github.com/openshift/cluster-authentication-operator/pull/698

bertinatto:
> /hold for #698

/hold cancel

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -106,9 +111,21 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
...
 	defer func() {
-		if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, statusUpdates...); updateErr != nil {
+		// only add valid conditions to the status. This is to prevent unintended changes in current behavior,
+		// where the sync function may return without setting a progressing condition.
+		conditions := []*applyoperatorv1.OperatorConditionApplyConfiguration{
+			available,
+			progressing,
+		}
+		for _, condition := range conditions {
+			if condition != nil {
```
p0lyn0mial, line 124: what happens to a condition that is `nil` ? I think it will be removed. Is this what we want :) ?

bertinatto: IIUC it won't be removed because the field is a [listType=map](https://github.com/openshift/api/blob/b66574fab23f29af23c8ea2abd9641bae44b388e/operator/v1/types.go#L118-L119). Based on this, we should only update the condition we are specifically interested in and leave the other existing conditions untouched.

This logic (only apply valid, non-nil conditions) was introduced to align with the current behavior. For example, imagine get into this branch:

https://github.com/openshift/cluster-authentication-operator/pull/705/files#diff-b6f4bf160fd7e801eadfbfac60b84bd00fcad21a0979c2d730549f10db015645R136-L125

Under the current behavior, only the available condition is set, the progressing condition is ignored.

I did try to set the progressing condition unconditionally, but that's not what CI expects.

Does that make sense?

deads2k: with SSA and listType=map, the behavior goes liek this

1. SSA for fieldmanager/one that sets condition/foo and condition/bar
2. objectect has condition/foo and condition/bar
3. SSA  for fieldmanager/one, sets condition/bar
4. object has condition/bar, condition/foo is removed

I suspect we wnt to always set available adn progressing.

bertinatto: hm, what if step 3 is:

3. SSA for fieldmanager/**two** sets condition/**baz**

then step 4 would be:

4. object has condition/bar, condition/foo **and** condition/baz?

is that correct?

deads2k: > is that correct?

Yes.  The fieldmanager/2 doesn't manipulate the conditions provided by fieldmanager/1

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -106,9 +111,16 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
 	}
 
 	// the code below this point triggers status updates, unify status update handling in defer
-	statusUpdates := []v1helpers.UpdateStatusFunc{}
+	available := applyoperatorv1.OperatorCondition().
+		WithType("WellKnownAvailable").
+		WithStatus(operatorv1.ConditionFalse)
+	progressing := applyoperatorv1.OperatorCondition().
+		WithType(common.ControllerProgressingConditionName(controllerName)).
+		WithStatus(operatorv1.ConditionTrue)
```
p0lyn0mial, line 119: This condition will be changed when the the code enters line `138`. It might be changed from `progressing=false` to `progressing=true`, not ideal. Do we care ?

I'm just wondering if, in this case, it would be better to read the previously applied condition and use it.
Thoughts ?

pkg/controllers/readiness/wellknown_ready_controller.go
p0lyn0mial: @bertinatto are you willing to check the controls that have already been changed and make sure they don't have this issue?

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -117,47 +131,48 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
...
+			WithStatus(operatorv1.ConditionFalse).
+			WithReason("PrereqsNotReady").
+			WithMessage(err.Error())
+		progressing = progressing.
+			WithStatus(operatorv1.ConditionTrue)
+		return err
 	}
 	if err != nil {
+		available = available.
+			WithStatus(operatorv1.ConditionFalse).
```
deads2k, line 144: followup: use unknown

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -117,47 +131,48 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
...
+			WithReason("SyncError").
+			WithMessage(err.Error())
+		progressing = progressing.
+			WithStatus(operatorv1.ConditionTrue)
```
deads2k, line 148: followup: use unknown

deads2k:
required followups, but not fatal

/lgtm
/approve

pkg/controllers/readiness/wellknown_ready_controller.go
bertinatto: https://github.com/openshift/cluster-authentication-operator/pull/716

pkg/controllers/readiness/wellknown_ready_controller.go
bertinatto: https://github.com/openshift/cluster-authentication-operator/pull/716

PR #707, opened 2024-10-03
NO-JIRA: Fix templates config test

liouk:
Due to the [enhanced routing patterns](https://tip.golang.org/doc/go1.22#enhanced_routing_patterns) of go 1.22, and because with multiple providers the provider's name is used as part of the login url (`/login/<provider-name>`), it is now impossible to use provider names that include spaces.

This PR fixes a test that uses spaces in its provider name.


ibihim:
I will try to add a unit test to the oauth-server.

/lgtm

PR #710, opened 2024-10-03
API-1835: demonstration mutation recording client

deads2k:
Now `make && ./authentication-operator apply-configuration --input-dir=/home/deads/Downloads/must-gather-01 --output-dir=../test-output` works.

Needs a couple MOM updates.  looks to be working.  Kind of amazing.

Need to re-align up to the real control loops.

/cc @p0lyn0mial @bertinatto


pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+func TestApplyConfiguration(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	mutationTrackingHTTPClient := manifestclient.NewHTTPClient("/home/deads/Downloads/must-gather-01")
```
p0lyn0mial, line 34: does it mean we will produce (the input dir) a directory structure similar to a must-gather?

pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
```diff
@@ -0,0 +1,140 @@
...
+		configInformers,
+		resourceSyncer,
+		enabledClusterCapabilities,
+		eventRecorder,
+	)
+	configInformers.Start(ctx.Done())
+	kubeInformersForNamespaces.Start(ctx.Done())
+	dynamicInformers.Start(ctx.Done())
+
+	// wait for cache sync (nearly instant from disk)
```
p0lyn0mial, line 98: why not to use `HasSynced` ?

pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"k8s.io/client-go/rest"
+)
+
+func TestApplyConfiguration(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	mutationTrackingHTTPClient := manifestclient.NewHTTPClient("/home/deads/Downloads/must-gather-01")
+
+	kubeClient, err := kubernetes.NewForConfigAndClient(&rest.Config{}, mutationTrackingHTTPClient.GetHTTPClient())
```
p0lyn0mial, line 36: :bmage

pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	)
+	configInformers.Start(ctx.Done())
+	kubeInformersForNamespaces.Start(ctx.Done())
+	dynamicInformers.Start(ctx.Done())
+
+	// wait for cache sync (nearly instant from disk)
+	time.Sleep(1 * time.Second)
+
+	syncCtx := factory.NewSyncContext("test-sync-context", eventRecorder)
+	syncErr := configObserver.Sync(ctx, syncCtx)
```
p0lyn0mial, line 102: do we need to expose `SyncOnce` method on the factory ?

pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
```diff
@@ -0,0 +1,171 @@
...
+	configClient, err := configclient.NewForConfigAndClient(&rest.Config{}, mutationTrackingHTTPClient.GetHTTPClient())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	now, err := time.Parse(time.RFC3339, "2006-01-02T15:04:05Z")
+	if err != nil {
+		t.Fatal(err)
+	}
+	v1helpers.Clock = clocktesting.NewFakeClock(now)
```
deads2k, line 55: We're going to need an authoritative FakeClock to set for operators dealing with time unless we want to pass around fake clocks everywhere.  Thoughts?

Without this, conditions always come back with different transition times and that transition time is actually important to be able to predict.  Thinking this through, we'll need to have different times for each different test.  That suggests to me that we will *will* need to plumb the fake.Clock through per instance. Counter proposals?

pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
deads2k: > do we need to expose `SyncOnce` method on the factory ?

I think we want something slightly different. We want something without a reference to the queue and without a key.

pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
deads2k: > does it mean we will produce (the input dir) a directory structure similar to a must-gather?

Yes. It's an interchange format we have tooling for already.

pkg/controllers/configobservation/configobservercontroller/apply_configuration_test.go
deads2k: it didn't exist on all our informer types. I'm forcing that into library-go now.

deads2k:
/hold cancel

As a basic starting point that gets us started, I think this is ready.

pkg/cmd/mom/apply_configuration_command.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+)
+
+func NewApplyConfigurationCommand(streams genericiooptions.IOStreams) *cobra.Command {
+	return libraryapplyconfiguration.NewApplyConfigurationCommand(RunApplyConfiguration, streams)
+}
+
+func RunApplyConfiguration(ctx context.Context, input libraryapplyconfiguration.ApplyConfigurationInput) (libraryapplyconfiguration.AllDesiredMutationsGetter, error) {
+	// TODO initialize dynamic clients, informers, operator clients, and kubeclients from the input to demonstrate.
```
p0lyn0mial, line 19: aren't you creating the clients in the `operator.CreateOperatorInputFromMOM` method ?

pkg/cmd/mom/apply_configuration_command.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/openshift/multi-operator-manager/pkg/library/libraryapplyconfiguration"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericiooptions"
+)
+
+func NewApplyConfigurationCommand(streams genericiooptions.IOStreams) *cobra.Command {
+	return libraryapplyconfiguration.NewApplyConfigurationCommand(RunApplyConfiguration, streams)
+}
+
+func RunApplyConfiguration(ctx context.Context, input libraryapplyconfiguration.ApplyConfigurationInput) (libraryapplyconfiguration.AllDesiredMutationsGetter, error) {
```
p0lyn0mial, line 18: why not to make it a private function ?

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+	)
+
+	return &AuthenticationOperatorInput{
+		kubeClient:                   kubeClient,
+		configClient:                 configClient,
+		operatorClient:               operatorClient,
+		authenticationOperatorClient: authenticationOperatorClient,
+		eventRecorder:                eventRecorder,
+		informers: []SimplifiedInformerFactory{
+			DynamicInformerFactoryAdapter(dynamicInformers), // we don't share the dynamic informers, but we only want to start when requested
```
p0lyn0mial, line 149: why not to move `DynamicInformerFactoryAdapter` to a common place ? I think it will be needed for all operators, right?

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+	for _, informer := range a.informers {
+		informer.Start(ctx)
+	}
+	for _, informer := range a.informers {
+		informer.WaitForCacheSync(ctx)
+	}
+	return nil
+}
+
+func (a authenticationOperatorStarter) Start(ctx context.Context) error {
```
p0lyn0mial, line 89: what is the purpose of this method?

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+
+type OperatorStarter interface {
+	RunOnce(ctx context.Context) error
+	Start(ctx context.Context) error
+}
+
+type authenticationOperatorStarter struct {
+	eventRecorder events.Recorder
+
+	informers        []SimplifiedInformerFactory
```
p0lyn0mial, line 57: informers or factories ?

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+	for _, controllerRunOnceFn := range a.controllerRunOnceFns {
+		// TODO add timeout.
+		syncCtx := factory.NewSyncContext("run-once-sync-context", a.eventRecorder)
+		err := controllerRunOnceFn(ctx, syncCtx)
+		errs = append(errs, err)
+	}
+	return errors.Join(errs...)
+}
+
+func (a authenticationOperatorStarter) startInformers(ctx context.Context) error {
```
p0lyn0mial, line 79: change the signature of this method, this method doesn't return any err

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+	"github.com/openshift/multi-operator-manager/pkg/library/libraryapplyconfiguration"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/utils/clock"
+)
+
+type AuthenticationOperatorInput struct {
```
p0lyn0mial, line 34: why not to make it private?

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+		v1helpers.CachedConfigMapGetter(authOperatorInput.kubeClient.CoreV1(), kubeInformersForNamespaces),
+		authOperatorInput.eventRecorder,
+	)
+	ret.controllerRunFns = append(ret.controllerRunFns, resourceSyncer.Run)
+	ret.controllerRunOnceFns = append(ret.controllerRunOnceFns, resourceSyncer.Sync)
+
+	return ret, nil
+}
+
+type GeneratedInformerFactory interface {
```
p0lyn0mial, line 255: why not to move `GeneratedInformerFactory` to a common place ? I think it will be needed for all operators, right?

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+		operatorClient:               operatorClient,
+		authenticationOperatorClient: authenticationOperatorClient,
+		eventRecorder:                eventRecorder,
+		informers: []SimplifiedInformerFactory{
+			DynamicInformerFactoryAdapter(dynamicInformers), // we don't share the dynamic informers, but we only want to start when requested
+		},
+	}, nil
+}
+
+func CreateControllerInputFromControllerContext(ctx context.Context, controllerContext *controllercmd.ControllerContext) (*AuthenticationOperatorInput, error) {
```
p0lyn0mial, line 154: `ctx` is not used.

p0lyn0mial, line 154: is this method actually used ?

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+	}, nil
+}
+
+func CreateOperatorStarter(ctx context.Context, authOperatorInput *AuthenticationOperatorInput) (OperatorStarter, error) {
+	ret := &authenticationOperatorStarter{
+		eventRecorder: authOperatorInput.eventRecorder,
+		informers:     append([]SimplifiedInformerFactory{}, authOperatorInput.informers...),
+	}
+	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(
```
p0lyn0mial, line 210: are all these namespaces required? This could be our chance to start small and add only what is needed.

pkg/cmd/mom/apply_configuration_command.go
p0lyn0mial, line 18: have you considered returning a type that would implement `OperatorStarter`? If we could do that, we would have more control over `RunOnce`.

pkg/operator/replacement_starter.go
```diff
@@ -0,0 +1,323 @@
...
+
+func (a authenticationOperatorStarter) RunOnce(ctx context.Context) error {
+	if err := a.startInformers(ctx); err != nil {
+		return err
+	}
+
+	errs := []error{}
+
+	for _, controllerRunOnceFn := range a.controllerRunOnceFns {
+		// TODO add timeout.
```
p0lyn0mial, line 71: why not to use the `ctx` ?

pkg/operator/replacement_starter.go
deads2k: doing this in https://github.com/openshift/multi-operator-manager/pull/12 and will revendor

pkg/operator/replacement_starter.go
deads2k: doing this in https://github.com/openshift/multi-operator-manager/pull/12 and will revendor and update here

pkg/operator/replacement_starter.go
deads2k: will rename

pkg/operator/replacement_starter.go
deads2k: > change the signature of this method, this method doesn't return any err

This one I think we want to keep to allow flexibility for operator authors

pkg/operator/replacement_starter.go
deads2k: > `ctx` is not used.

always pass a ctx

pkg/operator/replacement_starter.go
deads2k: > are all these namespaces required? This could be our chance to start small and add only what is needed.

based on my read, I think so

pkg/cmd/mom/apply_configuration_command.go
deads2k: > have you considered returning a type that would implement `OperatorStarter`? If we could do that, we would have more control over `RunOnce`.

doing this in https://github.com/openshift/multi-operator-manager/pull/12 and will revendor

deads2k:
holy crap that worked.

bertinatto (commented):
/lgtm

p0lyn0mial:
xref: https://github.com/openshift/library-go/pull/1819

PR #713, opened 2024-10-08
AUTH-541: OIDC structured auth config

liouk:
This PR adds a controller behind the `ExternalOIDC` feature gate that tracks the auth CR, and when auth type is configured to be OIDC, it:
- creates a structured auth config object based on the auth CR and validates it
- serializes it into JSON and stores it into a configmap
- syncs that configmap into `openshift-config-managed`, where it will be picked up by the KAS-o and synced into a static file and passed on to the KAS pods

KAS-o counterpart PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1760

Enhancement: https://github.com/openshift/enhancements/pull/1632

### DIFF
--- a/pkg/controllers/externaloidc/externaloidc_controller.go
+++ b/pkg/controllers/externaloidc/externaloidc_controller.go
@@ -1,0 +1,337 @@
+package externaloidc
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/retry"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"golang.org/x/net/http/httpproxy"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apiserverv1beta1 "k8s.io/apiserver/pkg/apis/apiserver/v1beta1"
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	configNamespace                 = "openshift-config"
+	managedNamespace                = "openshift-config-managed"
+	targetAuthConfigCMName          = "auth-config"
+	authConfigDataKey               = "auth-config.json"
+	oidcDiscoveryEndpointPath       = "/.well-known/openid-configuration"
+	kindAuthenticationConfiguration = "AuthenticationConfiguration"
+)
+
+type externalOIDCController struct {
+	name            string
+	eventName       string
+	authLister      configv1listers.AuthenticationLister
+	configMapLister corev1listers.ConfigMapLister
+	configMaps      corev1client.ConfigMapsGetter
+}
+
+func NewExternalOIDCController(
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	configInformer configinformers.SharedInformerFactory,
+	operatorClient v1helpers.OperatorClient,
+	configMaps corev1client.ConfigMapsGetter,
+	recorder events.Recorder,
+) factory.Controller {
+
+	c := &externalOIDCController{
+		name:      "ExternalOIDCController",
+		eventName: "external-oidc-controller",
+
+		authLister:      configInformer.Config().V1().Authentications().Lister(),
+		configMapLister: kubeInformersForNamespaces.ConfigMapLister(),
+		configMaps:      configMaps,
+	}
+
+	return factory.New().WithInformers(
+		// track openshift-config for changes to the provider's CA bundle
+		kubeInformersForNamespaces.InformersFor(configNamespace).Core().V1().ConfigMaps().Informer(),
+		// track auth resource
+		configInformer.Config().V1().Authentications().Informer(),
+	).WithFilteredEventsInformers(
+		// track openshift-config-managed/auth-config cm in case it gets changed externally
+		factory.NamesFilter(targetAuthConfigCMName),
+		kubeInformersForNamespaces.InformersFor(managedNamespace).Core().V1().ConfigMaps().Informer(),
+	).WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		ToController(c.name, recorder.WithComponentSuffix(c.eventName))
+}
+
+func (c *externalOIDCController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	auth, err := c.authLister.Get("cluster")
+	if err != nil {
+		return fmt.Errorf("could not get authentication/cluster: %v", err)
+	}
+
+	if auth.Spec.Type != configv1.AuthenticationTypeOIDC {
+		// auth type is "IntegratedOAuth", "" or "None"; delete structured auth configmap if it exists
+		return c.deleteAuthConfig(ctx, syncCtx)
+	}
+
+	authConfig, err := c.generateAuthConfig(*auth)
+	if err != nil {
+		return err
+	}
+
+	expectedApplyConfig, err := getExpectedApplyConfig(*authConfig)
+	if err != nil {
+		return err
+	}
+
+	existingApplyConfig, err := c.getExistingApplyConfig()
+	if err != nil {
+		return err
+	}
+
+	if existingApplyConfig != nil && equality.Semantic.DeepEqual(existingApplyConfig.Data, expectedApplyConfig.Data) {
+		return nil
+	}
+
+	if err := validateAuthConfig(*authConfig); err != nil {
+		return fmt.Errorf("auth config validation failed: %v", err)
+	}
+
+	if _, err := c.configMaps.ConfigMaps(managedNamespace).Apply(ctx, expectedApplyConfig, metav1.ApplyOptions{FieldManager: c.name, Force: true}); err != nil {
+		return fmt.Errorf("could not apply changes to auth configmap %s/%s: %v", managedNamespace, targetAuthConfigCMName, err)
+	}
+
+	syncCtx.Recorder().Eventf(c.eventName, "Synced auth configmap %s/%s", managedNamespace, targetAuthConfigCMName)
+
+	return nil
+}
+
+// deleteAuthConfig checks if the auth config ConfigMap exists in the managed namespace, and deletes it
+// if it does; it returns an error if it encounters one.
+func (c *externalOIDCController) deleteAuthConfig(ctx context.Context, syncCtx factory.SyncContext) error {
+	if _, err := c.configMapLister.ConfigMaps(managedNamespace).Get(targetAuthConfigCMName); errors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	if err := c.configMaps.ConfigMaps(managedNamespace).Delete(ctx, targetAuthConfigCMName, metav1.DeleteOptions{}); err == nil {
+		syncCtx.Recorder().Eventf(c.eventName, "Removed auth configmap %s/%s", managedNamespace, targetAuthConfigCMName)
+
+	} else if !apierrors.IsNotFound(err) {
+		return fmt.Errorf("could not delete existing configmap %s/%s: %v", managedNamespace, targetAuthConfigCMName, err)
+	}
+
+	return nil
+}
+
+// generateAuthConfig creates a structured JWT AuthenticationConfiguration for OIDC
+// from the configuration found in the authentication/cluster resource.
+func (c *externalOIDCController) generateAuthConfig(auth configv1.Authentication) (*apiserverv1beta1.AuthenticationConfiguration, error) {
+	authConfig := apiserverv1beta1.AuthenticationConfiguration{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       kindAuthenticationConfiguration,
+			APIVersion: apiserverv1beta1.ConfigSchemeGroupVersion.String(),
+		},
+	}
+
+	for _, provider := range auth.Spec.OIDCProviders {
+		jwt := apiserverv1beta1.JWTAuthenticator{
+			Issuer: apiserverv1beta1.Issuer{
+				URL:                 provider.Issuer.URL,
+				AudienceMatchPolicy: apiserverv1beta1.AudienceMatchPolicyMatchAny,
+			},
+			ClaimMappings: apiserverv1beta1.ClaimMappings{
+				Username: apiserverv1beta1.PrefixedClaimOrExpression{
+					Claim: provider.ClaimMappings.Username.Claim,
+				},
+				Groups: apiserverv1beta1.PrefixedClaimOrExpression{
+					Claim:  provider.ClaimMappings.Groups.Claim,
+					Prefix: &provider.ClaimMappings.Groups.Prefix,
+				},
+			},
+		}
+
+		for _, aud := range provider.Issuer.Audiences {
+			jwt.Issuer.Audiences = append(jwt.Issuer.Audiences, string(aud))
+		}
+
+		if len(provider.Issuer.CertificateAuthority.Name) > 0 {
+			caConfigMap, err := c.configMapLister.ConfigMaps(configNamespace).Get(provider.Issuer.CertificateAuthority.Name)
+			if err != nil {
+				return nil, fmt.Errorf("could not retrieve auth configmap %s/%s to check CA bundle: %v", configNamespace, provider.Issuer.CertificateAuthority.Name, err)
+			}
+
+			caData, ok := caConfigMap.Data["ca-bundle.crt"]
+			if !ok || len(caData) == 0 {
+				return nil, fmt.Errorf("configmap %s/%s key \"ca-bundle.crt\" missing or empty", configNamespace, provider.Issuer.CertificateAuthority.Name)
+			}
+
+			jwt.Issuer.CertificateAuthority = caData
+		}
+
+		switch provider.ClaimMappings.Username.PrefixPolicy {
+
+		case configv1.Prefix:
+			if provider.ClaimMappings.Username.Prefix == nil {
+				return nil, fmt.Errorf("nil username prefix while policy expects one")
+			} else {
+				jwt.ClaimMappings.Username.Prefix = &provider.ClaimMappings.Username.Prefix.PrefixString
+			}
+
+		case configv1.NoPrefix:
+			jwt.ClaimMappings.Username.Prefix = ptr.To("")
+
+		case configv1.NoOpinion:
+			prefix := ""
+			if provider.ClaimMappings.Username.Claim != "email" {
+				prefix = provider.Issuer.URL + "#"
+			}
+			jwt.ClaimMappings.Username.Prefix = &prefix
+
+		default:
+			return nil, fmt.Errorf("invalid username prefix policy: %s", provider.ClaimMappings.Username.PrefixPolicy)
+		}
+
+		for i, rule := range provider.ClaimValidationRules {
+			if rule.RequiredClaim == nil {
+				return nil, fmt.Errorf("empty validation rule at index %d", i)
+			}
+
+			jwt.ClaimValidationRules = append(jwt.ClaimValidationRules, apiserverv1beta1.ClaimValidationRule{
+				Claim:         rule.RequiredClaim.Claim,
+				RequiredValue: rule.RequiredClaim.RequiredValue,
+			})
+		}
+
+		authConfig.JWT = append(authConfig.JWT, jwt)
+	}
+
+	return &authConfig, nil
+}
+
+// getExpectedApplyConfig serializes the input authConfig into JSON and creates an apply configuration
+// for a configmap with the serialized authConfig in the right key.
+func getExpectedApplyConfig(authConfig apiserverv1beta1.AuthenticationConfiguration) (*corev1ac.ConfigMapApplyConfiguration, error) {
+	authConfigBytes, err := json.Marshal(authConfig)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal auth config into JSON: %v", err)
+	}
+
+	expectedCMApplyConfig := corev1ac.ConfigMap(targetAuthConfigCMName, managedNamespace).
+		WithData(map[string]string{
+			authConfigDataKey: string(authConfigBytes),
+		})
+
+	return expectedCMApplyConfig, nil
+}
+
+// getExistingApplyConfig checks if an authConfig configmap already exists, and returns an apply configuration
+// that represents it if it does; it returns nil otherwise.
+func (c *externalOIDCController) getExistingApplyConfig() (*corev1ac.ConfigMapApplyConfiguration, error) {
+	existingCM, err := c.configMapLister.ConfigMaps(managedNamespace).Get(targetAuthConfigCMName)
+	if apierrors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
+		return nil, fmt.Errorf("could not retrieve auth configmap %s/%s to check data before sync: %v", managedNamespace, targetAuthConfigCMName, err)
+	}
+
+	existingCMApplyConfig, err := corev1ac.ExtractConfigMap(existingCM, c.name)
+	if err != nil {
+		return nil, fmt.Errorf("could not extract ConfigMap apply configuration: %v", err)
+	}
+
+	return existingCMApplyConfig, nil
+}
+
+// validateAuthConfig performs validations that are not done at the server-side,
+// including validation that the provided CA cert (or system CAs if not specified) can be used for
+// TLS cert verification.
+func validateAuthConfig(auth apiserverv1beta1.AuthenticationConfiguration) error {
+	for _, jwt := range auth.JWT {
+		var caCertPool *x509.CertPool
+		var err error
+		if len(jwt.Issuer.CertificateAuthority) > 0 {
+			caCertPool, err = cert.NewPoolFromBytes([]byte(jwt.Issuer.CertificateAuthority))
+			if err != nil {
+				return fmt.Errorf("issuer CA is invalid: %v", err)
+			}
+		}
+
+		// make sure we can access the issuer with the given cert pool (system CAs used if pool is empty)
+		if err := validateCACert(jwt.Issuer.URL, caCertPool); err != nil {
+			certMessage := "using the specified CA cert"
+			if caCertPool == nil {
+				certMessage = "using the system CAs"
+			}
+			return fmt.Errorf("could not validate IDP URL %s: %v", certMessage, err)
+		}
+	}
+
+	return nil
+}
+
+// validateCACert makes a request to the provider's well-known endpoint using the
+// specified CA cert pool to validate that the certs in the pool match the host.
+func validateCACert(hostURL string, caCertPool *x509.CertPool) error {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{RootCAs: caCertPool},
+			Proxy: func(*http.Request) (*url.URL, error) {
+				if proxyConfig := httpproxy.FromEnvironment(); len(proxyConfig.HTTPSProxy) > 0 {
+					return url.Parse(proxyConfig.HTTPSProxy)
+				}
+				return nil, nil
+			},
+		},
+		Timeout: 5 * time.Second,
+	}
+
+	wellKnown := strings.TrimSuffix(hostURL, "/") + oidcDiscoveryEndpointPath
+	req, err := http.NewRequest(http.MethodGet, wellKnown, nil)
+	if err != nil {
+		return fmt.Errorf("could not create well-known HTTP request: %v", err)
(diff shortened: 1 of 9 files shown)


pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,363 @@
...
+	syncCtx.Recorder().Eventf(c.eventName, "Synced auth configmap %s/%s", targetConfigMapNamespace, targetAuthConfigCMName)
+
+	return nil
+}
+
+// generateAuthConfig creates a structured JWT AuthenticationConfiguration for OIDC
+// from the configuration found in the authentication/cluster resource
+func (c *externalOIDCController) generateAuthConfig(auth configv1.Authentication) (*apiserverv1beta1.AuthenticationConfiguration, error) {
+	authConfig := apiserverv1beta1.AuthenticationConfiguration{
+		TypeMeta: metav1.TypeMeta{
```
ibihim, line 154: So, by side-stepping api-machinery and using `json.Marshal` we need to set it by hand. Do we do that in other places as well or do we leverage the serializer?

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,363 @@
...
+func validateCACert(hostURL url.URL, caCertPool *x509.CertPool) error {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{RootCAs: caCertPool},
+		},
+		Timeout: 5 * time.Second,
+	}
+
+	wellKnown := strings.TrimSuffix(hostURL.String(), "/") + "/.well-known/openid-configuration"
+	req, err := http.NewRequest("GET", wellKnown, nil)
```
ibihim, line 335: nit: `http.MethodGet`.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,363 @@
...
+// specified CA cert pool to validate that the certs in the pool match the host
+func validateCACert(hostURL url.URL, caCertPool *x509.CertPool) error {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{RootCAs: caCertPool},
+		},
+		Timeout: 5 * time.Second,
+	}
+
+	wellKnown := strings.TrimSuffix(hostURL.String(), "/") + "/.well-known/openid-configuration"
```
ibihim, line 334: This path is a `const` defined by the OIDC spec: `"/.well-known/openid-configuration"`?

pkg/operator/replacement_starter.go
```diff
@@ -318,5 +320,24 @@ func CreateOperatorStarter(ctx context.Context, authOperatorInput *authenticatio
...
 	ret.ControllerNamedRunOnceFns = append(ret.ControllerNamedRunOnceFns, oauthAPIServerRunOnceFns...)
 
+	// OIDC
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		status.VersionForOperatorFromEnv(), "0.0.1-snapshot",
+		informerFactories.operatorConfigInformer.Config().V1().ClusterVersions(),
+		informerFactories.operatorConfigInformer.Config().V1().FeatureGates(),
+		authOperatorInput.eventRecorder,
+	)
+	go featureGateAccessor.Run(ctx)
```
ibihim, line 330: Not sure about this one in the context of mom.

ibihim (commented):
First look

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,363 @@
...
+
+			caData, ok := caConfigMap.Data["ca-bundle.crt"]
+			if !ok || len(caData) == 0 {
+				return nil, fmt.Errorf("configmap %s/%s key \"ca-bundle.crt\" missing or empty", caBundleConfigMapNamespace, provider.Issuer.CertificateAuthority.Name)
+			}
+
+			jwt.Issuer.CertificateAuthority = caData
+		}
+
+		switch provider.ClaimMappings.Username.PrefixPolicy {
```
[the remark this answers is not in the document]
liouk: The `UsernamePrefixPolicy` type won't allow any other value: https://github.com/liouk/cluster-authentication-operator/blob/de4ac96b4cd7c9f8de3a966d69ab31035403decb/vendor/github.com/openshift/api/config/v1/types_authentication.go#L410

If you prefer, for completeness, I can add a default with an error.

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: That's right, this is defined here: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig

This is also what the `kube-apiserver` is using when creating an instance of the authenticator. SeeL
- https://github.com/openshift/kubernetes-apiserver/blob/7742d76eabcb5f3d49e995b91b499b478d4f2f2e/plugin/pkg/authenticator/token/oidc/oidc.go#L278
- https://github.com/coreos/go-oidc/blob/v2.2.1/oidc.go#L114

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: `apiserverv1beta1.AuthenticationConfiguration` won't go via an API call, but instead it will be serialized into a json file and read back by the `kube-apiserver`, so for the moment there's no apimachinery to handle this AFAIK.

pkg/operator/replacement_starter.go
liouk: That's a good point, I will double check this and report back.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	apiserverv1beta1 "k8s.io/apiserver/pkg/apis/apiserver/v1beta1"
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	caBundleConfigMapNamespace = "openshift-config"
```
ibihim, line 37: Nit / Personal viewpoint - I like the naming in the kube-apiserver-operator better:

```go
	configNamespace       = "openshift-config"
	managedNamespace      = "openshift-config-managed"
```

It better describes what it is, rather how it is being used.

xingxingxia:
Pre-merge test found an issue, filed `https://issues.redhat.com/browse/OCPBUGS-44483` for better tracking per existing pre-merge test workflow.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	auth, err := c.authLister.Get("cluster")
+	if err != nil {
+		return fmt.Errorf("could not get authentication/cluster: %v", err)
+	}
+
+	if auth.Spec.Type != configv1.AuthenticationTypeOIDC {
+		// auth type is "IntegratedOAuth", "" or "None"; delete structured auth configmap if it exists
+		if err := c.configMaps.ConfigMaps(managedNamespace).Delete(ctx, targetAuthConfigCMName, metav1.DeleteOptions{}); err == nil {
```
ibihim, line 107: So at this line, we are in `!featureGates.Enabled(features.FeatureGateExternalOIDC) == false`, which means that the features is enabled. Now we check if it Auth is set to external OIDC and if it is, we attempt to delete the resource.

Wouldn't it be more beneficial to look up in the Informer Cache if it exists, before attempting to delete it? That way we could save a call to the apiserver? I mean this controller will be triggered a couple of times.

pkg/controllers/externaloidc/externaloidc_controller.go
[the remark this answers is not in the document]
ibihim: Ok, so there will be a validation error, when someone tries to create one with another value?

pkg/controllers/externaloidc/externaloidc_controller.go
ibihim: It has a [register.go](https://github.com/openshift/cluster-authentication-operator/blob/master/vendor/k8s.io/apiserver/pkg/apis/apiserver/v1beta1/register.go#L55). We would need to create a runtime.Scheme and add the object to the scheme, to then be able to use the codec for serialization and deserialization? I am not sure, I am not an expert in api machinery topics and can't tell if it is "the better way" or if it adds too much complexity.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,380 @@
...
+		return connErr == nil, connErr
+	})
+	cancel()
+	if connErr != nil {
+		return fmt.Errorf("GET well-known error: %v", connErr)
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
```
ibihim, line 371: Technically we only need to read the body, if the `resp.StatusCode` is NOT `http.StatusOK`, so I would only do it in the err case.

I never used it, but we might be even fine with http.MethodHEAD? Not sure, we could just use get and read the Body on err.

liouk: >Technically we only need to read the body, if the `resp.StatusCode` is NOT `http.StatusOK`, so I would only do it in the err case.

I wrote it like this because depending on the status code, sometimes the body can contain useful information about the respective code.

>I never used it, but we might be even fine with http.MethodHEAD? Not sure, we could just use get and read the Body on err.

I chose GET because:
- this is the one defined in the discovery endpoint spec: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest
- this is the one that KAS uses to do the same validation, so we're preserving the behavior

HEAD is risky in the sense that it might be blocked on some servers as it's not defined in the spec.

pkg/controllers/externaloidc/externaloidc_controller.go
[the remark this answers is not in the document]
liouk: Exactly, this will be rejected by the API when someone will try to edit the authentication/cluster resource.

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: Nice, wasn't very familiar with schemes, TIL :+1: Pushing a fix.

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: Good point, fixed; reading body only if we need to put it in an error message.

pkg/controllers/externaloidc/externaloidc_controller.go
ibihim: Now, that we use the codec, we can drop the TypeMeta, right?

ibihim (commented):
/lgtm

pkg/operator/starter.go
```diff
@@ -671,6 +674,52 @@ func prepareOauthAPIServerOperator(
...
+	var featureGates featuregates.FeatureGate
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		featureGates, _ = featureGateAccessor.CurrentFeatureGates()
+	case <-time.After(1 * time.Minute):
+		return nil, nil, fmt.Errorf("timed out waiting for FeatureGate detection")
+	}
+
+	if !featureGates.Enabled(features.FeatureGateExternalOIDC) {
+		return nil, nil, nil
```
benluddy, line 702: Do you want to set up a child context with cancellation so that the feature gate accessor that is launched above can terminate?

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	authConfig, err := c.generateAuthConfig(*auth)
+	if err != nil {
+		return err
+	}
+
+	encoded, err := runtime.Encode(codecs.EncoderForVersion(serializerInfo.Serializer, apiserverv1beta1.ConfigSchemeGroupVersion), authConfig)
+	if err != nil {
+		return fmt.Errorf("could not marshal auth config into JSON: %v", err)
+	}
+	authConfigJSON := strings.TrimSpace(string(encoded))
```
benluddy, line 128: I know the JSON serializer appends a trailing newline, but what problem was that causing?

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
+	"k8s.io/utils/ptr"
+)
+
+var (
+	cfgScheme         = runtime.NewScheme()
+	codecs            = serializer.NewCodecFactory(cfgScheme, serializer.EnableStrict)
```
benluddy, line 38: Nit: The EnableStrict option isn't really doing anything for you since you are only building an encoder from this.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	serializerInfo, _ = runtime.SerializerInfoForMediaType(codecs.SupportedMediaTypes(), runtime.ContentTypeJSON)
```
benluddy, line 39: Nit: Best to check the second return value and panic on `false` in case this should somehow break. That would be a clearer failure than a panic later inside of `runtime.Encode`.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,370 @@
...
+
+		return nil
+	}
+
+	authConfig, err := c.generateAuthConfig(*auth)
+	if err != nil {
+		return err
+	}
+
+	encoded, err := runtime.Encode(codecs.EncoderForVersion(serializerInfo.Serializer, apiserverv1beta1.ConfigSchemeGroupVersion), authConfig)
```
benluddy, line 124: I think this is fine, but I'm also unsure what benefit setting up and using a CodecFactory is providing. We're not converting or defaulting anything and we always want it to use JSON. May as well use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/json#Marshal?

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,370 @@
...
+		return nil
+	}
+
+	errList := validateAuthenticationConfiguration(*authConfig)
+	if len(errList) > 0 {
+		return fmt.Errorf("auth config validation failed: %v", errList)
+	}
+
+	cm := corev1ac.ConfigMap(targetAuthConfigCMName, managedNamespace).WithData(map[string]string{authConfigDataKey: authConfigJSON})
+	if _, err := c.configMaps.ConfigMaps(managedNamespace).Apply(ctx, cm, metav1.ApplyOptions{FieldManager: c.name}); err != nil {
```
benluddy, line 145: Is another field manager going to be writing to the same configmap? If not, it is probably reasonable to set `Force: true` to allow stomping conflicts.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	if existingCM != nil && existingCM.Data[authConfigDataKey] == authConfigJSON {
+		return nil
+	}
+
+	errList := validateAuthenticationConfiguration(*authConfig)
+	if len(errList) > 0 {
+		return fmt.Errorf("auth config validation failed: %v", errList)
+	}
+
+	cm := corev1ac.ConfigMap(targetAuthConfigCMName, managedNamespace).WithData(map[string]string{authConfigDataKey: authConfigJSON})
```
benluddy, line 144: We've been trying to reduce no-op Apply requests by extracting the current apply configuration from the local informer's cache with https://pkg.go.dev/k8s.io/client-go/applyconfigurations/core/v1#ExtractConfigMap and doing a `apiequality.Semantic.DeepEqual` between them first. I haven't heard of any issues arising from that approach yet, does it make sense to do it here to avoid making a write on every resync?

pkg/controllers/externaloidc/externaloidc_controller.go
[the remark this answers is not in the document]
benluddy: What if a new valid option is added in the future? Can there be a period of time during an upgrade from N to N+1 where there is a cluster-authentication-operator at N and a CRD at N+1? I would include a default case to avoid all doubt.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	return &authConfig, nil
+}
+
+// validateAuthenticationConfiguration runs as many validations as possible
+// on the AuthenticationConfiguration in order to catch validation errors early in the process
+// instead of waiting for the KAS pods to roll out and consume the configuration
+func validateAuthenticationConfiguration(auth apiserverv1beta1.AuthenticationConfiguration) (errs []error) {
+	// TODO currently validations from k8s.io/apiserver/pkg/apis/apiserver/validation cannot be used here
+	// since they aren't defined for the beta type; once the feature goes out of beta, we should replace
+	// this func with the upstream validations (but keep CA cert validation)
```
benluddy, line 232: They're defined for the internal type, would it be easier to convert to internal and use those? This seems to be how it is done for unserved APIs like the apiserver configuration files, I don't see any external-versioned validations there.

Also, what happens if there is a bug in the validation used by cluster-authentication-operator that causes the validation to be overly strict? Even if this changes to use the validations from k8s.io/apiserver, this can be out of sync with whatever a particular kube-apiserver was compiled against. This is a form of client-side validation, which we have been trying to move away from. Revisioned kube-apiserver rollouts should mitigate the risk of writing an invalid config here, right?

pkg/operator/starter.go
liouk: Don't we want to keep the accessor running so that it can os.Exit if the feature gates change?

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: Unit tests, mainly; I'll leave it intact here and fix the test.

pkg/controllers/externaloidc/externaloidc_controller.go
ibihim: Yes, this might be my fault. I mentioned that I am not sure if using the api machinery is better / more idiomatic than json.Marshal.

@liouk's original solution was based on json.Marshal.

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: No, the CAO must be the only one -- great point.

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: I'll revert back to using json.Marshal, this turned out to be more complicated than I thought as well :+1:

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: That's the intention of this check: https://github.com/openshift/cluster-authentication-operator/pull/713/files#diff-3c99f304cc2949488aa2fa2b8aea2d7e8ddb0c8baa42e66f128eacd7cdbda11aR135-R137
```go
	if existingCM != nil && existingCM.Data[authConfigDataKey] == authConfigJSON {
		return nil
	}
```

Do you think the `DeepEqual()` is preferable?

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: The main benefit was to serialize the data in a more structured way, e.g. without having to define the type meta manually. But it gets more complicated than what I thought, so I will revert back to using json.Marshal.

pkg/controllers/externaloidc/externaloidc_controller.go
liouk: Alright, I see the point in keeping validations at the server-side, especially with revisioned rollouts being in place.

However, the KAS pods do not really validate the CA cert, if specified. If the CA cert is not the correct one, the KAS pods will log an error but will not crash, so the rollout will be completed correctly. Therefore I'm considering keeping the CA cert validation at the CAO side, but dropping the rest. What do you think?

liouk: Pushed a fixup 20a1f729f3f38fe4ee614fc34fa3bcc92b14a0cf that demonstrates what I described above. Placing a hold until this gets squashed or dropped.

/hold

xingxingxia:
From test result perspective, based on good pre-merge test results in https://issues.redhat.com/browse/OCPBUGS-44592?focusedId=26134688&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26134688 , adding below label:
/label qe-approved

pkg/operator/starter.go
benluddy: You're right, I saw that nothing else was using it and did not make the connection that the comment at the top was saying that this all depends on the exit-on-feature-change behavior.

pkg/controllers/externaloidc/externaloidc_controller.go
everettraven: Looks like we are still using `json.Marshal` and not an encoder that utilizes the `runtime.Scheme` for encoding resources - is this still the intended direction?

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,292 @@
...
+
+	return nil
+}
+
+// generateAuthConfig creates a structured JWT AuthenticationConfiguration for OIDC
+// from the configuration found in the authentication/cluster resource
+func (c *externalOIDCController) generateAuthConfig(auth configv1.Authentication) (*apiserverv1beta1.AuthenticationConfiguration, error) {
+	authConfig := apiserverv1beta1.AuthenticationConfiguration{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AuthenticationConfiguration",
```
everettraven, line 145: Nit: If we are keeping this approach, it might be nice to make this a constant

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,292 @@
...
+			APIVersion: "apiserver.config.k8s.io/v1beta1",
```
everettraven, line 146: I would encourage importing the `SchemeGroupVersion` from the [register.go](https://github.com/openshift/cluster-authentication-operator/blob/c36b1653ea9cc3e33e7bf9242a32381da461fd26/vendor/k8s.io/apiserver/pkg/apis/apiserver/v1beta1/register.go#L29) and using it here to set the APIVersion using the `schema.GroupVersion.String()` method: https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#GroupVersion.String

pkg/controllers/externaloidc/externaloidc_controller_test.go
```diff
@@ -0,0 +1,964 @@
...
+	}
+}
+
+func TestExternalOIDCController_generateAuthConfig(t *testing.T) {
+	testServer, err := createTestServer(baseCACert, baseCAPrivateKey, nil)
+	if err != nil {
+		t.Fatalf("could not create test server: %v", err)
+	}
+	defer testServer.Close()
+	testServer.StartTLS()
```
everettraven, line 387: Do we need this test server for testing the auth config generation?

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,292 @@
...
+			return fmt.Errorf("could not validate IDP URL %s: %v", certMessage, err)
+		}
+	}
+
+	return nil
+}
+
+// validateCACert makes a request to the provider's well-known endpoint using the
+// specified CA cert pool to validate that the certs in the pool match the host
+func validateCACert(hostURL string, caCertPool *x509.CertPool) error {
```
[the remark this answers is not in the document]
everettraven: For context, I am asking because I think doing a best-effort static analysis would reduce the complexity of performing a live HTTP request and reduce other networking related points of failure for enabling the OIDC functionality.

pkg/controllers/externaloidc/externaloidc_controller_test.go
```diff
@@ -0,0 +1,964 @@
...
+			authIndexer := cache.NewIndexer(func(obj interface{}) (string, error) {
+				return "cluster", nil
+			}, cache.Indexers{})
+
+			if tt.configMapIndexer == nil {
+				tt.configMapIndexer = cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+			}
+
+			if tt.auth != nil {
+				authIndexer.Add(tt.auth)
```
everettraven, line 320: Nit: I've never seen the use of indexers for mocking out test data before, and while it makes sense, I wonder if creating mocks for the lister/getter interfaces could be easier to reason about and/or simplify the complexity of the test logic a bit?

pkg/controllers/externaloidc/externaloidc_controller_test.go
liouk: Right, that's a leftover -- CA validation used to be inside `generateAuthConfig()` :+1:

pkg/controllers/externaloidc/externaloidc_controller.go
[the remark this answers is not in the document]
liouk: We can't avoid a network call, because we need the server cert chain to validate that the CA is correct, which we don't have; we only have the CA cert that was (presumably) used to sign the server's serving cert.

While it should be possible to reduce the overhead, e.g. by doing a `tls.Dial()` instead of a full HTTP request in order to fetch the server cert and validate offline, I chose the HTTP call to the OIDC discovery endpoint because this is exactly what the KAS pods are doing when the authenticator is initialized.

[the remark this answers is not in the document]
everettraven: Ah, that makes sense - my head was still stuck in OLM land where messing with certs was we were the server :). Thanks for the clarification!

everettraven (commented):
Overall looks good. Have a handful of new comments

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,325 @@
...
+					Claim: provider.ClaimMappings.Username.Claim,
+				},
+				Groups: apiserverv1beta1.PrefixedClaimOrExpression{
+					Claim:  provider.ClaimMappings.Groups.Claim,
+					Prefix: &provider.ClaimMappings.Groups.Prefix,
+				},
+			},
+		}
+
+		if len(provider.Issuer.Audiences) > 0 {
```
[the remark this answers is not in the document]
liouk: Since there's a for loop that appends the strings, it's not needed anyways. Will remove!

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,325 @@
...
+			}
+
+			jwt.Issuer.CertificateAuthority = caData
+		}
+
+		switch provider.ClaimMappings.Username.PrefixPolicy {
+		case configv1.NoOpinion:
+			jwt.ClaimMappings.Username.Prefix = ptr.To("")
+		case configv1.NoPrefix:
+			jwt.ClaimMappings.Username.Prefix = ptr.To("-")
```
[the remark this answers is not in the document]
liouk: This is actually a great find. You're right -- we must translate `NoPrefix` to `""`. But that's not the only issue; we also need to translate `NoOpinion` to a special prefix logic.

From [here](https://github.com/openshift/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/apis/apiserver/v1beta1/types.go#L332-L336):
> `--oidc-username-prefix=""` and  `--oidc-username-claim != "email"`, prefix was `"<value of --oidc-issuer-url>#"`. For the same behavior using authentication config, set `username.prefix="<value of issuer.url>#"`

Therefore, `NoOpinion` means:
- set `username.prefix = ""`, if `username.claim == "email"`
- set `username.prefix = <issuer.URL>#` otherwise

This seems to be an issue in [hypershift](https://github.com/openshift/hypershift/blob/2f395fed4faf2d9dd21e78be6d223ae481f2a27c/control-plane-operator/controllers/hostedcontrolplane/kas/auth.go#L76-L82) as well; will look further into that.

Pushed a fix for this in 5fecaf64816da553823c33aab3095410773a78d4.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,325 @@
...
+	if err != nil {
+		return err
+	} else if expectedApplyConfig == nil {
+		// no change detected, nothing to do
+		return nil
+	}
+
+	if err := validateAuthConfig(*authConfig); err != nil {
+		return fmt.Errorf("auth config validation failed: %v", err)
+	}
```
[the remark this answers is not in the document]
liouk: As discussed offline, we decided to keep this sequence due to the fact that the validation performs a network call to the provider to grab the server certificates, which we can avoid if the two configs are equivalent.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -0,0 +1,337 @@
...
+			} else {
+				jwt.ClaimMappings.Username.Prefix = &provider.ClaimMappings.Username.Prefix.PrefixString
+			}
+
+		case configv1.NoPrefix:
+			jwt.ClaimMappings.Username.Prefix = ptr.To("")
+
+		case configv1.NoOpinion:
+			prefix := ""
+			if provider.ClaimMappings.Username.Claim == "email" {
```
everettraven, line 205: If I understood https://github.com/openshift/cluster-authentication-operator/pull/713#discussion_r1937312035 correctly, this should be set to `provider.Issuer.URL` + `#` if the username claim is _not_ `email`:
```suggestion
			if provider.ClaimMappings.Username.Claim != "email" {
```

liouk:
Fixups have been squashed and [this comment](https://github.com/openshift/cluster-authentication-operator/pull/713#issuecomment-2618879567) has been addressed as well.

/hold cancel

everettraven (approved):
/lgtm

PR #718, opened 2024-10-18
API-1835: separate apply-configuration-live from apply-configuration

deads2k:
This does the initial split of live functions (things that must reach out to external services and may or may not work on hypershift due to networking configuration), from things that don't have that problem.

Builds on top of some library-go and multi-operator-manager PRs that need to land.

### DIFF
--- a/pkg/operator/workload/sync_openshift_oauth_apiserver.go
+++ b/pkg/operator/workload/sync_openshift_oauth_apiserver.go
@@ -232,6 +232,11 @@ func (c *OAuthAPIServerWorkload) syncDeployment(ctx context.Context, operatorSpe
 	}
 	required.Spec.Replicas = masterNodeCount
 
+	// TODO MOM this call here is the one that fails.  Using the return value of ApplyDeployment to make decisions doesn't work in a MOM world.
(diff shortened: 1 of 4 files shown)


pkg/operator/workload/sync_openshift_oauth_apiserver.go
```diff
@@ -232,6 +232,11 @@ func (c *OAuthAPIServerWorkload) syncDeployment(ctx context.Context, operatorSpe
 	}
 	required.Spec.Replicas = masterNodeCount
 
+	// TODO MOM this call here is the one that fails.  Using the return value of ApplyDeployment to make decisions doesn't work in a MOM world.
+	// TODO MOM workload.NewController needs to (somehow) use the data read *after* this write, without  using data *in* this write.
+	// TODO MOM this may be possible by storing a hash of the desired content.  If the hash of the desired content matches
+	// TODO MOM then there is no new deployment required.  This would not prevent rapid writes to fields we aren't controlling.
+	// TODO MOM perhaps we actually need to find a way to request MOM inject a generation after a write?  It is an operator standard.
```
deads2k, line 239: This is sticky.  The hash probably won't work.  Making MOM aware of writing workload generation may be the only way to achieve parity.

@bertinatto

PR #719, opened 2024-10-19
OCPBUGS-43587: bindata/oauth-openshift - set httpmux121

ibihim:
## What

Set `GODEBUG=httpmuxgo121=1`,

## Why

`http.ServeMux` changed massively in behavior in Go 1.22. We allowed customers to create IDPs with whitespaces in their names and used that as their path. This will now cause a panic with the new handler behavior.

### How to reproduce

- Run oauth-server with Go >= 1.22.
- Create two htpasswd IDPs with whitespaces in their names.


liouk:
Closing in favor of https://github.com/openshift/oauth-server/pull/162.

/close

PR #722, opened 2024-10-22
NO-JIRA: add relatedResources to authentication for collection if operator crashes

deads2k:
The CVO pre-creates the relatedresoruces in status.  if the operator fails to run for some reason, this ensures we get some basic collection.  I pulled these values from a successful CI run.

/assign @liouk @ibihim


manifests/08_clusteroperator.yaml
```diff
@@ -7,6 +7,45 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     exclude.release.openshift.io/internal-openshift-hosted: "true"
 status:
+  relatedObjects:
```
p0lyn0mial, line 10: why is this list important ?

PR #726, opened 2024-10-23
API-1835: first integration-tests

deads2k:
These now run and fail.  I think we should merge the failing set so that we can create an option CI job that runs pre-merge so we can see them.

This is actually highlighting bugs in the current operator that produce unstable output, leading to excessive status writes.  Additionally, it's exposing a need to allow comparisons of TLS artifacts that don't rely on the same output each time.


pkg/operator/replacement_starter.go
```diff
@@ -59,31 +59,36 @@ type authenticationOperatorInput struct {
 const componentName = "cluster-authentication-operator"
 
 func CreateOperatorInputFromMOM(ctx context.Context, momInput libraryapplyconfiguration.ApplyConfigurationInput) (*authenticationOperatorInput, error) {
-	kubeClient, err := kubernetes.NewForConfigAndClient(&rest.Config{}, momInput.MutationTrackingClient.GetHTTPClient())
+	// TODO replace with the library-go function in https://github.com/openshift/library-go/pull/1857 once it merges
+	recommendedRESTConfig := &rest.Config{
+		QPS:   1000,
+		Burst: 10000,
+	}
```
bertinatto, line 66: https://github.com/openshift/library-go/pull/1857 already merged

bertinatto (commented):
/lgtm

PR #737, opened 2024-11-05
WIP: AUTH-544: Extend e2e test to leverage new functionality

liouk:
This PR extends the baseline e2e test to use functionality introduced by the following PRs:
- #713 
- openshift/cluster-kube-apiserver-operator#1760

Thus, it requires these PRs in order to work.

Extensions:
- lets CAO consume auth CR OIDC configuration and generate structured auth config
- adds tests for the outcome of the configuration:
  - KAS arguments
  - configmap contents
- adds tests for the validity and correctness of the access & ID OIDC tokens

This PR is set as WIP; the goal is eventually to move this from a pre-merge test to `origin`.


everettraven (commented):
Used this pass to get up to speed on the changes being made here and leave comments/questions on anything that stood out to me.

I'll make another pass tomorrow morning with more context loaded and a fresh set of eyes

test/e2e/external_oidc_test.go
```diff
@@ -187,263 +289,458 @@ func newTestClient(t *testing.T) (*testClient, error) {
...
-			tc.t.Fatalf("cleanup failed for secret '%s/%s': %v", oidcServingCA.Namespace, oidcServingCA.Name, err)
-		}
-	})
 	if err != nil {
-		return
+		return false
 	}
 
-	return
+	return true
```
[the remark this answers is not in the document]
liouk: - `requireMissing == true && err == nil`: you're right, in this case `assertMissing` should return `false`; didn't catch this as the test isn't testing the `requireMissing == true` case yet
- `requireMissing == false && err == nil`: that's expected, because `err == nil` means that the resource exists

liouk:
@everettraven pushed some changes to hopefully simplify the oauth resource assertions a bit.

test/e2e/external_oidc_test.go
```diff
@@ -187,263 +290,460 @@ func newTestClient(t *testing.T) (*testClient, error) {
...
+	return parsedAccessToken, parsedIDToken, nil
+}
+
+func (tc *testClient) setupKeycloakClient(t *testing.T, ctx context.Context, keycloakURL string) *test.KeycloakClient {
+	transport, err := rest.TransportFor(tc.kubeConfig)
+	require.NoError(t, err)
+
+	kcClient := test.KeycloakClientFor(t, transport, keycloakURL, "master")
+	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 30*time.Second, true, func(ctx context.Context) (bool, error) {
+		err := kcClient.AuthenticatePassword(oidcClientId, "", "admin", "password")
```
everettraven, line 417: We have a way to use an existing keycloak deployment using an environment variable when running this e2e test - what happens if `admin` and `password` are not the right login credentials for that existing deployment?

Should this also be configurable in the same way? 

Alternatively, should we just always stand up a new Keycloak instance for this test?

test/e2e/external_oidc_test.go
```diff
@@ -187,263 +290,460 @@ func newTestClient(t *testing.T) (*testClient, error) {
...
-			"ca-bundle.crt": string(signingKey.Data["tls.crt"]),
-		},
+	expectedAuthConfigJSON := fmt.Sprintf(`{"kind":"AuthenticationConfiguration","apiVersion":"apiserver.config.k8s.io/v1beta1","jwt":[{"issuer":{"url":"%s","certificateAuthority":"%s","audiences":[%s],"audienceMatchPolicy":"MatchAny"},"claimMappings":{"username":{"claim":"%s","prefix":"%s"},"groups":{"claim":"%s","prefix":"%s"},"uid":{}}}]}`,
+		idpURL,
+		strings.ReplaceAll(certData, "\n", "\\n"),
+		strings.Join([]string{fmt.Sprintf(`"%s"`, oidcClientId)}, ","),
+		authSpec.OIDCProviders[0].ClaimMappings.Username.Claim,
+		usernamePrefix,
+		oidcGroupsClaim,
+		oidcGroupsPrefix,
```
everettraven, line 505: Don't mix global variables and function/method parameters where you can avoid it. IMO it causes unnecessary confusion when looking at what a particular function/method does.

test/e2e/external_oidc_test.go
```diff
@@ -187,263 +290,460 @@ func newTestClient(t *testing.T) (*testClient, error) {
...
+	httpClient := &http.Client{Transport: &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}}
+
+	formData := url.Values{
+		"grant_type": []string{"password"},
+		"client_id":  []string{oidcClientId},
+		"scope":      []string{"openid"},
+		"username":   []string{user},
+		"password":   []string{password},
```
everettraven, line 679: Do we need to use an http client to post this form? Would it be reasonable to use a new keycloak client for the newly created user to get the token?

test/e2e/external_oidc_test.go
```diff
@@ -64,110 +69,208 @@ type oidcAuthResponse struct {
...
+			t.Fatalf("cleanup failed for authentication '%s' while updating object: %v", auth.Name, err)
+		}
+
+		err = test.WaitForNewKASRollout(t, testCtx, tc.operatorConfigClient.OperatorV1().KubeAPIServers(), kasOriginalRevision)
+		require.NoError(t, err, "failed to wait for KAS rollout during cleanup")
+
+		err = wait.PollUntilContextTimeout(testCtx, 10*time.Second, 5*time.Minute, true, func(ctx context.Context) (done bool, err error) {
+			return tc.validateOAuthResources(testCtx, false)
+		})
+		require.NoError(t, err, "failed to wait for OAuth resources to be missing during cleanup")
```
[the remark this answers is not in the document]
liouk: Copy-paste typo :+1:

test/e2e/external_oidc_test.go
liouk: Using an existing instance for the test is only meant to aid development (as we'll be able to run the test immediately instead of creating and waiting for the new keycloak to come up). The primary mode of operation of the test will be to create a new instance using `keycloakidp.go`.

everettraven: While I see the value from the devex perspective, is the keycloak instance cleaned up after the test ends in the primary mode of operation?

I am a bit hesitant to have this injection mode that could potentially be abused beyond devex and cause confusion. IMO we either go fully configurable to use an existing keycloak instance or not at all.

everettraven: How much time is actually saved per run by not having to stand up keycloak?

liouk: In the primary mode of operation, the keycloak instance is cleaned up; its whole namespace gets deleted. There's another env var (`OPENSHIFT_KEEP_IDP`) that is used to prevent the cleanup, for debugging purposes.

It's not so much about how much time we save, but rather sometimes there can be failures associated with the keycloak being deployed successfully which makes it frustrating when you're trying to test something else. This was mainly useful during the bulk of the development process, so I'm happy to drop this now. It's easy to quickly modify the test in this way again anyway.

liouk:
Latest changes not tested yet due to recent CI/clusterbot issues; will remove hold once tested.

/hold

test/e2e/external_oidc_test.go
```diff
@@ -187,263 +276,388 @@ func newTestClient(t *testing.T) (*testClient, error) {
...
-				{
-					ClientID:           "console",
-					ClientSecret:       configv1.SecretNameReference{Name: "console-secret"},
-					ComponentName:      "console",
-					ComponentNamespace: "openshift-console",
-				},
-			},
-		},
+	for _, obj := range []struct {
+		gvr       string
```
everettraven, line 506: Any reason to not just make this `schema.GroupVersionResource`?

I know this makes the actual test definitions a bit more verbose, but it prevents an additional failure mode for the test of having to parse the string into the type.

liouk: I opted for the full string representation as I feel it is a more familiar view of a fully qualified resource name, thus making it easier to read, at the expense of parsing it. And besides, since these are const strings, there wouldn't be any spontaneous failure of the test, maybe during development if we add something there.

Happy to change it to an object though :+1:

test/e2e/external_oidc_test.go
```diff
@@ -64,110 +66,197 @@ type oidcAuthResponse struct {
...
+				TokenClaimMapping: configv1.TokenClaimMapping{
+					Claim: "email",
+				},
+				PrefixPolicy: configv1.NoOpinion,
+			}),
+			expectedUsernamePrefix:      "",
+			expectOperatorDegraded:      false,
+			expectOIDCRolloutSuccessful: true,
+		},
+	} {
```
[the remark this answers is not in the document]
liouk: The reason we're testing the possible claim mappings changes is because this will be a behavioral change, in the sense that the resulting username present in the token will be different. I'll also add one case where the username claim is an empty string :+1: 

I tried to pick one example from each failure class; e.g. an invalid config results in operator going degraded (tested by providing a non-existing CA configmap). Other cases should be tested extensively in unit tests (for example, [config map related failures](https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/controllers/externaloidc/externaloidc_controller_test.go#L494)).

Finally, there are some cases where errors will not manifest in an easily detectable way; if I remember correctly, your example with an invalid URL will only get caught at the KAS side and won't degrade the CAO/KAS-o. I'll play around with this case though, as we should be able to test this by being unable to login.

test/e2e/external_oidc_test.go
everettraven: While I think that is a reasonable argument, in a set of e2e tests that can already have numerous failure modes I'd much prefer to reduce those where possible. 

IMO the small sacrifice of readability for the gain of one less failure mode is a beneficial change for long running tests like this.

test/e2e/external_oidc_test.go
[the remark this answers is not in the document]
liouk: Keeping in mind this discussion, I'll have a think of whether there are any such cases that we should add an e2e test for, while avoiding too much overlap with units (e.g. incorrect URL results in failure to login).

> a gap in our validation
I've tried to keep validations at the operator side to a minimum, as per [this discussion](https://github.com/openshift/cluster-authentication-operator/pull/713#discussion_r1850572236), to avoid skew between what the server-side expects and what the operator does.

> I don't see a unit test that uses a totally invalid cert value
Maybe we missed this; I'll have another look and add any tests that might be missing.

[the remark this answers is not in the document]
everettraven: >I've tried to keep validations at the operator side to a minimum, as per https://github.com/openshift/cluster-authentication-operator/pull/713#discussion_r1850572236, to avoid skew between what the server-side expects and what the operator does.

Thanks for the link for further context. I generally agree with Ben's comment. That being said, I think this is a gap in our validations on the actual OpenAPI validations for the field in the CRD. That is where I think this validation should be added, but because we don't right now I think it is worth having an e2e test that attempts to use an invalid URL so we know what happens in that case.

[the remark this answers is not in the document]
liouk: Actually, it just occurred to me:

> Invalid issuer URL (right now, we only validate that the issuer URL contains https://[1](https://github.com/openshift/api/blob/b7680e188c5efa5780f93cb89dfa73d728ed5f12/config/v1/types_authentication.go#L232)). Something like https://!@^#&!@^#.com is probably not a valid input

This case will cause a validation error, because the operator validates the issuer URL and CA; therefore this will cause the operator to go degraded.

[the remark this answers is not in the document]
everettraven: >This case will cause a validation error, because the operator validates the issuer URL and CA; therefore this will cause the operator to go degraded.

Awesome, lets make sure we continue to expect an error in that case with a test

liouk:
The `TestExternalOIDCWithKeycloak` test runs more than 1h, which is the [default e2e test timeout](https://github.com/openshift/cluster-authentication-operator/blob/master/Makefile#L46). I am planning on moving this test to its own sub-dir and creating its own test invocation, in the spirit of [test-e2e-encryption](https://github.com/openshift/cluster-authentication-operator/blob/master/Makefile#L60-L66).

To facilitate the ongoing review though, I will do so after the current changes have been reviewed, to preserve the diffs.

/hold

test/e2e/external_oidc_test.go
```diff
@@ -2,54 +2,244 @@ package e2e
...
+	authCM    = "auth-config"
 )
 
-type testClient struct {
-	t *testing.T
+func TestExternalOIDCWithKeycloak(t *testing.T) {
+	testCtx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	tc, err := newTestClient(t)
```
everettraven, line 55: Nit: More descriptive name like `testClient`? A few times going through this PR I got confused because I commonly see `tc` represented as "test case" so it threw me off a bit that this meant "test client".

test/e2e/external_oidc_test.go
```diff
@@ -2,54 +2,244 @@ package e2e
...
+	t.Run("invalid OIDC config crashes kube-apiserver pods during rollout", func(t *testing.T) {
+		tc.updateAuthResource(t, testCtx, testSpec, func(baseSpec *configv1.AuthenticationSpec) {
+			baseSpec.OIDCProviders[0].ClaimMappings.Username.Claim = ""
+		})
+
+		require.NoError(t, test.WaitForClusterOperatorStatusAlwaysAvailable(t, testCtx, tc.configClient.ConfigV1(), "authentication"))
+		require.NoError(t, test.WaitForClusterOperatorStatusAlwaysProgressing(t, testCtx, tc.configClient.ConfigV1(), "kube-apiserver"))
+
+		tc.requireKASRolloutCrash(t, testCtx, nil, 0)
+	})
```
everettraven, line 161: I think we want to make sure we _don't_ crash the kube-apiserver wherever possible. 

I know that this would cause the e2e test to permafail, but I'd rather we not have a false sense of "success" and instead fix this gap. If you don't think it makes sense to fix this as part of this PR, that's fine - but I _do_ think we should assert what we actually want to happen here (i.e we go degraded instead of applying the config) and deal with the permafail there.

This might mean that this PR doesn't merge until after a fix of some kind is put in place, but I think that price is worth paying for us to not forget about this gap and push it under the rug.

test/e2e/external_oidc_test.go
```diff
@@ -2,54 +2,244 @@ package e2e
...
+		}{
+			{"email", configv1.Prefix, &configv1.UsernamePrefix{PrefixString: "oidc-test:"}, "oidc-test:"},
+			{"email", configv1.NoPrefix, nil, ""},
+			{"sub", configv1.NoOpinion, nil, idpURL + "#"},
+			{"email", configv1.NoOpinion, nil, ""},
+		} {
+			policyStr := "NoOpinion"
+			if len(tt.prefixPolicy) > 0 {
+				policyStr = string(tt.prefixPolicy)
+			}
```
everettraven, line 198: Just set it to the prefix policy defined in the test case? All existing test cases specify the prefix policy explicitly and I can't think of a scenario where it would make sense to define a test without an explicit policy.

test/e2e/external_oidc_test.go
```diff
@@ -64,110 +254,172 @@ type oidcAuthResponse struct {
...
 
-	data, err := io.ReadAll(resp.Body)
-	require.NoError(t, err)
-	require.NotEmpty(t, data)
+	if len(featureGates.Status.FeatureGates) != 1 {
+		// fail test if there are multiple feature gate versions (i.e. ongoing upgrade)
+		t.Fatalf("multiple feature gate versions detected")
+	} else {
+		for _, fgDisabled := range featureGates.Status.FeatureGates[0].Disabled {
+			if fgDisabled.Name == features.FeatureGateExternalOIDC {
```
everettraven, line 367: We know we are going to end up with multiple feature-gates corresponding to this functionality due to the way we need to navigate HCP having gone GA with this feature already and that impacting our ability to make API changes under a single gate.

With that in mind, maybe it is worth preemptively making this a set based check so we can easily add to the set when new feature gates are added?

The set based check may be easier to do with a flipped polarity of this check though. Checking if any of the feature gates in the set are enabled is likely easier than checking if _all_ the feature gates are in the disabled set.

test/e2e/external_oidc_test.go
```diff
@@ -187,263 +439,311 @@ func newTestClient(t *testing.T) (*testClient, error) {
...
+		found := false
+		for _, route := range ingress.Status.ComponentRoutes {
+			if route.Name == obj.name && route.Namespace == obj.namespace {
+				found = true
+				break
+			}
+		}
+
+		if requireMissing != !found {
+			return false, fmt.Errorf("route %s wanted missing: %v; got: %v", obj, requireMissing, !found)
```
everettraven, line 607: Make it explicit in the error message that it should be missing or not from the cluster ingress status?

test/library/waits.go
```diff
@@ -168,3 +140,87 @@ func WaitForNewKASRollout(t *testing.T, ctx context.Context, kasClient operatorv
...
+	if progressing != "" {
+		progressingStatusIsMatch = v1helpers.IsStatusConditionPresentAndEqual(conditions, configv1.OperatorProgressing, progressing)
+	}
+	if degraded != "" {
+		degradedStatusIsMatch = v1helpers.IsStatusConditionPresentAndEqual(conditions, configv1.OperatorDegraded, degraded)
+	}
+	if upgradable != "" {
+		upgradableStatusIsMatch = v1helpers.IsStatusConditionPresentAndEqual(conditions, configv1.OperatorDegraded, upgradable)
+	}
+	done := availableStatusIsMatch && progressingStatusIsMatch && degradedStatusIsMatch && upgradableStatusIsMatch
```
everettraven, line 224: Nit: Could probably simplify this function a touch with something like:
```suggestion
// pseudo-coded type, but should get the idea across
type requiredCondition struct {
    ConditionType
    ConditionStatus
}

func checkClusterOperatorStatus(t *testing.T, ctx context.Context, client configv1client.ConfigV1Interface, name string, requireConditions ...requiredCondition) (bool, error) {
	clusterOperator, err := client.ClusterOperators().Get(ctx, name, metav1.GetOptions{})
	if errors.IsNotFound(err) {
		t.Logf("clusteroperators.config.openshift.io/%v: %v", name, err)
		return false, nil
	}
	if retry.IsHTTPClientError(err) {
		t.Logf("clusteroperators.config.openshift.io/%v: %v", name, err)
		return false, nil
	}
	conditions := clusterOperator.Status.Conditions
	t.Logf("clusteroperators.config.openshift.io/%v: %v", name, conditionsStatusString(conditions))
	degradedCondition := v1helpers.FindStatusCondition(conditions, configv1.OperatorDegraded)
	if degradedCondition != nil && degradedCondition.Status == configv1.ConditionTrue {
		t.Logf("clusteroperators.config.openshift.io/%v: degraded is true!: %s:%s", name, degradedCondition.Reason, degradedCondition.Message)
	}
	done := true
	for _, condition := range requiredConditions {
	    if condition.Status == "" {
	        continue
	    }
	    if !v1helpers.IsStatusConditionPresentAndEqual(conditions, condition.Type, condition.Status) {
	        done = false
	        break
	    }
	}
```

Totally up to you if you want to take this suggestion or not. I think something like this would make it a touch easier to add checks against different condition types as needed in the future, but if we will never extend it beyond what is already here, then what is here is fine

everettraven: I'm realizing now that the suggested diff is not all that clear. In case it helps, the change I made explicitly were:
- Adding the new `requiredCondition` type
- Updating the function signature from `func checkClusterOperatorStatus(t *testing.T, ctx context.Context, client configv1client.ConfigV1Interface, name string, available configv1.ConditionStatus, progressing configv1.ConditionStatus, degraded configv1.ConditionStatus, upgradable configv1.ConditionStatus) (bool, error)` --> `func checkClusterOperatorStatus(t *testing.T, ctx context.Context, client configv1client.ConfigV1Interface, name string, requireConditions ...requiredCondition) (bool, error)`, which removes the `available, progressing, degraded, upgradeable` parameters in favor of `requireConditions ...requiredCondition`
- Instead of doing each explicit type check, now do a loop that exits early and sets `done` to `false` if _any_ of the conditions don't match the desired ones.

everettraven (commented):
A handful more comments.

test/e2e/external_oidc_test.go
liouk: Well, `configv1.NoOpinion` is an empty string; I added this explicitly because the test name would read weirdly as "username_claim_email_prefix_policy_".

everettraven: Ahhhhh - Gotcha. Thanks for the clarification there. I missed that :). Feel free to ignore

test/e2e/external_oidc_test.go
liouk: My intention was to fix this separately to avoid crashing KAS pods at all, and still have the test validate the current behavior. As discussed offline, we'll skip this test for now, and re-enable once the fix is in place.

test/e2e/external_oidc_test.go
liouk: My intention was to fix this separately to avoid crashing KAS pods at all, and still have the test validate the current behavior. As discussed offline, we'll skip this test for now, and re-enable once the fix is in place.

test/e2e/external_oidc_test.go
everettraven: We discussed offline that planning for the future changes needing feature-gates is not as straightforward as I had originally thought it might be.

Feel free to ignore this suggestion.

everettraven (approved):
Had one non-blocking question, but overall this LGTM. Tagging

/lgtm

test/library/waits.go
```diff
@@ -34,66 +35,82 @@ func WaitForOperatorToPickUpChanges(t *testing.T, configClient configv1client.Co
...
+		t.Logf("clusteroperators.config.openshift.io/%s required status reached after %s: %v", name, time.Now().Sub(ts), conditionsStatusString(conditions))
+		return nil
+	}
+
+	t.Logf("clusteroperators.config.openshift.io/%s required status not reached after %s: %v", name, time.Now().Sub(ts), conditionsStatusString(conditions))
+	return err
+}
+
+// WaitForClusterOperatorStatusStable checks that the specified cluster operator's status does not diverge
+// from the conditions specified for 10 minutes. It returns nil if all conditions were matching expectations for that
```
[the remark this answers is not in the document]
liouk: No reason other than following what `WaitForClusterOperatorStatus` does (although for a different type of check). I'm open to suggestions, but 10m felt more or less appropriate, while not prolonging the test excessively.

PR #740, opened 2024-11-14
WIP: AUTH-543: OIDC/OAuth resource configuration

liouk:
This PR adjusts all OAuth related controllers to remove (or not) their operands depending on whether authentication is external OIDC or not.


liouk:
This contains commits from #713 and must be rebased once that one merges; putting a hold until that's done.

/hold

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -289,3 +312,46 @@ func (c *customRouteController) getFieldManager() string {
 	// TODO find a way to get the client name and combine it with the controller name automatically
 	return "AuthenticationCustomRouteController"
 }
+
+func (c *customRouteController) removeOperands(ctx context.Context, ingressConfig *configv1.Ingress, secretName string) error {
```
ibihim, line 316: secretName is unused.

pkg/controllers/customroute/custom_route_controller.go
```diff
@@ -116,6 +133,12 @@ func (c *customRouteController) sync(ctx context.Context, syncCtx factory.SyncCo
 		return fmt.Errorf("custom route configuration failed verification: %v", errors)
 	}
 
+	if oidcAvailable, err := common.ExternalOIDCConfigAvailable(c.authLister, c.kasLister, c.kasConfigMapLister); err != nil {
+		return err
+	} else if oidcAvailable {
+		return c.removeOperands(ctx, ingressConfigCopy, secretName)
```
ibihim, line 139: Why do we do it that late? We don't need `secretName`. We could start immediately after the deep copy of ingressConfig, could we?

pkg/controllers/deployment/deployment_controller.go
```diff
@@ -135,6 +150,13 @@ func NewOAuthServerWorkloadController(
 }
 
 func (c *oauthServerDeploymentSyncer) PreconditionFulfilled(_ context.Context) (bool, error) {
+	if oidcAvailable, err := common.ExternalOIDCConfigAvailable(c.authLister, c.kasLister, c.kasConfigMapLister); err != nil {
+		return false, err
+	} else if oidcAvailable {
+		// the route is no longer a pre-requisite
+		return true, nil
```
ibihim, line 157: The route is gone. But we can still fulfill the precondition without the need for a deployment of the oauth-server?

pkg/controllers/ingressnodesavailable/ingress_nodes_available_controller.go
```diff
@@ -72,6 +87,12 @@ func countReadyWorkerNodes(nodes []*corev1.Node) int {
...
 func (c *ingressNodesAvailableController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	if oidcAvailable, err := common.ExternalOIDCConfigAvailable(c.authLister, c.kasLister, c.kasConfigMapLister); err != nil {
+		return err
+	} else if oidcAvailable {
+		return common.ApplyControllerConditions(ctx, c.operatorClient, c.controllerInstanceName, knownConditionNames, nil)
+	}
+
 	foundConditions := []operatorv1.OperatorCondition{}
 
 	workers, err := c.nodeLister.List(labels.SelectorFromSet(labels.Set{"node-role.kubernetes.io/worker": ""}))
```
ibihim, line 98: Isn't it odd, that we put the ingress pods on a worker node, while the oauth-servers are on the master node?

pkg/controllers/customroute/custom_route_controller.go
liouk: Good catch! This is leftover from an initial version where I deleted the secret manually but then I reused the syncer.

pkg/controllers/customroute/custom_route_controller.go
liouk: Fixed with 4200edde03363dde52a1473266ce0f0b22be825d

pkg/controllers/customroute/custom_route_controller.go
liouk: Fixed with 4200edde03363dde52a1473266ce0f0b22be825d

liouk:
Pre-merge testing has revealed some issues with this PR; turning this into WIP until fixed.

/retitle WIP: AUTH-543: OIDC/OAuth resource configuration

everettraven (commented):
Doing an initial pass.

Change I'd like to see throughout - wrapped errors. Right now, I see a lot of `return err` that might result in some fairly opaque errors that could make it harder to diagnose issues in the future. Could we wrap all of these with some relevant context?

pkg/controllers/common/external_oidc.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
+)
+
+type AuthConfigChecker struct {
+	authentications        configv1informers.AuthenticationInformer
+	authLister             configv1listers.AuthenticationLister
+	kubeAPIServers         operatorv1informers.KubeAPIServerInformer
+	kasLister              operatorv1listers.KubeAPIServerLister
```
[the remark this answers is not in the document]
liouk: The AuthenticationInformer and KubeAPIServerInformer are accessed via the accessor methods in controllers that need to use them to trigger syncs. But indeed, the ConfigMapInformer one isn't used anywhere, this was probably a left over during development; I'll drop that one.

pkg/controllers/oauthclientsswitchedinformer/oauthclientsswitchedinformer_controller.go
```diff
@@ -0,0 +1,130 @@
...
+
+	oauthinformersv1 "github.com/openshift/client-go/oauth/informers/externalversions/oauth/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+// InformerWithSwitch is a controller that can start and stop an informer based on
```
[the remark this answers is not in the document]
liouk: This is partially correct; we do not intend to stop any controllers, as they still need to monitor auth for a change back to OAuth and handle their operands accordingly (i.e. we want controllers to run and react to auth changes by deleting or creating their operands).

We _do_ want to stop/start informers that track OAuth stack dependent resources when OAuth is down/up though, and indeed we want to avoid CAO restarts to achieve this.

The switched informer is a mechanism that was implemented in the console-operator for external OIDC for this purpose (console has an oauthclient that the operator is tracking), so I basically transferred the same mechanism here.

Note for your InformerWithSwitch comments: I chose to transfer this code in the CAO in an attempt to reduce development effort, and I was planning on generalizing this and moving it into an appropriate library (e.g. lib-go) so that it can be reused in the console-operator as well, after GA.

I'll try to make some improvements based on your comments that we can also leverage elsewhere without blowing up the refactoring :sweat_smile:

pkg/controllers/oauthclientsswitchedinformer/oauthclientsswitchedinformer_controller.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (s *InformerWithSwitch) ensureRunning() {
+	if s.runCtx != nil {
+		return
+	}
+
+	klog.Infof("%s delegate informer starting", s.switchController.Name())
+	s.runCtx, s.stopFunc = context.WithCancel(s.parentCtx)
+	go s.delegateInformer.Informer().Run(s.runCtx.Done())
```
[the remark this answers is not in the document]
liouk: You're actually right, nice catch -- this seems to be a bug in the initial implementation as well (see my comment further down). I'll try to fix this here, and when we have a more generic version that can be used elsewhere, I'll pull this into console-operator as well.

pkg/libs/endpointaccessible/endpoint_accessible_controller_test.go
```diff
@@ -48,12 +49,27 @@ func Test_endpointAccessibleController_sync(t *testing.T) {
...
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			c := &endpointAccessibleController{
-				operatorClient: v1helpers.NewFakeOperatorClient(&operatorv1.OperatorSpec{}, &operatorv1.OperatorStatus{}, nil),
-				endpointListFn: tt.endpointListFn,
+				operatorClient:            v1helpers.NewFakeOperatorClient(&operatorv1.OperatorSpec{}, &operatorv1.OperatorStatus{}, nil),
+				endpointListFn:            tt.endpointListFn,
+				endpointCheckDisabledFunc: tt.endpointCheckDisabledFunc,
```
everettraven, line 72: Should we add a check to this test that if this returns true that the controller conditions are properly removed?

pkg/controllers/common/external_oidc.go
```diff
@@ -0,0 +1,105 @@
...
+	return c.authentications
+}
+
+func (c *AuthConfigChecker) KubeAPIServers() operatorv1informers.KubeAPIServerInformer {
+	return c.kubeAPIServers
+}
+
+func (c *AuthConfigChecker) KubeAPIServerNamespaceConfigMaps() corev1informers.ConfigMapInformer {
+	return c.kasNamespaceConfigMaps
+}
```
everettraven, line 53: I'm still wondering if we actually need the distinction between the informers here or if it is reasonable to assume that any controller using the `AuthConfigChecker` would want to inherently add all of its informers to the set of informers that send a signal to the controller?

Having the distinction between the informers means that if we end up needing to add an informer here we then have to go back through every controller that uses all the informers from this type and update it.

If we did something like:
```go
func (c *AuthConfigChecker) Informers() []cache.SharedIndexInformer {...}
```
then callers could use it like:
```go
return factory.New().WithInformers(..., authConfigChecker.Informers()...)
```

With that approach, adding new informers that should trigger a sync evaluation based on the OIDC configuration should automatically be picked up by all controllers using it.

pkg/controllers/switchedcontroller/switched_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if c.switchContextCancel != nil {
+		c.switchContextCancel()
+	}
+
+	c.switchContext = nil
+	c.delegateController = nil
+}
+
+func (c *ControllerWithSwitch) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
everettraven, line 92: Is it safe for us to assume that the `ctx` parameter that is passed in here is a long lived context that isn't cancelled after the sync method completes?

If it does get cancelled after the sync method completes, it likely is not safe for us to use this context as the basis for starting and stopping the delegate controller as we would start it, return from `sync` and then it would be immediately stopped.

pkg/controllers/switchedcontroller/switched_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	if c.delegateFactoryFn == nil {
+		return fmt.Errorf("no delegate factory function defined")
+	}
+
+	if c.switchConditionFn == nil {
+		klog.Infof("no switch condition defined; controller '%s' will never be stopped", c.delegateName)
+		c.runDelegate(ctx, syncCtx)
+		return nil
+	}
```
everettraven, line 101: Should we even allow this scenario to occur in the sync method? If you want to have the controller run unconditionally, just don't use a switched controller?

Maybe this should be a check that panics in the `NewControllerWithSwitch` function?

pkg/controllers/switchedcontroller/switched_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	c.switchContext = nil
+	c.delegateController = nil
+}
+
+func (c *ControllerWithSwitch) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	if c.delegateFactoryFn == nil {
+		return fmt.Errorf("no delegate factory function defined")
+	}
```
everettraven, line 95: Also seems reasonable as a panic in the `NewControllerWithSwitch` function

pkg/controllers/switchedcontroller/switched_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+type ControllerWithSwitch struct {
+	delegateName       string
+	delegateFactoryFn  DelegateFactoryFunc
+	delegateController factory.Controller
```
everettraven, line 19: Do we actually need to store a reference to the delegate controller here? Can we just blindly create a new delegate controller and run it in a goroutine with the right context and when it gets shutdown it automatically get cleaned up?

pkg/controllers/switchedcontroller/switched_controller.go
```diff
@@ -0,0 +1,127 @@
...
+		// context alive, delegate running
+
+	case !switchOn && c.switchContext == nil:
+		// we haven't been asked to start yet
+
+	case !switchOn && c.switchContext != nil && c.switchContext.Err() == nil:
+		c.stopDelegate()
+
+	default:
+		return fmt.Errorf("this should never happen; switchOn = %v; switchContext error = %v", switchOn, c.switchContext.Err())
```
everettraven, line 123: We would end up hitting this in the scenario where:
- `switchOn = true`
- `c.switchContext != nil`
- `c.switchContext.Err() != nil`

I.e - we want the controller to be running, but the context has been canceled and is therefore not running.

I think we would _also_ hit this statement in the scenario where we want to switch the controller off but it has already been switched off by a canceled context.

Because we are using a passed in context, are we actually able to guarantee that never happens?

pkg/controllers/switchedcontroller/switched_controller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		WithInformers(informers...).
+		ResyncEvery(resyncInterval).
+		ToController(delegateName+"_SwitchedController", eventRecorder)
+}
+
+func (c *ControllerWithSwitch) runDelegate(ctx context.Context, syncCtx factory.SyncContext) {
```
everettraven, line 70: Should we gate this method with a check that if the delegate is already running we don't end up running it again?

pkg/controllers/oauthendpoints/oauth_endpoints_controller.go
```diff
@@ -34,6 +34,7 @@ func NewOAuthRouteCheckController(
 	kubeInformersForConfigManagedNS informers.SharedInformerFactory,
 	routeInformerNamespaces routev1informers.RouteInformer,
 	ingressInformerAllNamespaces configv1informers.IngressInformer,
+	authConfigChecker common.AuthConfigChecker,
```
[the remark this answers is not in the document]
liouk: My reasoning for all controllers that perform checks (instead of managing operands) was the following: we don't need to trigger a sync when a change occurs in these informers just to end up skipping the check. We can rely on the standard sync, and since the caches will be synced, eventually we'll start skipping the check.

What do you think?

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -30,22 +31,26 @@ type proxyConfigChecker struct {
...
 	routeNamespace  string
 	caConfigMaps    map[string][]string // ns -> []configmapNames
+
+	authConfigChecker common.AuthConfigChecker
 }
 
 func NewProxyConfigChecker(
 	routeInformer routeinformer.RouteInformer,
 	configMapInformers v1helpers.KubeInformersForNamespaces,
+	authConfigChecker common.AuthConfigChecker,
```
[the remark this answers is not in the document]
liouk: Same as above.

pkg/controllers/common/external_oidc.go
liouk: Fair point -- I made it so initially for two reasons:
- some controllers were already tracking some of those informers (basically just the auth one in some cases)
- some controllers might need a lister off those informers and this would make it easy to obtain one

For the 1st point: I will remove extra informers from those controllers and just have them track the ones from the `authConfigTracker`; this will basically hide the fact that the controller is interested in that informer for other reasons as well, but that's minor.

The 2nd point seems like it's actually not happening, even if I originally thought it'd be needed so it's redundant.

liouk: If we decide to bundle all in one call, we might need something like this:
```go
type AuthConfigChecker struct {
	authenticationsInformer        cache.SharedIndexInformer
	kubeAPIServersInformer         cache.SharedIndexInformer
	kasNamespaceConfigMapsInformer cache.SharedIndexInformer

	authLister         configv1listers.AuthenticationLister
	kasLister          operatorv1listers.KubeAPIServerLister
	kasConfigMapLister corelistersv1.ConfigMapLister
}

func NewAuthConfigChecker(authentications configv1informers.AuthenticationInformer, kubeapiservers operatorv1informers.KubeAPIServerInformer, configmaps corev1informers.ConfigMapInformer) AuthConfigChecker {
	return AuthConfigChecker{
		authenticationsInformer:        authentications.Informer(),
		kubeAPIServersInformer:         kubeapiservers.Informer(),
		kasNamespaceConfigMapsInformer: configmaps.Informer(),
		authLister:                     authentications.Lister(),
		kasLister:                      kubeapiservers.Lister(),
		kasConfigMapLister:             configmaps.Lister(),
	}
}

// usable by factory.Factory.WithInformers(...factory.Informer)
func (c *AuthConfigChecker) Informers() []factory.Informer {
	return []factory.Informer{
		c.authenticationsInformer,
		c.kubeAPIServersInformer,
		c.kasNamespaceConfigMapsInformer,
	}
}

// usable by StaticResourceController.AddInformer(cache.SharedIndexInformer)
func (c *AuthConfigChecker) SharedIndexInformers() []cache.SharedIndexInformer {
	return []cache.SharedIndexInformer{
		c.authenticationsInformer,
		c.kubeAPIServersInformer,
		c.kasNamespaceConfigMapsInformer,
	}
}
```

Otherwise we'll have to resort to adding for loops everywhere to call `WithInformers()` with `cache.SharedIndexInformer` one-by-one :/

What do you think?

pkg/controllers/switchedcontroller/switched_controller.go
liouk: You're right, I was actually thinking that the `delegateFactoryFn` should result in a panic as well -- no point in setting up an operator instance, running it just to degrade it because you didn't set up the controller correctly.

pkg/controllers/switchedcontroller/switched_controller.go
liouk: To avoid checking in multiple places I'll probably drop the `runDelegate/stopDelegate` funcs, they'll only be used in one place and they're rather short anyway, so I'll inline them.

pkg/controllers/common/external_oidc.go
liouk: As discussed offline, we'll use an approach with generics.

liouk:
> Change I'd like to see throughout - wrapped errors. Right now, I see a lot of `return err` that might result in some fairly opaque errors that could make it harder to diagnose issues in the future. Could we wrap all of these with some relevant context?

@everettraven  I've tried to wrap errors where sensible, while sticking to what controllers seem to be currently doing (i.e. `sync()` funcs usually return plain errors since they're wrapped in a handler and will degrade operator).

everettraven:
This should be ready to go. Removing hold.

/hold cancel

PR #744, opened 2024-11-26
API-1835: bring MOM up to date for testing and the like

deads2k:
Now using the latest library-go and MOM.


deads2k:
neat thing. This is showing unstable output which is goign to cause repeated writes for the operator.  We never knew before, but now we do.

deads2k:
Oh, and we didn't cause the unstable writes. It's simply made it obvious to us.

p0lyn0mial:
/lgtm

/hold

so that @ibihim and @liouk have a chance to review it :)

deads2k:
time's up

/hold cancel

PR #748, opened 2025-01-07
CNTRLPLANE-71: update cao to manage rolebindingrestriction crd

everettraven:
## Description

Updates the cluster-authentication-operator to manage the `RoleBindingRestriction` CRD.

More specifically, this PR:
- Vendors the `authorization.openshift.io/RoleBindingRestriction` CRD manifest from https://github.com/openshift/api
- Adds targets to the Makefile to:
    - Copy the `RoleBindingRestriction` CRD manifests to the `bindata/` directory so the manifest can be embedded into the binary using the embed FS.
    - Verify the `bindata/` directory is up to date
    - Verify the `RoleBindingRestriction` CRD manifest in `bindata/` is up to date based on the latest vendored manifest.
- Adds the `oauth-openshift/authorization.openshift.io_rolebindingrestrictions.yaml` file to the list of manifest files managed by the static resource controller
- Adds an apiextensions client to support the management of CustomResourceDefinition resources using the static resource controller.
- Adds a `render` subcommand to be used by the openshift installer during bootstrapping because the `RoleBindingRestriction` CRD is required during bootstrapping. Without it, the `authorization.openshift.io/RestrictSubjectBindings` admission plugin will reject creation of `system:*` RoleBindings during the installation process.

## Motivation

Update the cluster-authentication-operator to manage the `RoleBindingRestriction` CRD, as outlined in https://github.com/openshift/enhancements/pull/1726, to allow for this CRD to be removed from the cluster when the oauth stack is no longer desired.


pkg/operator/starter.go
```diff
@@ -142,8 +141,9 @@ func prepareOauthOperator(
 			"oauth-openshift/oauth-service.yaml",
 			"oauth-openshift/trust_distribution_role.yaml",
 			"oauth-openshift/trust_distribution_rolebinding.yaml",
+			"oauth-openshift/authorization.openshift.io_rolebindingrestrictions.yaml",
```
liouk, line 144: If you want to use a StaticResourceController to manage the CRD, use a conditional resource (via `WithConditionalResources`; `WithPrecondition` might be useful too), to manage when this should be created/deleted and avoid contention.

liouk (changes requested):

pkg/operator/starter.go
everettraven: Launching a cluster with this PR as-is seemed to work just fine and didn't appear to have any contention between this and CVO.

I'm happy to add the conditional resources, but it would essentially be:
- Create only when the resource is not found or requires updates
- Delete never

Walking through the underlying logic, this seems to be the exact flow the static resource controller already takes:
- Calls `ApplyDirectly()`: https://github.com/openshift/library-go/blob/4ea50293b28af3d28f547c670f52c175af9e4427/pkg/operator/staticresourcecontroller/static_resource_controller.go#L334
- For CustomResourceDefinitions, `ApplyDirectly()` calls `ApplyCustomResourceDefinitionV1()`: https://github.com/openshift/library-go/blob/4ea50293b28af3d28f547c670f52c175af9e4427/pkg/operator/resource/resourceapply/generic.go#L179
- `ApplyCustomResourceDefinitionV1()` checks if the resource exists or has been modified to determine if it needs to sync it: https://github.com/openshift/library-go/blob/4ea50293b28af3d28f547c670f52c175af9e4427/pkg/operator/resource/resourceapply/apiextensions.go#L19-L35

I'm not sure it makes much sense to gate this resource on the same conditions that would be done by default from the static resource controller. We _will_ add conditional logic for create/delete in the future based on desired state of the OAuth stack (i.e if the OAuth stack should be removed, this resource should also be removed and vice versa)

pkg/cmd/mom/output_resources_command.go
```diff
@@ -38,6 +39,8 @@ func runOutputResources(ctx context.Context) (*libraryoutputresources.OutputReso
 				libraryoutputresources.ExactRoleBinding("openshift-config-managed", "system:openshift:oauth-servercert-trust"),
 
 				libraryoutputresources.ExactPDB("openshift-oauth-apiserver", "oauth-apiserver-pdb"),
+
+				libraryoutputresources.ExactResource(apiextensionsv1.SchemeGroupVersion.Group, apiextensionsv1.SchemeGroupVersion.Version, "customresourcedefinitions", "", "rolebindingrestrictions.authorization.openshift.io"),
```
deads2k, line 43: UserWorkloadResources

everettraven: Should be updated in https://github.com/openshift/cluster-authentication-operator/pull/748/commits/bca7a996c65d0cd7ae33691bdc1c98b7fb892661

everettraven:
After testing with some related PRs, this PR should be ready to go. Removing the hold.

Tested standing up a cluster with this PR and https://github.com/openshift/installer/pull/9424 as well as this PR, https://github.com/openshift/installer/pull/9424, and https://github.com/openshift/api/pull/2138

/hold cancel

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/openshift/cluster-authentication-operator/bindata"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+type Permission os.FileMode
```
liouk, line 13: Why rename the std lib type? Use the original type instead. You'll avoid unnecessary casting when using `os` functions as well.

pkg/cmd/render/render.go
```diff
@@ -0,0 +1,76 @@
...
+const (
+	PermissionDirectoryDefault Permission = 0755
+	PermissionFileDefault      Permission = 0644
+)
+
+type AssetsFunc func(name string) ([]byte, error)
+
+type RenderOptions struct {
+	AssetOutputDir string
+	Assets         AssetsFunc
```
liouk, line 24: How about using `resourceapply.AssetFunc` from library-go instead of defining a new func type?

liouk (changes requested):

pkg/cmd/render/render.go
everettraven: I was using https://github.com/openshift/api/blob/4e32bc25f4f950292cc3a8b7322cba3407a62d2f/payload-command/render/renderassets/assets.go#L14-L20 as inspiration to try and stay consistent with how it is done elsewhere in the openshift org

I can update this to be the std lib type.

pkg/cmd/render/render.go
everettraven: Good catch. I hadn't even considered that

Makefile
```diff
@@ -15,7 +15,7 @@ include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 #
 # Example:
 #   make check
-check: | verify test-unit
+check: | verify verify-bindata test-unit
```
liouk, line 18: Just to make sure -- CI will run `check` which means that bindata will also be verified, and in case there's a diff between the api copy and the local copy, we'll get a failed test, right?

everettraven: Looking at https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-authentication-operator/openshift-cluster-authentication-operator-master.yaml I don't actually think `check` is run in CI explicitly, but I added this here since it seemed like a natural fit.

I'll plan to add a new CI check to run this make target today (although it will fail until this PR is merged).

everettraven: Created https://github.com/openshift/release/pull/61456 to add this as an explicit CI check

PR #750, opened 2025-01-10
NO-JIRA: fix: resolve some integration tests

eggfoobar:
There are three issues that are currently the reason for the integration tests failing

1. errors do not come back in lexical order so the match fails
2. status controller does not use mom timestamp to timestamps do not match
3. certs get generated differently between runs so certs do not match

This PR pulls in the a latest from library-go and resolves number 1 and number 2. 

/hold

We'll need to wait for https://github.com/openshift/library-go/pull/1916 to be merged in.

PR #752, opened 2025-01-16
MON-4129: adjust Prometheus classic histograms 'le' related selectors in relabel config to accommodate the update to Prometheus v3

machine424:


machine424:
/retitle MON-4129: adjust Prometheus histograms 'le' related selectors in relabel config

machine424:
/retitle MON-4129: adjust Prometheus classic histograms 'le' related selectors in relabel config to accommodate the update to Prometheus v3

deads2k:
interesting that the integration test doesn't apply thing

/approve

machine424:
> interesting that the integration test doesn't apply thing

Took a look (even though the job isn't required) and I assumed that's a known issue and would be addressed in a following PR: https://github.com/openshift/cluster-authentication-operator/pull/750#issue-2779422617

PR #757, opened 2025-02-12
OCPBUGS-48176: Avoid duplicate OAuth client creation

kramaranya:
### What
[OCPBUGS-48176](https://issues.redhat.com/browse/OCPBUGS-48176)

Changing the way we create OAuth clients - now the code first checks if the OAuth client already exists. If it exists, it updates the client only when needed. This stops the repeated "AlreadyExists" (409) errors in the logs.

### Verification Steps
- Review the changes
- Provision a cluster for this pr
- Check the audit logs and confirm that you do not see repeated 409 errors for OAuth clients
```
oc adm node-logs ip-10-0-108-43.us-east-2.compute.internal --path=oauth-apiserver/audit.log | grep '"code":409' | grep 'openshift-authentication-operator'
```
- Confirm that scaling the operator deployment does not trigger extra 409 errors

### DIFF
--- a/pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
+++ b/pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
@@ -138,7 +138,7 @@ func (c *oauthsClientsController) ensureBootstrappedOAuthClients(ctx context.Con
 			GrantMethod:  oauthv1.GrantHandlerAuto,
 		},
 	} {
-		if err := ensureOAuthClient(ctx, c.oauthClientClient, client); err != nil {
+		if err := c.ensureOAuthClient(ctx, client); err != nil {
 			return fmt.Errorf("unable to ensure existence of a bootstrapped OAuth client %q: %w", client.Name, err)
 		}
 	}
@@ -158,14 +158,19 @@ func randomBits(bits uint) []byte {
 	return b
 }
 
-func ensureOAuthClient(ctx context.Context, oauthClients oauthclient.OAuthClientInterface, client oauthv1.OAuthClient) error {
-	_, err := oauthClients.Create(ctx, &client, metav1.CreateOptions{})
-	if err == nil || !apierrors.IsAlreadyExists(err) {
+func (c *oauthsClientsController) ensureOAuthClient(ctx context.Context, client oauthv1.OAuthClient) error {
+	_, err := c.oauthClientLister.Get(client.Name)
+	if apierrors.IsNotFound(err) {
+		_, err = c.oauthClientClient.Create(ctx, &client, metav1.CreateOptions{})
+		return err
+	}
+
+	if err != nil {
 		return err
 	}
 
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		existing, err := oauthClients.Get(ctx, client.Name, metav1.GetOptions{})
+		existing, err := c.oauthClientLister.Get(client.Name)
 		if err != nil {
 			return err
 		}
@@ -188,7 +193,7 @@ func ensureOAuthClient(ctx context.Context, oauthClients oauthclient.OAuthClient
 			return nil
 		}
 
-		_, err = oauthClients.Update(ctx, existingCopy, metav1.UpdateOptions{})
+		_, err = c.oauthClientClient.Update(ctx, existingCopy, metav1.UpdateOptions{})
 		return err
 	})
 }
--- a/pkg/controllers/oauthclientscontroller/oauthclientscontroller_test.go
+++ b/pkg/controllers/oauthclientscontroller/oauthclientscontroller_test.go
@@ -104,19 +104,19 @@ func newRouteLister(t *testing.T, routes ...*routev1.Route) routev1listers.Route
 	return routev1listers.NewRouteLister(routeIndexer)
 }
 
-func newTestOAuthsClientsController(t *testing.T) *oauthsClientsController {
+func newTestOAuthsClientsController(t *testing.T) (*oauthsClientsController, cache.Indexer) {
+	indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
 	return &oauthsClientsController{
 		oauthClientClient: fakeoauthclient.NewSimpleClientset().OauthV1().OAuthClients(),
-		oauthClientLister: oauthv1listers.NewOAuthClientLister(cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})),
+		oauthClientLister: oauthv1listers.NewOAuthClientLister(indexer),
 		routeLister:       newRouteLister(t, defaultRoute),
 		ingressLister:     newIngressLister(t, defaultIngress),
-	}
+	}, indexer
 }
 
 func Test_sync(t *testing.T) {
 	ctx := context.TODO()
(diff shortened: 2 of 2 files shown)


pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -159,8 +159,12 @@ func randomBits(bits uint) []byte {
 }
 
 func ensureOAuthClient(ctx context.Context, oauthClients oauthclient.OAuthClientInterface, client oauthv1.OAuthClient) error {
-	_, err := oauthClients.Create(ctx, &client, metav1.CreateOptions{})
-	if err == nil || !apierrors.IsAlreadyExists(err) {
+	_, err := oauthClients.Get(ctx, client.Name, metav1.GetOptions{})
```
ibihim, line 162: We would ideally use the lister, which caches this information for us.

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -159,8 +159,12 @@ func randomBits(bits uint) []byte {
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
liouk, line 164: Nit: maybe unnest the nested if? You can check whether `apierrors.IsNotFound()` regardless of whether `err` is `nil`.

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
ibihim: A good guide, when working with controllers is this: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1675#discussion_r1596768462.

kramaranya: that's actually makes so much sense, thanks Krzys for the advice
I've updated the code

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -158,16 +158,24 @@ func randomBits(bits uint) []byte {
 	return b
 }
 
-func ensureOAuthClient(ctx context.Context, oauthClients oauthclient.OAuthClientInterface, client oauthv1.OAuthClient) error {
-	_, err := oauthClients.Create(ctx, &client, metav1.CreateOptions{})
-	if err == nil || !apierrors.IsAlreadyExists(err) {
+func ensureOAuthClient(ctx context.Context, oauthClients oauthclient.OAuthClientInterface, oauthClientLister oauthv1listers.OAuthClientLister, client oauthv1.OAuthClient) error {
```
liouk, line 161: @kramaranya @ibihim what do you think about making this a method on `oauthsClientsController`, instead of passing in the client and the lister?

ibihim: Functions are nearly always better than methods, but if they get like plenty of arguments from the object... it might be better to make them methods :)

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -158,16 +158,24 @@ func randomBits(bits uint) []byte {
...
 
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		existing, err := oauthClients.Get(ctx, client.Name, metav1.GetOptions{})
+		existing, err := oauthClientLister.Get(client.Name)
 		if err != nil {
-			return err
+			existing, err = oauthClients.Get(ctx, client.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
```
everettraven, line 178: Are there any particular errors that we could get from the lister that would make us think doing a live GET will result in something different?

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
```diff
@@ -158,16 +158,24 @@ func randomBits(bits uint) []byte {
...
+func ensureOAuthClient(ctx context.Context, oauthClients oauthclient.OAuthClientInterface, oauthClientLister oauthv1listers.OAuthClientLister, client oauthv1.OAuthClient) error {
+	_, err := oauthClientLister.Get(client.Name)
+	if apierrors.IsNotFound(err) {
+		_, err = oauthClients.Create(ctx, &client, metav1.CreateOptions{})
 		return err
+	} else if err != nil {
+		if _, err = oauthClients.Get(ctx, client.Name, metav1.GetOptions{}); err != nil {
+			return err
+		}
 	}
```
everettraven, line 170: Same question here, what error state from the lister will result in us getting a better result from a live GET?

pkg/controllers/oauthclientscontroller/oauthclientscontroller_test.go
```diff
@@ -512,21 +513,19 @@ func Test_ensureOAuthClient(t *testing.T) {
...
+						schema.GroupVersionResource{Group: oauthv1.GroupName, Version: oauthv1.GroupVersion.Version, Resource: "oauthclients"},
+						tt.updateOAuthClient.Namespace, tt.updateOAuthClient.Name,
+					)
+					if err != nil {
+						t.Fatal(err)
+					}
+
+					if !equality.Semantic.DeepEqual(expected, updated) {
+						t.Errorf("client after update does not equal the expected one; got: %v; want: %v", updated, expected)
+					}
```
everettraven, line 528: Why is this particular change being made?

kramaranya: This change was made because with new updates in the function this test is failing. The function now returns an error early if it can't get the object, and that's why the test should stop trying to update if there's an error.

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
everettraven: If the lister has not yet synced, will this sync loop even run? My understanding is that in general we wait for our listers to be synced before controllers actually run.

I believe we can solely rely on the lister containing the information we care about. If it does not, doing a live GET is unlikely to give us different information.

pkg/controllers/oauthclientscontroller/oauthclientscontroller_test.go
everettraven: It looks like the pre-existing test code should have handled that scenario? It looks to me like if the update fails it tries to ensure the existing resource has not been modified, not continue to update it.

What is the failure that you are seeing here?

kramaranya: I was getting `got error: get oauthclients fake error; want error: false` because the old chnages triggered this fake get error https://github.com/openshift/cluster-authentication-operator/blob/a53f649d6ac1423a602786a165640c77a66d17e3/pkg/controllers/oauthclientscontroller/oauthclientscontroller_test.go#L321-L323 so my quick fix was to stop trying to update after that fake error. But as we switched to a lister, which doesnt trigger that fake error, I reverted that change and test now passes

pkg/controllers/oauthclientscontroller/oauthclientscontroller.go
kramaranya: That makes sense, thanks a lot for pointing it out

everettraven (approved):
/lgtm

PR #759, opened 2025-02-25
OCPEDGE-1610: bump(kube,library-go,ocp/api):

eggfoobar:
bumping kubernetes v0.32.2, library-go and ocp/api to latest 

we will need to update the openshift/api soon which will include additional API needed for DualReplica topology, bumping needed dependencies to avoid too many changes in a single PR


eggfoobar:
/label acknowledge-critical-fixes-only

Needed to unblock testing TNF

PR #760, opened 2025-02-25
OCPBUGS-51981: fix: update starter path to account for mom integration

eggfoobar:
updatedinitialization to parameterize feature gate accessor to allow mom to call the binary in CI
updated test to include OIDC check

### DIFF
--- a/pkg/operator/replacement_starter.go
+++ b/pkg/operator/replacement_starter.go
@@ -18,6 +18,8 @@ import (
 
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 
+	ocpconfigv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	configinformer "github.com/openshift/client-go/config/informers/externalversions"
@@ -30,6 +32,7 @@ import (
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/manifestclient"
 	libgoetcd "github.com/openshift/library-go/pkg/operator/configobserver/etcd"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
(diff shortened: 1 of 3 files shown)


pkg/operator/replacement_starter.go
```diff
@@ -339,3 +344,33 @@ func CreateOperatorStarter(ctx context.Context, authOperatorInput *authenticatio
...
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		featureGates, _ = featureGateAccessor.CurrentFeatureGates()
+	case <-time.After(1 * time.Minute):
+		return nil, fmt.Errorf("timed out waiting for FeatureGate detection")
+	}
+	return featureGates, nil
+}
+
+func momFeatureGateAccessor() featureGateAccessorFunc {
+	enabled := []ocpconfigv1.FeatureGateName{"ExternalOIDC"}
```
liouk, line 371: Use `features.FeatureGateExternalOIDC` instead of hardcoding the name.

pkg/operator/starter.go
```diff
@@ -681,22 +680,9 @@ func prepareExternalOIDC(
 	informerFactories authenticationOperatorInformerFactories,
 ) ([]libraryapplyconfiguration.NamedRunOnce, []libraryapplyconfiguration.RunFunc, error) {
 
-	// By default, this will exit(0) if the featuregates change
```
liouk, line 684: Let's keep this comment; it's not immediately obvious.

pkg/operator/replacement_starter.go
```diff
@@ -140,6 +143,7 @@ func CreateOperatorInputFromMOM(ctx context.Context, momInput libraryapplyconfig
 		apiextensionClient:           apiextensionClient,
 		eventRecorder:                eventRecorder,
 		clock:                        momInput.Clock,
+		featureGateAccessor:          momFeatureGateAccessor(),
```
liouk, line 146: It might make the creation of `authenticationOperatorInput` clearer if the enabled/disabled featuregates are obvious on the constructor; something like:
```suggestion
		featureGateAccessor:          staticFeatureGateAccessor([]ocpconfigv1.FeatureGateName{features.FeatureGateExternalOIDC}, []ocpconfigv1.FeatureGateName{}),


```

Just an idea though, I'll leave it up to you to decide what you prefer.

pkg/operator/replacement_starter.go
eggfoobar: Good idea, updated.

pkg/operator/starter.go
eggfoobar: Sure thing, re-added.

PR #762, opened 2025-03-11
OCPBUGS-52843: externaloidc: use `sub` claim for UID

everettraven:
To ensure that components that expect a `UserInfo` object to have a UID continue to behave as expected when the external OIDC functionality is enabled, we need to explicitly set the UID.

Currently, it is not possible for users to configure this claim mapping using the `authentications.config.openshift.io` CRD, so we are defaulting to the `sub` claim that should be present on all issued JWTs by OIDC compatible IdPs.


xingxingxia:
From test perspective, pre-merge test result of https://issues.redhat.com/browse/OCPBUGS-52843?focusedId=26772156&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26772156 looks good, the bug is fixed.
/label qe-approved

PR #769, opened 2025-06-04
CNTRLPLANE-939: test/e2e-oidc: poll test preconditions for a short time instead of checking once

liouk:
While running rehearsals in the scope of https://github.com/openshift/release/pull/65561, an issue with the test preconditions check was identified, where the check happened too quickly after a change and did not allow enough time for the preconditions to get fulfilled.

This PR introduces a short polling for the preconditions and will thus allow the operators to stabilize before checking.


everettraven (approved):
/lgtm

PR #772, opened 2025-06-20
DO-NOT-MERGE: Prove library-go#1902

liouk:
This is a proof PR for https://github.com/openshift/library-go/pull/1902.

It also demonstrates how the workload deletion will be used in the context of external OIDC, where we need to delete the oauth-server & oauth-apiserver deployments.


liouk:
Proof PR, holding.

/hold

PR #775, opened 2025-07-23
OCPBUGS-57808, AUTH-543: (deps): Bump to Go 1.24 and k8s 1.33

everettraven:
Bumps go to 1.24 and k8s 1.33.

Builds on #771 to solve CI failures.

Needed to fix CI issues in #774 and progress the library-go changes it is testing further


.gitignore
```diff
@@ -20,3 +20,4 @@ tags
 authentication-operator
 telepresence.log
 /test-output/
+_output
```
everettraven, line 23: Added to ignore the output results from `make verify`

PR #779, opened 2025-08-11
WIP: Add retry logic for WellKnownAvailable endpoint checks

sdodson:
The operator now retries the well-known endpoint check every second for up to 15 seconds before setting itself unavailable. This prevents premature "unavailable" status during temporary network issues or API server startup scenarios.

Changes:
- Add retry loop with 15-second timeout and 1-second intervals
- Preserve existing ControllerProgressingError handling logic
- Set WellKnownAvailable=False with reason "NotReady" after retries exhausted
- Maintain proper progressing status during retry attempts

This change improves operator reliability by giving well-known endpoints time to become ready while maintaining the same final error handling behavior.

Assisted-by: Cursor, Claude Sonnet 4

Related to https://issues.redhat.com/browse/OCPBUGS-20056 but since this only addresses one of the paths I'm not linking this until I can prove whether or not this seems to help.

### DIFF
--- a/pkg/controllers/readiness/wellknown_ready_controller.go
+++ b/pkg/controllers/readiness/wellknown_ready_controller.go
@@ -169,13 +169,51 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


sdodson:
/payload-test periodic-ci-openshift-release-master-ci-4.20-e2e-aws-upgrade-ovn-single-node

PR #781, opened 2025-08-12
OCPBUGS-59909: Don't cache server errors when checking for password grant support

jacobsee:
~~If we encounter an OIDC server which is not properly responding using JSON, we should occasionally retry pinging it for the purpose of checking password grant support rather than assuming that it doesn't support password grant forever.~~

Now: Reject and don't cache 5xx errors outright while checking for password grant support.

### DIFF
--- a/pkg/controllers/configobservation/oauth/idp_conversions.go
+++ b/pkg/controllers/configobservation/oauth/idp_conversions.go
@@ -425,6 +425,10 @@ func checkOIDCPasswordGrantFlow(
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode >= 500 && resp.StatusCode <= 599 {
+		return false, fmt.Errorf("OIDC token endpoint returned server error %d (%s)", resp.StatusCode, tokenURL)
+	}
+
 	respJSON := json.NewDecoder(resp.Body)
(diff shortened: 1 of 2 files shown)


pkg/controllers/configobservation/oauth/idp_conversions.go
```diff
@@ -425,6 +425,10 @@ func checkOIDCPasswordGrantFlow(
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode >= 500 && resp.StatusCode < 600 {
```
liouk, line 428: nit: 600 reads weird in http :sweat_smile: Would you mind using `<=599` instead?

liouk (changes requested):
A couple of comments on the commit structure:
- we don't need the new caching mechanism, so no point in adding it and then removing it in separate commits; can you please drop those and keep only the commit for the 5xx?
- please don't prefix commits with the jira; we have this info in the PR. Use a component prefix instead, i.e. `configobservation/oauth:`

pkg/controllers/configobservation/oauth/idp_conversions.go
jacobsee: Sure thing 😄

PR #815, opened 2025-12-03
Revert "OCPBUGS-65675: externaloidc: return errors when node statuses cannot be used to determine oidc state"

neisw:
Reverts openshift/cluster-authentication-operator#801

Testing revert for [TRT-2460](https://issues.redhat.com/browse/TRT-2460)


neisw:
/hold
not sure this is related, hold while testing

PR #825, opened 2026-01-23
CNTRLPLANE-2610: Create network policies for AUTH components

liouk:
This PR adds network policies to the authentication operator, oauth-server and oauth-apiserver. For each component, there are two policies:
- one to allow all Ingress/Egress known traffic for each component
- one to deny all other Ingress/Egress traffic for each component

All known and required connections must be reflected to respective allow rules.

Note that, in case of pods that require traffic to/from hostNetwork pods (such as the kube-apiserver), we need to allow all ingress/egress TCP traffic; NetworkPolicies do not affect pods on hostNetwork, but we still need a rule to allow ingress/egress from/to them.

In some cases there might be some overlap in the policy rules, but this is intentional for the sake of documentation/future reference.

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes NetworkPolicy resources to enforce network security rules for OAuth and authentication components across multiple namespaces.
  * Implemented default-deny network policies restricting ingress and egress traffic unless explicitly permitted between authorized services.


liouk:
@coderabbitai review

pkg/cmd/mom/output_resources_command.go
```diff
@@ -41,6 +41,13 @@ func runOutputResources(ctx context.Context) (*libraryoutputresources.OutputReso
 				libraryoutputresources.ExactRoleBinding("openshift-config-managed", "system:openshift:oauth-servercert-trust"),
 
 				libraryoutputresources.ExactPDB("openshift-oauth-apiserver", "oauth-apiserver-pdb"),
+
+				libraryoutputresources.ExactResource("networking.k8s.io", "v1", "networkpolicies", "openshift-authentication-operator", "authentication-operator-networkpolicy"),
+				libraryoutputresources.ExactResource("networking.k8s.io", "v1", "networkpolicies", "openshift-authentication-operator", "default-deny-all"),
```
everettraven: Does MOM need to also know the manifests for deploying the cluster-authentication-operator itself?

everettraven (approved):
This LGTM in general.

We should probably run payload jobs for a sanity check that this won't cause component readiness issues and payload build failures.

/lgtm

dusk125 (changes requested):

PR #838, opened 2026-02-06
CNTRLPLANE-2247:Add Test encryption provider migration

gangwgr:
[CNTRLPLANE-2247](https://issues.redhat.com//browse/CNTRLPLANE-2247):Add Test encryption provider migration
This PR brings TestEncryptionProvidersMigration scenario in https://github.com/openshift/library-go/blob/7bced6e899b65175945fc309a2e125de49cac3dc/test/library/encryption/scenarios.go#L161 in here.


gangwgr:
> @gangwgr [openshift/cluster-kube-apiserver-operator#2028](https://github.com/openshift/cluster-kube-apiserver-operator/pull/2028) has merged. You can update this PR with the latest changes.

updated

ardaguclu:
@gangwgr that would be great if we can add this into cluster-openshift-apiserver-operator too.

PR #844, opened 2026-02-24
[release-4.21] OCPBUGS-74418: Add KMS test scenarios

gangwgr:
Backport kms test cases  to 4.21


test/e2e-encryption-kms/main_test.go
```diff
@@ -0,0 +1,31 @@
...
+	pointerVal := reflect.ValueOf(m)
+	val := reflect.Indirect(pointerVal)
+
+	testsMember := val.FieldByName("tests")
+	ptrToTests := unsafe.Pointer(testsMember.UnsafeAddr())
+	realPtrToTests := (*[]testing.InternalTest)(ptrToTests)
+
+	tests := *realPtrToTests
+
+	rand.Seed(time.Now().UnixNano())
```
liouk: nit: `rand.Seed()` is a no-op on 1.24, which is the current go version unless I'm mistaken -- we could drop this.

PR #858, opened 2026-03-27
CNTRLPLANE-2523: deploy oauth-apiserver in new OIDC mode when authentication type is OIDC

everettraven:
# Description

This PR updates the cluster-authentication-operator to deploy the oauth-apiserver in the new external OIDC mode when the authentication type is set to `OIDC` _and_ the new `ExternalOIDCExternalClaimsSourcing` feature gate is enabled.

Currently, this feature gate is only enabled in DevPreviewNoUpgrade.

For more information as to the motivation of this change, see https://github.com/openshift/enhancements/pull/1907

## Release Notes

* **New Features**
  * Introduced new external OIDC authentication architecture with dedicated api-server deployment option
  * Added feature gate to enable/disable new OIDC external claims sourcing capability
  * Integrated automatic configuration synchronization for external OIDC scenarios

* **Refactor**
  * Updated OIDC availability detection to support both legacy and new architecture paths

### DIFF
--- a/bindata/oauth-apiserver/externaloidc-deploy.yaml
+++ b/bindata/oauth-apiserver/externaloidc-deploy.yaml
@@ -1,0 +1,109 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: openshift-oauth-apiserver
+  name: apiserver
+  labels:
+    app: openshift-oauth-apiserver
+    apiserver: "true"
+# The number of replicas will be set in code to the number of master nodes.
+spec:
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      # To ensure that only one pod at a time writes to the node's
+      # audit log, require the update strategy to proceed a node at a
+      # time. Only when a master node has its existing
+      # oauth-apiserver pod stopped will a new one be allowed to
+      # start.
+      maxUnavailable: 1
+      maxSurge: 0
+  selector:
+    matchLabels:
+      app: openshift-oauth-apiserver
+      apiserver: "true"
+  template:
+    metadata:
+      name: openshift-oauth-apiserver
+      labels:
+        app: openshift-oauth-apiserver
+        apiserver: "true"
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: privileged
+    spec:
+      serviceAccountName: oauth-apiserver-sa
+      priorityClassName: system-node-critical
+      containers:
+      - name: oauth-apiserver
+        terminationMessagePolicy: FallbackToLogsOnError
+        image: ${IMAGE}
+        imagePullPolicy: IfNotPresent
+        command: ["/bin/bash", "-ec"]
+        args:
+          - |
+            if [ -s /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem ]; then
+              echo "Copying system trust bundle"
+              cp -f /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+            fi
+            exec oauth-apiserver external-oidc \
+              --secure-port=8443 \
+              --tls-private-key-file=/var/run/secrets/serving-cert/tls.key \
+              --tls-cert-file=/var/run/secrets/serving-cert/tls.crt \
+              --config=/var/run/configmaps/auth-config/auth-config.json \
+              ${FLAGS}
+        resources:
+          requests:
+            memory: 200Mi
+            cpu: 150m
+        # we need to set this to privileged to be able to copy trust bundles to /etc/pki/ca-trust/extracted/pem/...
+        securityContext:
+          privileged: true
+          runAsUser: 0
+        ports:
+        - containerPort: 8443
+        volumeMounts:
+        - mountPath: /var/run/configmaps/trusted-ca-bundle
+          name: trusted-ca-bundle
+        - mountPath: /var/run/secrets/serving-cert
+          name: serving-cert
+        - mountPath: /var/run/configmaps/auth-config
+          name: auth-config
+      terminationGracePeriodSeconds: 120 # a bit more than the 60 seconds timeout of non-long-running requests + the shutdown delay
+      volumes:
+      - name: serving-cert
+        secret:
+          secretName: serving-cert
+      - name: trusted-ca-bundle
+        configMap:
+          name: trusted-ca-bundle
+          optional: true
+          items:
+          - key: ca-bundle.crt
+            path: tls-ca-bundle.pem
+      - name: auth-config
+        configMap:
+          name: auth-config
+          items:
+          - key: auth-config.json
+            path: auth-config.json
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      tolerations:
+        # Ensure pod can be scheduled on master nodes
+        - key: "node-role.kubernetes.io/master"
+          operator: "Exists"
+          effect: "NoSchedule"
+          # Ensure pod can be evicted if the node is unreachable
+        - key: "node.kubernetes.io/unreachable"
+          operator: "Exists"
+          effect: "NoExecute"
+          tolerationSeconds: 120
+          # Ensure scheduling is delayed until node readiness
+          # (i.e. network operator configures CNI on the node)
+        - key: "node.kubernetes.io/not-ready"
+          operator: "Exists"
+          effect: "NoExecute"
+          tolerationSeconds: 120
+      # Anti-affinity is configured in code due to the need to scope
+      # selection to the computed pod template.
--- a/pkg/controllers/common/external_oidc.go
+++ b/pkg/controllers/common/external_oidc.go
@@ -5,11 +5,13 @@ import (
 	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions/operator/v1"
 	operatorv1listers "github.com/openshift/client-go/operator/listers/operator/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	corev1informers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/tools/cache"
 
@@ -19,23 +21,30 @@ import (
 )
 
 type AuthConfigChecker struct {
-	authenticationsInformer        cache.SharedIndexInformer
-	kubeAPIServersInformer         cache.SharedIndexInformer
-	kasNamespaceConfigMapsInformer cache.SharedIndexInformer
+	authenticationsInformer         cache.SharedIndexInformer
+	kubeAPIServersInformer          cache.SharedIndexInformer
+	kasNamespaceConfigMapsInformer  cache.SharedIndexInformer
+	oaasNamespaceConfigMapsInformer cache.SharedIndexInformer
 
-	authLister         configv1listers.AuthenticationLister
-	kasLister          operatorv1listers.KubeAPIServerLister
-	kasConfigMapLister corelistersv1.ConfigMapLister
+	authLister          configv1listers.AuthenticationLister
+	kasLister           operatorv1listers.KubeAPIServerLister
+	kasConfigMapLister  corelistersv1.ConfigMapLister
+	oaasConfigMapLister corelistersv1.ConfigMapLister
+
+	featureGateAccessor featuregates.FeatureGateAccess
 }
 
-func NewAuthConfigChecker(authentications configv1informers.AuthenticationInformer, kubeapiservers operatorv1informers.KubeAPIServerInformer, configmaps corev1informers.ConfigMapInformer) AuthConfigChecker {
+func NewAuthConfigChecker(authentications configv1informers.AuthenticationInformer, kubeapiservers operatorv1informers.KubeAPIServerInformer, kasConfigMaps, oaasConfigMaps corev1informers.ConfigMapInformer, featureGateAccessor featuregates.FeatureGateAccess) AuthConfigChecker {
 	return AuthConfigChecker{
-		authenticationsInformer:        authentications.Informer(),
-		kubeAPIServersInformer:         kubeapiservers.Informer(),
-		kasNamespaceConfigMapsInformer: configmaps.Informer(),
-		authLister:                     authentications.Lister(),
-		kasLister:                      kubeapiservers.Lister(),
-		kasConfigMapLister:             configmaps.Lister(),
+		authenticationsInformer:         authentications.Informer(),
+		kubeAPIServersInformer:          kubeapiservers.Informer(),
+		kasNamespaceConfigMapsInformer:  kasConfigMaps.Informer(),
+		oaasNamespaceConfigMapsInformer: oaasConfigMaps.Informer(),
+		authLister:                      authentications.Lister(),
+		kasLister:                       kubeapiservers.Lister(),
+		kasConfigMapLister:              kasConfigMaps.Lister(),
+		oaasConfigMapLister:             oaasConfigMaps.Lister(),
+		featureGateAccessor:             featureGateAccessor,
 	}
 }
 
@@ -48,6 +57,7 @@ func AuthConfigCheckerInformers[T factory.Informer](c *AuthConfigChecker) []T {
 		c.authenticationsInformer.(T),
 		c.kubeAPIServersInformer.(T),
 		c.kasNamespaceConfigMapsInformer.(T),
+		c.oaasNamespaceConfigMapsInformer.(T),
 	}
 }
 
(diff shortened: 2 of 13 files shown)


pkg/operator/workload/sync_openshift_oauth_apiserver_test.go
everettraven, line 1: TODO: Add tests here

bindata/oauth-apiserver/externaloidc-deploy.yaml
```diff
@@ -0,0 +1,116 @@
...
+      - name: trusted-ca-bundle
+        configMap:
+          name: trusted-ca-bundle
+          optional: true
+          items:
+          - key: ca-bundle.crt
+            path: tls-ca-bundle.pem
+      - name: auth-config
+        configMap:
+          name: auth-config
```
everettraven, line 86: TODO: Verify if this needs to be revisioned.

pkg/operator/starter.go
```diff
@@ -447,14 +453,48 @@ func prepareOauthAPIServerOperator(
...
 		informerFactories.operatorInformer.Operator().V1().KubeAPIServers(),
 		informerFactories.kubeInformersForNamespaces.InformersFor("openshift-kube-apiserver").Core().V1().ConfigMaps(),
+		featureGateAccessor,
 	)
 
-	featureGateAccessor, err := authOperatorInput.featureGateAccessor(ctx, authOperatorInput, informerFactories)
-	if err != nil {
+	// conditionally sync the auth-config configmap from the openshift-config-managed namespace
+	// to the openshift-oauth-apiserver if OIDC is available.
+	// This precondition never returns an error to prevent unnecesary degradation.
```
everettraven: TODO: decide whether or not this is reasonable over updating the external oidc controller to just write the config directly to the `openshift-oauth-apiserver` namespace.

bindata/oauth-apiserver/externaloidc-deploy.yaml
everettraven: Doesn't seem like it.

bindata/oauth-apiserver/externaloidc-deploy.yaml
```diff
@@ -0,0 +1,116 @@
...
+          # used to emit shutdown lifecycles events in the correct namespace
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          # used to emit shutdown lifecycles events in the correct namespace
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
```
everettraven: From what I can tell, these won't be used in the new mode of operation and can safely be removed.

pkg/operator/starter.go
everettraven: For now, I think that syncing from the `openshift-config-managed` namespace to the `openshift-oauth-apiserver` namespace is a "cleaner" approach to prevent adding more complexity to the existing external oidc controller that generates the structured auth config file while we have the two different configuration paths based on feature gate state.

bindata/oauth-apiserver/externaloidc-deploy.yaml
```diff
@@ -0,0 +1,104 @@
...
+          optional: true
+          items:
+          - key: ca-bundle.crt
+            path: tls-ca-bundle.pem
+      - name: auth-config
+        configMap:
+          name: auth-config
+          items:
+          - key: auth-config.json
+            path: auth-config.json
```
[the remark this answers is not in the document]
everettraven: This is intentionally not marked as optional because the `oauth-apiserver external-oidc` command requires this configuration file to be present to run successfully.

This deployment is only created under the same conditions in which the `auth-config` configmap is synced.

bindata/oauth-apiserver/externaloidc-deploy.yaml
```diff
@@ -0,0 +1,104 @@
...
+            if [ -s /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem ]; then
+              echo "Copying system trust bundle"
+              cp -f /var/run/configmaps/trusted-ca-bundle/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+            fi
+            exec oauth-apiserver external-oidc \
+              --secure-port=8443 \
+              --tls-private-key-file=/var/run/secrets/serving-cert/tls.key \
+              --tls-cert-file=/var/run/secrets/serving-cert/tls.crt \
+              --config=/var/run/configmaps/auth-config/auth-config.json \
+              ${FLAGS}
```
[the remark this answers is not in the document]
everettraven: I seemed to be able to run a cluster in this configuration just fine, but I could have missed something.

Will investigate this further tomorrow to try and identify if this is really an issue or if the trust bundle is not getting synced properly.

bindata/oauth-apiserver/externaloidc-deploy.yaml
```diff
@@ -0,0 +1,104 @@
...
+      app: openshift-oauth-apiserver
+      apiserver: "true"
+  template:
+    metadata:
+      name: openshift-oauth-apiserver
+      labels:
+        app: openshift-oauth-apiserver
+        apiserver: "true"
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
```
liouk, line 32: We should still pin the required SCC for these pods; `restricted-v2` by the looks of it?

pkg/operator/workload/sync_openshift_oauth_apiserver.go
```diff
@@ -277,6 +323,97 @@ func (c *OAuthAPIServerWorkload) syncDeployment(ctx context.Context, operatorSpe
 	return deployment, err
 }
 
+func (c *OAuthAPIServerWorkload) syncExternalOIDCDeployment(ctx context.Context, operatorSpec *operatorv1.OperatorSpec, operatorStatus *operatorv1.OperatorStatus, eventRecorder events.Recorder) (*appsv1.Deployment, error) {
```
liouk, line 326: It looks like there's lots of common code between this and `syncStandardDeployment()`, is there a way we can extract a common helper maybe?

Disclaimer: this was picked up by claude.

bindata/oauth-apiserver/externaloidc-deploy.yaml
```diff
@@ -0,0 +1,104 @@
...
+        secret:
+          secretName: serving-cert
+      - name: trusted-ca-bundle
+        configMap:
+          name: trusted-ca-bundle
+          optional: true
+          items:
+          - key: ca-bundle.crt
+            path: tls-ca-bundle.pem
+      - name: auth-config
```
liouk, line 84: Not entirely familiar with volumes, but why are we not marking this one as optional, like `trusted-ca-bundle`? Could this lead to issues when the configmap isn't available yet?

liouk (changes requested):

bindata/oauth-apiserver/externaloidc-deploy.yaml
everettraven: Based on another comment from coderabbit I think this will still need to be privileged because of the trusted-ca-bundle copying logic

pkg/operator/workload/sync_openshift_oauth_apiserver.go
everettraven: There is, but there is subtle enough differences that I felt it was worth carrying two different paths instead of trying to carry a common helper that has to account for all the nuances of correctly configuring both deployment modes. I suspect they will end up drifting even further apart in the future as certain functionality may only apply to one deployment mode (i.e applicable to an aggregated apiserver, but not the standard HTTP server that runs for the external OIDC mode)

bindata/oauth-apiserver/externaloidc-deploy.yaml
everettraven: Coderabbit had a similar comment here: https://github.com/openshift/cluster-authentication-operator/pull/858#discussion_r3053588480

My thinking for this not being optional is because the `oauth-apiserver external-oidc` command expects a configuration file to be present. The command will fail out if it isn't able to resolve a configuration correctly. IMO, a pending Pod waiting for existence of the `auth-config` ConfigMap is better than a Pod that is crash looping. Additionally, this deployment and the `auth-config` ConfigMap should be provisioned only under the exact same conditions. 

IIUC, `trusted-ca-bundle` is optional because it is only applicable if an end-user is configuring some custom trust bundles they want to be applied broadly across the cluster.

Related to this, one thing I've noticed as part of updating the e2e tests that I need to account for is to not attempt to stamp out this new deployment and clean up old resources if the `auth-config` generation is not successful. This will prevent a scenario where an invalid configuration completely breaks the authentication layer.

pkg/operator/workload/sync_openshift_oauth_apiserver.go
liouk: Sounds good :+1: Let's leave it as is.

bindata/oauth-apiserver/externaloidc-deploy.yaml
liouk: > IMO, a pending Pod waiting for existence of the auth-config ConfigMap is better than a Pod that is crash looping.

Totally agree -- since that's the behavior when the volume is required, then this is definitely the way to go :+1:

liouk: And yes, I think this is the reasoning behind the `trusted-ca-bundle` -- I just wasn't entirely clear on what would happen with the required volume when the actual file isn't there :+1:

pkg/operator/workload/sync_openshift_oauth_apiserver.go
```diff
@@ -277,6 +323,97 @@ func (c *OAuthAPIServerWorkload) syncDeployment(ctx context.Context, operatorSpe
...
+
+	// we watch some resources so that our deployment will redeploy without explicitly and carefully ordered resource creation
+	inputHashes, err := resourcehash.MultipleObjectHashStringMapForObjectReferences(
+		ctx,
+		c.kubeClient,
+		resourcehash.NewObjectRef().ForConfigMap().InNamespace(c.targetNamespace).Named("trusted-ca-bundle"),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("invalid dependency reference: %q", err)
+	}
```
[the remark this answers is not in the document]
everettraven: Not including this in the hash is intentional because the `oauth-apiserver external-oidc` command supports hot-reloads of the configuration file that this configmap mounts.

Because non subPath ConfigMap volume mounts automatically update (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically for details), we can rely on the hot-reload functionality to ensure that updates are reflected in the command behavior.

Having this hot reload capability means we don't need to modify the deployment to trigger a new rollout like is needed for the `trusted-ca-bundle` (because those are only copied to the system trust bundle once at startup)

pkg/controllers/common/external_oidc.go
```diff
@@ -68,12 +78,41 @@ func (c *AuthConfigChecker) OIDCAvailable() (bool, error) {
...
+	// Doing so ensures that any errors encountered during the generation of the authentication configuration
+	// file doesn't cause the entirety of our authentication stack from falling over.
+	if featureGates.Enabled(features.FeatureGateExternalOIDCExternalClaimsSourcing) {
+		_, err := c.oaasConfigMapLister.ConfigMaps("openshift-oauth-apiserver").Get("auth-config")
+		if errors.IsNotFound(err) {
+			return false, nil
+		} else if err != nil {
+			return false, fmt.Errorf("getting configmap openshift-oauth-apiserver/auth-config: %w", err)
+		}
+		return true, nil
```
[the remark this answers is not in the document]
everettraven: Fixed.

pkg/operator/replacement_starter.go
```diff
@@ -143,7 +143,7 @@ func CreateOperatorInputFromMOM(ctx context.Context, momInput libraryapplyconfig
 		apiextensionClient:           apiextensionClient,
 		eventRecorder:                eventRecorder,
 		clock:                        momInput.Clock,
-		featureGateAccessor:          staticFeatureGateAccessor([]ocpconfigv1.FeatureGateName{features.FeatureGateExternalOIDC}, []ocpconfigv1.FeatureGateName{features.FeatureGateKMSEncryption}),
+		featureGateAccessor:          staticFeatureGateAccessor([]ocpconfigv1.FeatureGateName{features.FeatureGateExternalOIDC}, []ocpconfigv1.FeatureGateName{features.FeatureGateKMSEncryption, features.FeatureGateExternalOIDCExternalClaimsSourcing}),
```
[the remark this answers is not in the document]
everettraven: MOM isn't actually used in production and as far as I can tell the gates here are meant to distinguish what is on by default and what is not. I'm following precedent here, so this should be OK.

everettraven:
Manually verified that the changes work as expected and got a clean run of the oidc e2e tests on a devpreview enabled clusterbot cluster.

/verified by @everettraven

pkg/controllers/common/external_oidc.go
```diff
@@ -74,6 +97,26 @@ func (c *AuthConfigChecker) OIDCAvailable() (bool, error) {
...
```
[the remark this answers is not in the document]
everettraven: Updated!

pkg/controllers/common/external_oidc_test.go
ehearne-redhat, line 31: Just checking if there is a reason why we don't have `oaasInformerSynced` bool ?

ehearne-redhat (commented):
I have one concern about not checking `oaasInformerSynced` not being included in testing, but other than that changes look fine.

ehearne-redhat:
@everettraven What I would also like to know is if you know why some tests are failing? I can see they are also failing for others too.

pkg/controllers/common/external_oidc_test.go
everettraven: `cmInformerSynced` is the thing that is used for both the oauth-apiserver and kube-apiserver namespaced configmap informers.

This means the new line in `external_oidc.go` that checks if the oauth-apiserver namespaced configmap informer has synced technically would never have it's error return path validated as part of our tests. I didn't feel like covering that path was all that necessary as it is pretty straightforward behavior.

`kasInformerSynced` is a separate informer entirely to watch for changes to the [`KubeAPIServer` type](https://github.com/openshift/api/blob/464776f9520793f0f20645fbcb82d945abb0a05b/operator/v1/types_kubeapiserver.go#L20) that is used to communicate the state of the kube-apiserver configuration. This was necessary for the older architecture so that we could ensure that all instances of the kube-apiserver were successfully configured and running with the new OIDC authentication mode before removing the oauth-apiserver.

There isn't a need for us to do this same check here because the oauth-apiserver will always be deployed and all kube-apiserver instances will be configured to communicate with it.

everettraven:
> @everettraven What I would also like to know is if you know why some tests are failing? I can see they are also failing for others too.

All the tests that contain `encryption` in the name are optional and seem to be related to the KMS feature that is currently in development. They seem to be failing due to issues related to the work to add/migrate tests for this feature to the openshift-tests-extension pattern.

The required tests that are failing are:
- `e2e-oidc` and `e2e-oidc-techpreview` - From what it looks like, these are also related to wider CI issues but I am going to run these tests against a fresh non-devpreview 4.22 cluster to verify it doesn't have to do with the changes I've made in this PR.
- `e2e-agnostic` - This one looks to be failing due to the known kubelet version issue affecting pre-submit jobs.

pkg/operator/workload/sync_openshift_oauth_apiserver.go
```diff
@@ -18,6 +18,8 @@ import (
 	appsv1listers "k8s.io/client-go/listers/apps/v1"
 	"k8s.io/klog/v2"
 
+	v1 "github.com/openshift/api/config/v1"
```
liouk: nit: this is usually aliased as `configv1`

```suggestion
	configv1 "github.com/openshift/api/config/v1"
```

pkg/operator/starter.go
```diff
@@ -447,14 +454,49 @@ func prepareOauthAPIServerOperator(
...
 		informerFactories.kubeInformersForNamespaces.InformersFor("openshift-kube-apiserver").Core().V1().ConfigMaps(),
+		informerFactories.kubeInformersForNamespaces.InformersFor("openshift-oauth-apiserver").Core().V1().ConfigMaps(),
+		featureGateAccessor,
 	)
 
-	featureGateAccessor, err := authOperatorInput.featureGateAccessor(ctx, authOperatorInput, informerFactories)
-	if err != nil {
+	// conditionally sync the auth-config configmap from the openshift-config-managed namespace
+	// to the openshift-oauth-apiserver if OIDC is available and FeatureGateExternalOIDCExternalClaimsSourcing is enabled.
+	// This precondition never returns an error to prevent unnecesary degradation.
```
liouk: nit/typo

```suggestion
	// This precondition never returns an error to prevent unnecessary degradation.
```

liouk (changes requested):

ehearne-redhat:
thanks for the explanation @everettraven ! 

/lgtm

liouk:
The `e2e-agnostic` job is failing due to a [known ingress component readiness regression](https://sippy.dptools.openshift.org/sippy-ng/component_readiness/test_details?Architecture=amd64&FeatureSet=default&Installer=ipi&LayeredProduct=none&Network=ovn&Platform=aws&Suite=unknown&Topology=ha&Upgrade=micro&baseEndTime=2026-02-03T23%3A59%3A59Z&baseRelease=4.21&baseStartTime=2026-01-04T00%3A00%3A00Z&capability=Other&columnGroupBy=Network%2CPlatform%2CTopology&component=Networking+%2F+router&confidence=95&dbGroupBy=Architecture%2CFeatureSet%2CInstaller%2CLayeredProduct%2CNetwork%2CPlatform%2CSuite%2CTopology%2CUpgrade&environment=Architecture%3Aamd64+FeatureSet%3Adefault+Installer%3Aipi+LayeredProduct%3Anone+Network%3Aovn+Platform%3Aaws+Suite%3Aunknown+Topology%3Aha+Upgrade%3Amicro&flakeAsFailure=false&ignoreDisruption=true&ignoreMissing=false&includeMultiReleaseAnalysis=true&includeVariant=Architecture%3Aamd64&includeVariant=Architecture%3Aarm64&includeVariant=Architecture%3Amulti&includeVariant=CGroupMode%3Av2&includeVariant=ContainerRuntime%3Acrun&includeVariant=ContainerRuntime%3Arunc&includeVariant=FeatureSet%3Adefault&includeVariant=FeatureSet%3Atechpreview&includeVariant=Installer%3Ahypershift&includeVariant=Installer%3Aipi&includeVariant=Installer%3Aupi&includeVariant=JobTier%3Ablocking&includeVariant=JobTier%3Ainforming&includeVariant=JobTier%3Astandard&includeVariant=LayeredProduct%3Anone&includeVariant=LayeredProduct%3Avirt&includeVariant=Network%3Aovn&includeVariant=OS%3Arhcos9&includeVariant=Owner%3Aeng&includeVariant=Owner%3Aservice-delivery&includeVariant=Platform%3Aaws&includeVariant=Platform%3Aazure&includeVariant=Platform%3Agcp&includeVariant=Platform%3Ametal&includeVariant=Platform%3Avsphere&includeVariant=Topology%3Aexternal&includeVariant=Topology%3Aha&includeVariant=Topology%3Amicroshift&includeVariant=Upgrade%3Amicro&includeVariant=Upgrade%3Aminor&includeVariant=Upgrade%3Anone&minFail=3&passRateAllTests=0&passRateNewTests=95&pity=5&sampleEndTime=2026-04-17T12%3A00%3A00Z&sampleRelease=4.22&sampleStartTime=2026-04-10T00%3A00%3A00Z&testId=openshift-tests%3A158feae642f0200bf3dd69e9f579f6ff&testLifecycles=blocking&view=4.22-main). Since this is the only failure in the most recent run of the job, but also because the changes in this PR are gated in DevPreviewNoUpgrade, we're overriding that job.

/override ci/prow/e2e-agnostic

PR #859, opened 2026-03-27
CNTRLPLANE-2589: Migrate to OTE support,fix pod security,Pathological Event monitor failures

ropatil010:
Hi Team,

PR about: Migrates authentication operator e2e tests to OpenShift Tests Extension (OTE) framework and fixes three CI monitor test failures: PodSecurityViolation, known-image-checker, and pathological events.

1. OTE Framework Migration
```
### Dual-Mode Test Pattern
Tests now support both `go test` and Ginkgo (OTE) execution simultaneously:
- **Test logic** in `.go` files uses `testing.TB` interface (compatible with both frameworks)
- **Go test wrappers** in `*_test.go` for backward compatibility
- **Ginkgo registration** in `.go` files for OTE integration

### Migrated Tests
**Encryption Tests:**
- `test/e2e-encryption/` - Created encryption.go, refactored encryption_test.go
- `test/e2e-encryption-rotation/` - Created encryption_rotation.go, refactored test file
- `test/e2e-encryption-perf/` - Created encryption_perf.go, refactored test file
- `test/e2e-oidc/` - Extracted 1133 lines from external_oidc_test.go into external_oidc.go

### OTE Registration in main.go
**Test suites registered:**
- `operator-encryption/serial` - **ClusterStability: Disruptive** (triggers OAuth API server rollouts)
- `operator-encryption-rotation/serial` - **ClusterStability: Disruptive**
- `operator-encryption-perf/serial` - **ClusterStability: Disruptive**
- Excluded `[Disruptive]` tests from regular serial suite

### New Test Libraries
**1. `test/library/encryption/scenarios.go`** - Centralized encryption test scenarios (TestEncryptionTypeIdentity, TestEncryptionTypeUnset, TestEncryptionTurnOnAndOff)
**2. `test/library/encryption/perf_helpers.go`** - Performance test utilities
**3. `test/library/encryption_wrappers.go`** - Authentication-specific encryption wrappers
**4. `test/library/featuregates.go`** - Feature gate management utilities
**5. `test/library/imagestream.go`** - **NEW**: `ImportImageToImageStream()` function to import external images into cluster internal registry (fixes known-image-checker failures - see below)
**6. `test/library/client.go`** - Added `TestNamespaceBuilder` with PSA methods:
- `WithPSaEnforcement(level)` - Sets enforce/audit/warn labels (prevents PSA audit violations)
- `WithRestrictedPSaEnforcement()`, `WithBaselinePSaEnforcement()`, `WithPrivilegedPSaEnforcement()`
**7. Enhanced existing libraries:**
- `test/library/idpdeployment.go` - Added configurable security context, `deployPodWithImageStream()` wrapper
- `test/library/keycloakidp.go` - ImageStream integration, restricted security context, fixed context handling
- `test/library/gitlabidp.go` - ImageStream integration
- `test/library/waits.go` - Enhanced wait utilities
- `test/library/encryption/helpers.go` - Documentation improvements
```

2. Fix PodSecurityViolation Monitor Failures
```
**Problem:** Tests hardcoded privileged security context for all IDPs.
**Root Cause:** Only GitLab requires privileged mode. Keycloak can run with restricted PSA.
**Solution:** Made IDP security context configurable:
- **Keycloak:** Now uses **restricted** mode (runs as non-root)
- **GitLab:** Continues using **privileged** mode (manages PostgreSQL, Redis, nginx, Sidekiq as root)

**Implementation:**
func createContainerSecurityContext(usePrivileged bool) *corev1.SecurityContext {
    if usePrivileged {
        return &corev1.SecurityContext{Privileged: boolptr(true)}
    }
    // Restricted PSA compliant
    return &corev1.SecurityContext{
        AllowPrivilegeEscalation: boolptr(false),
        RunAsNonRoot: boolptr(true),
        Capabilities: &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
        SeccompProfile: &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
    }
}
**Result:** 50% reduction in PodSecurityViolation errors (Keycloak now compliant, GitLab still requires privileged)
```

3. Fix known-image-checker Monitor Failures
```
**Problem:** Tests pulled images directly from external registries:
image not in mirrored registry: quay.io/keycloak/keycloak:25.0
image not in mirrored registry: docker.io/gitlab/gitlab-ce:13.8.4-ce.0

**Root Cause:** known-image-checker monitor only allows pulls from:
1. Internal cluster registry (`image-registry.openshift-image-registry.svc:5000`)
2. Mirrored repository (`fromRepository`)

**Solution:** Import external images into ImageStreams before pod deployment.
**Implementation:**
**Created `test/library/imagestream.go`:**
func ImportImageToImageStream(
    t testing.TB, kubeconfig *rest.Config, namespace string,
    registry string,      // "quay.io", "docker.io"
    name string,          // "keycloak/keycloak"
    version string,       // "25.0"
    imageStreamName string,
) (internalImage string, cleanup func(), error)

**How it works:**
1. Creates ImageStream in namespace
2. Sets `from.kind: DockerImage` pointing to external registry
3. Waits for image import to complete
4. Returns internal registry reference

**Updated deployments:**
// Keycloak: quay.io/keycloak/keycloak:25.0
// GitLab: docker.io/gitlab/gitlab-ce:13.8.4-ce.0
// Both now pull from: image-registry.openshift-image-registry.svc:5000/<namespace>/<image>
**Result:** 100% compliance with known-image-checker monitor
```

4. Fix Pathological Event Monitor Failures
```
**Problem:** Encryption tests triggered API server rollouts causing disruption events, failing pathological event monitor.
**Solution:** Marked encryption test suites with `ClusterStability: Disruptive` in main.go.
**Result:** Pathological event failures eliminated (CI now expects disruption during encryption tests)
```

5. Bug Fixes
```
### Bug 1: Keycloak Token Expiration (`test/e2e/keycloak.go`)
- **Problem:** Test discarded kcClient with 30-minute token, created new client with 5-minute token → 401 errors
- **Fix:** Reuse kcClient from `AddKeycloakIDP()` with extended token lifetime
- **Impact:** Fixes intermittent `TestKeycloakAsOIDCPasswordGrantCheckAndGroupSync` failures

### Bug 2: Dead Code Removal (`test/library/idpdeployment.go`)
- **Problem:** `deployPod()` function (139 lines) was 95% duplicate with zero callers
- **Fix:** Deleted `deployPod()`, kept `deployPodInNamespace()`
- **Impact:** File reduced from 632 to 495 lines (21% reduction)

### Bug 3: Orphaned Comments (`test/library/keycloakidp.go`)
- **Fix:** Removed 7 instances of comments for next function appearing inside previous function
```


ropatil010:
Before fix in CI logs we were hitting the same error which was mentioned in the Reverting PR: https://github.com/openshift/cluster-authentication-operator/pull/857
```
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-agnostic-ipv6/2037594585412145152
```
After fix the CI logs got pass
```
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-agnostic-ipv6/2038524461480677376
```

/assign @liouk @stbenjam 
Discussion: https://redhat-internal.slack.com/archives/C01CQA76KMX/p1774601343401799

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -37,17 +37,21 @@ import (
...
-var kasServicePort int
+var (
+	kasServicePort     int
+	kasServicePortOnce sync.Once
+)
 
-func init() {
-	(&sync.Once{}).Do(func() {
+func getKASServicePort() int {
+	kasServicePortOnce.Do(func() {
```
liouk, line 46: This now means that this will run for the first time at the first controller's sync -- that looks OK to me, I don't believe it'll make any difference or cause any race in practice.

test/library/client.go
```diff
@@ -35,11 +35,10 @@ func NewClientConfigForTest(t testing.TB) *rest.Config {
 	loader := clientcmd.NewDefaultClientConfigLoadingRules()
 	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loader, &clientcmd.ConfigOverrides{ClusterInfo: api.Cluster{InsecureSkipTLSVerify: true}})
 	config, err := clientConfig.ClientConfig()
-	if err == nil {
-		fmt.Printf("Found configuration for host %v.\n", config.Host)
-	}
-
 	require.NoError(t, err)
+	if t != nil {
```
liouk: This check feels out of place; if t is `nil`, then the line above would also panic. Also, if t is nil we can't really fail the test, so all we can do is panic. I'd suggest skipping this  test.

test/library/encryption/helpers.go
```diff
@@ -52,10 +52,10 @@ func NewClientConfigForTest(t testing.TB) *rest.Config {
 	loader := clientcmd.NewDefaultClientConfigLoadingRules()
 	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loader, &clientcmd.ConfigOverrides{ClusterInfo: kubecmdapi.Cluster{InsecureSkipTLSVerify: true}})
 	config, err := clientConfig.ClientConfig()
-	if err == nil {
-		fmt.Printf("Found configuration for host %v.\n", config.Host)
-	}
 	require.NoError(t, err)
+	if t != nil {
```
liouk: This check feels out of place; if t is `nil`, then the line above would also panic. Also, if t is nil we can't really fail the test, so all we can do is panic. I'd suggest skipping this  test.

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -37,17 +37,21 @@ import (
...
 		var err error
 		kasServicePort, err = strconv.Atoi(os.Getenv("KUBERNETES_SERVICE_PORT_HTTPS"))
 		if err != nil {
-			klog.Warningf("Defaulting KUBERNETES_SERVICE_PORT_HTTPS to 443 due to parsing error: %v", err)
+			klog.V(4).Infof("Defaulting KUBERNETES_SERVICE_PORT_HTTPS to 443 due to parsing error: %v", err)
```
liouk: Why do we need to downgrade this to an Info instead of Warning?

liouk (changes requested):

liouk:
@ropatil010 please correct me if I'm wrong, but this issue seems that it's caused by OTE itself, is there a plan on fixing that on the OTE side?

cmd/cluster-authentication-operator-tests-ext/main.go
```diff
@@ -4,6 +4,10 @@ import (
 	"fmt"
 	"os"
 
+	// Import init package FIRST to configure klog before ANY other non-stdlib imports.
+	// This prevents klog warnings from corrupting JSON output during test listing.
+	_ "github.com/openshift/cluster-authentication-operator/pkg/test/init"
```
liouk, line 9: The approach to force a specific import order can be fragile. As far as I understand, the fix in `pkg/controllers/readiness/wellknown_ready_controller.go` is enough to address the problem. Do we really need `pkg/test/init/init.go` and `klog.SetOutput(os.Stderr)` call in `main()`? To me it seems like we don't; the controller's init fix should be sufficient.

liouk (changes requested):

cmd/cluster-authentication-operator-tests-ext/main.go
ropatil010: Hi @liouk  
 
  If we remove pkg/test/init/init.go:                                                                                                                                                                           
  - Dependencies' init() functions run before main()                                                                                                                                                              
  - They log warnings to stdout during initialization                                                                                                                                                               
  - JSON output gets corrupted BEFORE you can fix it in main()
  - Result: "invalid character 'W'" error still happens 

If we remove klog.SetOutput(os.Stderr) from main():                                                                                                                                                           
  - klog stays suppressed at io.Discard forever                                                                                                                                                                     
  - No logging during actual test execution                                                                                                                                                                         
  - You lose all debug/error output

Both components are necessary:                                                                                                                                                                                  
  1. init package: Suppresses klog during Go's init phase (runs BEFORE main)                                                                                                                                    
  2. main() call: Restores klog to stderr for normal operation

There is NO klog fix in that controller! it just uses klog normally (lines 37, 50, 278). That controller is part of the operator binary, not the test binary where the JSON corruption happens

ropatil010:
Waiting for PR: https://github.com/openshift/origin/pull/31009 to gets merged which fixes the issue "error: encountered errors while listing tests: invalid character 'W' looking for beginning of value". 

The above mentioned issue has been fixed in this PR using init things but looks like it is not acceptable.

ropatil010:
Failure logs are not wrt the PR. Here the tc execution list 
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-serial-ote/2049005575357861888/artifacts/e2e-aws-operator-serial-ote/openshift-e2e-test/artifacts/e2e.log
         
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-parallel-ote/2049005573680140288/artifacts/e2e-aws-operator-parallel-ote/openshift-e2e-test/artifacts/e2e.log
         
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-encryption-serial-ote-2of2/2049005571230666752/artifacts/e2e-aws-operator-encryption-serial-ote/openshift-e2e-test/artifacts/e2e.log
         
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-encryption-perf-serial-ote-1of2/2049005556902924288/artifacts/e2e-aws-operator-encryption-perf-serial-ote/openshift-e2e-test/artifacts/e2e.log
```
PASS logs for profile: e2e-agnostic-ipv6
```
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-agnostic-ipv6/2049008063356080128
```

The dependent PR is merged: https://github.com/openshift/origin/pull/31009

/verified by CI results
@liouk Can you PTAL on this PR which is waiting from long time. Rebased this PR more than 3 times.

test/library/encryption/scenarios.go
```diff
@@ -0,0 +1,371 @@
...
+	return ts
+}
+
+// LocalClientSet represents the client set for local encryption tests.
+// This matches the structure of library.ClientSet but is defined locally.
+type LocalClientSet struct {
+	Etcd            library.EtcdClient
+	ApiServerConfig configv1client.APIServerInterface
+	Kube            kubernetes.Interface
+}
```
liouk: Why do we need this duplicate type? Can we not use `library.ClientSet` everywhere instead and drop the wrappers?

liouk (changes requested):

liouk:
@coderabbitai help

liouk:
@ropatil010 could you please also address the following [comments](https://github.com/openshift/cluster-authentication-operator/pull/859#pullrequestreview-4186244148) from @coderabbitai:
  - `test/library/encryption/perf_helpers.go:119` — "All workers completed successfully" logged even when errors occurred in `r.errorStore`
  - `test/e2e-encryption-rotation/encryption_rotation.go:76` and `test/e2e-encryption/encryption.go:72` — use `configv1.EncryptionType("aescbc")` instead of `configv1.EncryptionTypeAESCBC`
  - `test/e2e-oidc/external_oidc.go:950-969` — Keycloak group+user created but never cleaned up
  - `test/e2e-oidc/external_oidc.go:390-439` — `fetchIssuerJWKS` doesn't accept context for cancellation

ropatil010:
@coderabbitai help

ropatil010:
Hi @liouk PTAL on the PR. 
FI Could be: Even if we take coderabbitai suggestion, we may fix as per the suggestion and push, next time if we ask again it will give newer suggestions. Do let me know your suggestions. 

Failures profiles are wrt Monitor cases 
e2e-aws-operator-serial-ote
e2e-aws-operator-parallel-ote
e2e-aws-operator-encryption-serial-ote-2of2

Currently added for these jobs:
e2e-aws-operator-encryption-serial-ote
e2e-aws-operator-encryption-perf-serial-ote
e2e-aws-operator-encryption-rotation-serial-ote

Wrt OIDC cases PASS in e2e-aws-operator-serial-ote
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-serial-ote/2051487650946224128/artifacts/e2e-aws-operator-serial-ote/openshift-e2e-test/artifacts/e2e.log

ropatil010:
Hi @liouk 
Currently e2e-gcp-operator-disruptive Profile skipping execution of OIDC case here to merge this PR. 

If we include OIDC case to execute it is taking 2.5hours to complete 
```
**[sig-auth] authentication operator [OIDC][Serial][Disruptive] TestExternalOIDCWithKeycloak [Timeout:3h]**
```
The above OIDC tc might be causing other cases to fail ex: 
```
 [Monitor:audit-log-analyzer] [bz-apiserver-auth][invariant] audit analysis PodSecurityViolation
 [Monitor:audit-log-analyzer][Jira:"kube-apiserver"] API resources are not updated excessively
 [Monitor:known-image-checker][sig-arch] Only known images used by tests
 [Monitor:legacy-node-invariants][sig-node] kubelet should be able to delete cgroups path
```
Root cause, Need to analyse for other cases. Will try to open New PR to fix them/analyse the failures
```
PodSecurityViolation - [Monitor:audit-log-analyzer] audit analysis PodSecurityViolation
  - Cause: Keycloak pod (keycloak-5f4cd6c46f) created by OIDC test violates PodSecurity policy
  - Details: Container runs as privileged, doesn't drop capabilities, doesn't set runAsNonRoot=true
  - NOT SCC-related: This is Pod Security Standards violation
```

liouk:
As discussed offline with @ropatil010, we will tackle the failing OIDC test at a separate PR. The issue should be fixed easily by removing the privileged context from the keycloak pod -- this is currently unnecessary, as seen in the respective [origin test deployment](https://github.com/openshift/origin/blob/main/test/extended/authentication/keycloak_helpers.go) as well.

/lgtm

ropatil010:
Below Failure profiles are related to monitor cases.
1. e2e-aws-operator-encryption-perf-serial-ote-2of2
```
0 pass, 0 flaky, 0 skip (7m41s)
```
2. e2e-aws-operator-encryption-serial-ote-2of2
```
passed: (30m34s) 2026-06-10T15:20:57 "[sig-auth] authentication operator [Encryption][Serial] TestEncryptionTurnOnAndOff [Timeout:3h]"
passed: (5m1s) 2026-06-10T15:25:58 "[sig-auth] authentication operator [Encryption][Serial] TestEncryptionTypeIdentity"
```
3. e2e-aws-operator-serial-ote
```
passed: (2m57s) 2026-06-10T14:41:03 "[sig-auth] authentication operator [OIDC][Serial] TestKeycloakAsOIDCPasswordGrantCheckAndGroupSync"
passed: (3m50s) 2026-06-10T14:44:52 "[sig-auth] authentication operator [OIDC][Serial] TestGitLabAsOIDCPasswordGrantCheck"
```
4. e2e-gcp-operator-disruptive
```
passed: (45.5s) 2026-06-10T13:45:09 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should enforce cross-namespace ingress traffic"
passed: (13.7s) 2026-06-10T13:45:23 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should enforce NetworkPolicy allow/deny basics in a test namespace"
passed: (13.3s) 2026-06-10T13:45:37 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should enforce authentication-operator NetworkPolicies"
passed: (1m29s) 2026-06-10T13:47:06 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should block unauthorized namespace traffic"
passed: (15m45s) 2026-06-10T14:02:52 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should restore auth NetworkPolicies after delete or mutation[Timeout:30m]"
passed: (1m16s) 2026-06-10T14:04:08 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should enforce oauth-apiserver NetworkPolicies"
passed: (10.1s) 2026-06-10T14:04:19 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should ensure auth NetworkPolicies are defined"
passed: (13.3s) 2026-06-10T14:04:33 "[sig-auth] authentication operator [NetworkPolicy][Disruptive] should enforce auth NetworkPolicies"
passed: (2h27m16s) 2026-06-10T16:31:50 "[sig-auth] authentication operator [OIDC][Serial][Disruptive] TestExternalOIDCWithKeycloak [Timeout:3h]"
```

Hi @liouk PTAL on this PR. Any suggestion would you like to give? 

/test e2e-aws-operator-encryption-perf-serial-ote-2of2

test/library/idpdeployment.go
```diff
@@ -33,6 +33,38 @@ func boolptr(b bool) *bool {
...
+			Drop: []corev1.Capability{"ALL"},
+		},
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: corev1.SeccompProfileTypeRuntimeDefault,
+		},
+	}
+}
+
+// SecurityContextConfig defines the security context requirements for an IDP deployment.
+type SecurityContextConfig struct {
```
liouk: This is needlessly complicated, I don't see the need for a type with a single field and a constructor. As far as I can see a boolean will suffice.

liouk (changes requested):
As far as I can see, the direction of the most recent changes isn't the right one:

1. Adding `[Disruptive]` to encryption tests breaks the dedicated encryption OTE suites. The encryption suite qualifiers in `main.go` filter on `[Encryption]` and `[Serial]` but explicitly exclude `[Disruptive]` (via `!name.contains("[Disruptive]")`). By adding `[Disruptive]` to these tests, they'll be
  excluded from their own suites and instead fall into the operator/disruptive suite. The encryption-specific OTE jobs (`operator-encryption/serial`, `operator-encryption-rotation/serial`,
  `operator-encryption-perf/serial`) will have zero tests to run. I don't think we need to add `[Disruptive]` to the tests, we should instead set cluster stability to disruptive; these are two different things (one is a tag used for test filtering, the other is a suite property that affects what monitor tests are executed on the suite).

2. The `e2e-gcp-operator-disruptive` job fails on `[Monitor:known-image-checker]` Only known images used by tests -- the Keycloak/GitLab images aren't in the allowed list. Making the security context restricted
  doesn't fix this. Even with this commit, the OIDC test still runs in the disruptive suite and the image check will still fail. Setting cluster stability to disruptive for the OIDC tests is the solution here as well, as it will not check for known images, but at the same time the OIDC suite is indeed disruptive as it currently rolls out the kube-apiserver.

3. This change contains excessive comments, where in most cases we could do with 1-2 lines. The code is self explanatory and in most cases the comments don't add any information. Additionally, there seem to be a few misplaced comments in `keycloakidp.go`.

ropatil010:
Hi @liouk PTAL on the PR.  Execution results adding below
```
suite: e2e-aws-operator-encryption-perf-serial-ote-1of2, folder: test/e2e-encryption-perf/
passed: (14m19s) 2026-06-16T17:18:31 "[sig-auth] authentication operator [Encryption][Serial] TestPerfEncryptionTypeAESCBC"

suite: e2e-aws-operator-encryption-rotation-serial-ote, folder: test/e2e-encryption-rotation/
passed: (15m24s) 2026-06-16T17:22:27 "[sig-auth] authentication operator [Encryption][Serial] TestEncryptionRotation [Timeout:3h]"

suite: e2e-aws-operator-encryption-serial-ote, folder: test/e2e-encryption/
passed: (5m0s) 2026-06-16T17:10:51 "[sig-auth] authentication operator [Encryption][Serial] TestEncryptionTypeUnset"

suite: e2e-aws-operator-encryption-serial-ote, folder: test/e2e-encryption/
passed: (30m32s) 2026-06-16T17:34:59 "[sig-auth] authentication operator [Encryption][Serial] TestEncryptionTurnOnAndOff [Timeout:3h]"
passed: (5m0s) 2026-06-16T17:40:00 "[sig-auth] authentication operator [Encryption][Serial] TestEncryptionTypeIdentity"

suite: e2e-aws-operator-parallel-ote, folder: test/e2e/
passed: (6.6s) 2026-06-16T16:56:23 "[sig-auth] authentication operator [Operator][Certs] TestRouterCerts"
passed: (32.4s) 2026-06-16T16:56:56 "[sig-auth] authentication operator [Operator][Routes] TestCustomRouterCerts"
passed: (3m10s) 2026-06-16T16:59:27 "[sig-auth] authentication operator [Templates] TestTemplatesConfig"
passed: (27m11s) 2026-06-16T17:23:28 "[sig-auth] authentication operator [Tokens] TestTokenInactivityTimeout [Timeout:1h]"

suite: e2e-aws-operator-serial-ote, folder: test/e2e/
passed: (4m40s) 2026-06-16T17:01:00 "[sig-auth] authentication operator [OIDC][Serial] TestKeycloakAsOIDCPasswordGrantCheckAndGroupSync"
passed: (4m51s) 2026-06-16T17:05:51 "[sig-auth] authentication operator [OIDC][Serial] TestGitLabAsOIDCPasswordGrantCheck"

e2e-gcp-operator-disruptive
passed: (2h17m28s) 2026-06-16T19:07:46 "[sig-auth] authentication operator [OIDC][Serial][Disruptive] TestExternalOIDCWithKeycloak [Timeout:3h]"
```

test/library/imagestream.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to convert ImageStream to unstructured: %w", err)
+	}
+
+	// Create the ImageStream
+	_, err = dynamicClient.Resource(imageStreamGVR).Namespace(namespace).Create(
+		context.TODO(), unstructuredIS, metav1.CreateOptions{})
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to create imagestream: %w", err)
+	}
```
liouk: You have an ImageStream object already, why do you need to convert to unstructured and then use a dynamic client to create the resource? We are already importing `github.com/openshift/api` so we can use directly a client from `imagev1client.NewForConfig()` and then call `imageclientv1.ImageV1Client()` to create the ImageStream.

test/library/imagestream.go
```diff
@@ -0,0 +1,166 @@
...
+	})
+
+	if err != nil {
+		cleanup()
+		return "", nil, fmt.Errorf("timeout waiting for image import: %w", err)
+	}
+
+	// Return the internal registry reference
+	// Format: image-registry.openshift-image-registry.svc:5000/<namespace>/<imagestream>:<tag>
+	internalImage := fmt.Sprintf("image-registry.openshift-image-registry.svc:5000/%s/%s:%s",
```
liouk: AFAICT this is the same with what we're using above in line 132-133: `tag.Items[0].DockerImageReference` -- why not use that and construct the URL again?

test/library/imagestream.go
```diff
@@ -0,0 +1,166 @@
...
+//   - name: image name (e.g., "keycloak/keycloak", "gitlab/gitlab-ce")
+//   - version: image tag/version (e.g., "25.0", "13.8.4-ce.0")
+//   - imageStreamName: the name for the ImageStream (e.g., "keycloak", "gitlab")
+//
+// Returns:
+//   - internalImage: the internal registry reference that pods should use
+//   - cleanup: function to delete the ImageStream
+//   - error: any error that occurred
+func ImportImageToImageStream(
+	t testing.TB,
```
liouk, line 34: `context.TODO()` and `context.Background()` are both used inconsistently around this func -- we should instead derive a context from `testing.TB`, i.e. `t.Context()` and use that across the whole func instead.

test/library/imagestream.go
```diff
@@ -0,0 +1,166 @@
...
+			if tag.Tag == version && len(tag.Items) > 0 && tag.Items[0].DockerImageReference != "" {
+				t.Logf("Image successfully imported: %s", tag.Items[0].DockerImageReference)
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+
+	if err != nil {
+		cleanup()
```
liouk: We're calling the cleanup func here but also returning it up the caller chain; this should only be a responsibility of the caller, so returning it should suffice. The caller must call cleanup when/as necessary.

test/library/featuregates.go
liouk, line 1: The code in this file is never used and should be deleted.

liouk (changes requested):

test/library/imagestream.go
ropatil010: Hi @liouk 
After fixing as per your suggestions the cases got failed with below reasons. 
Since Ginkgo doesn't support t.Context(), we should use context.Background() or context.TODO() as we had before. I'm reverting back the code to my original code
Here is the logs: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/859/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-serial-ote/2067288905291403264

Reason: 
The GinkgoTBWrapper has an embedded testing.TB interface field that is NIL. When you call t.Context() on a GinkgoTBWrapper:
  1. Go looks for Context() method on GinkgoTBWrapper - not found
  2. Go looks at the embedded testing.TB field - it's an interface that is NIL
  3. Trying to call a method on a nil interface → nil pointer dereference panic!

ropatil010:
closing the PR, there could be better solution as this is time consuming.
After CI profiles passing or fixing some suggestions which is breaking other cases to fail always(Going in inifite loop)
/close

PR #880, opened 2026-05-01
CNTRLPLANE-3210: Update external OIDC config generation to support external claims sourcing

everettraven:
## Release Notes

* **New Features**
  * Enabled External OIDC claims sourcing via feature gate
  * Added support for multiple authentication methods (request-provided token, anonymous, client credentials)
  * TLS certificate authority and claim mapping configuration support

* **Tests**
  * Added comprehensive test coverage for External OIDC claims sourcing feature with various authentication scenarios


everettraven:
@coderabbit full review

ShazaAldawamneh:
The dependency bump properly brings in the external OIDC claims sourcing API types with good validation rules and feature gate protection. 
/lgtm

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
```diff
@@ -752,3 +762,296 @@ func isConstField(exp *exprpb.Expr, field string) bool {
...
+	return out, nil
+}
+
+func generateExternalClaimsSource(source configv1.ExternalClaimsSource, cmLister corev1listers.ConfigMapLister, secretLister corev1listers.SecretLister, seenClaimNames sets.Set[string]) (*authenticationv1alpha1.ExternalClaimsSource, error) {
+	authentication, err := generateExternalClaimsSourceAuthentication(source.Authentication, secretLister, cmLister)
+	if err != nil {
+		return nil, err
+	}
+
+	tls, err := generateExternalClaimsSourceTLS(source.TLS, cmLister)
```
liouk: AFAICS, if we omit TLS (in order to use system default CAs), generation will fail because it'll unconditionally try to retrieve a configmap; we need to handle the case where TLS is unset.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
```diff
@@ -752,3 +762,296 @@ func isConstField(exp *exprpb.Expr, field string) bool {
 	c := exp.GetConstExpr()
 	return c != nil && c.GetStringValue() == field
 }
+
+func generateExternalClaimsSources(cmLister corev1listers.ConfigMapLister, secretLister corev1listers.SecretLister, sources ...configv1.ExternalClaimsSource) ([]authenticationv1alpha1.ExternalClaimsSource, error) {
+	out := []authenticationv1alpha1.ExternalClaimsSource{}
+	seenClaimNames := sets.New[string]()
+	for _, source := range sources {
+		source, err := generateExternalClaimsSource(source, cmLister, secretLister, seenClaimNames)
```
liouk: nit: This works as is, but we're shadowing the `source` variable, so this might be confusing to the reader.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
```diff
@@ -752,3 +762,296 @@ func isConstField(exp *exprpb.Expr, field string) bool {
...
+			return nil, fmt.Errorf("validating client secret: %w", kubeErrorListToGoError(err))
+		}
+	*/
+
+	scopes, err := generateClientCredentialScopes(clientCredentialConfig.Scopes...)
+	if err != nil {
+		return nil, fmt.Errorf("generating scopes: %w", err)
+	}
+
+	var certificateAuthority *string = nil
```
liouk: nit: `= nil` is unnecessary.


```suggestion
	var certificateAuthority *string
```

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
```diff
@@ -752,3 +762,296 @@ func isConstField(exp *exprpb.Expr, field string) bool {
...
+	}
+
+	return out, errors.Join(errs...)
+}
+
+func generateExternalClaimsSourceConditions(externalSourcePredicates ...configv1.ExternalSourcePredicate) ([]authenticationv1alpha1.ExternalSourceCondition, error) {
+	out := make([]authenticationv1alpha1.ExternalSourceCondition, 0, len(externalSourcePredicates))
+
+	errs := []error{}
+	// seenConditions := sets.New[string]()
```
liouk, line 1022: How is this intended to be used? I assume that `ValidateExternalSourceCondition()` will set it, but then what is its purpose? I don't see it used anywhere else in the (commented out) code.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
liouk: On a similar topic, we don't seem to have any tests where `ExternalSourceTLS.CertificateAuthority.Name` is empty (which would mean "use system CAs").

liouk (changes requested):
Looks great overall; my main concern is the TLS CA handling inconsistency flagged below.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
everettraven: Ah, good catch - I don't know why I did that 🤦

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
everettraven: It is intended to be used to ensure that the external source conditions are unique. It is used by the commented out call to `ValidateExternalSourceCondition()` on line 1028.

The source for that validation function we will eventually call is at https://github.com/openshift/oauth-apiserver/blob/97a820bd5412c2d3bb22f083152107ef41631c8c/pkg/externaloidc/apis/authentication/validation/validation.go#L178-L180 and it uses a `sets.Set` of the already seen condition expressions to prevent duplicate expressions being specified for whether or not we should source claims from an external source.

everettraven: We also enforce this at admission at the API level via https://github.com/openshift/api/blob/261e3a0546f3668dad89a722175e0fa96b505784/config/v1/types_authentication.go#L917

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
everettraven: Fixed in the latest push.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
everettraven: Fixed both of these in the latest push.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
everettraven: Fixed in the latest push.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
liouk: Right, I understand why `ValidateExternalSourceCondition()` needs it, but I just wanted to clarify whether `generateExternalClaimsSourceConditions()` will use it in some other way additionally -- all clear :slightly_smiling_face:

PR #893, opened 2026-05-12
CNTRLPLANE-3210: Refactor `externaloidc` controller to support different configuration generation logic

everettraven:
## Description

This PR refactors the existing `externaloidc` controller to better facilitate creating unique generation logic depending on feature gate state. As part of this:
- Validation of generated configuration has been moved to be the responsibility of a generator
- All existing validation and generation logic has been moved to deeper nested packages:
    - `/generation/kube-apiserver` for kube-apiserver specific generation/validation logic
    - `/generation/oauth-apiserver` for oauth-apiserver specific generation/validation logic
- Tests have been adjusted and adapted

> [!NOTE]
> The key difference between the `generation/kube-apiserver` and `generation/oauth-apiserver` packages are the types they use for conversion and config generation.

This PR sets the foundation for future changes to the oauth-apiserver specific configuration generation process as the external claims sourcing feature continues to progress.

This PR was split out from https://github.com/openshift/cluster-authentication-operator/pull/880 which adds new oauth-apiserver configuration generation logic - in case reviewers would like to see how this will be updated in the future.

## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift dependencies and added oauth-apiserver entry.

* **Refactor**
  * External OIDC controller now delegates auth config generation to separate kube-apiserver and oauth-apiserver generators, simplifying controller logic.

* **New**
  * Added dedicated generators for producing apiserver AuthenticationConfiguration for kube-apiserver and oauth-apiserver.

* **Tests**
  * Expanded and modernized test suites covering generation, validation, CA/discovery checks, and many mapping/validation scenarios.

* **Documentation**
  * Added package-level documentation for the kube-apiserver generator.


pkg/controllers/externaloidc/generation/common.go
```diff
@@ -0,0 +1,235 @@
...
+		return false
+	}
+	return hasSelectExp(ast.Expr(), "claims", "email")
+}
+
+func usesEmailVerifiedClaim(ast *celgo.Ast) bool {
+	if ast == nil {
+		return false
+	}
+	return hasSelectExp(ast.Expr(), "claims", "email_verified")
```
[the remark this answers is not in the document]
everettraven: I'm going to reject this suggestion for now. This was behavior copied from upstream Kubernetes previously and is still how Kubernetes does it today.

Because this is primarily a refactor of existing behavior, and this looks to be recently deprecated, I'm going to consider this fix out of scope for this PR.

pkg/controllers/externaloidc/generation/kubeapiserver/doc.go
```diff
@@ -0,0 +1,9 @@
+// kubeapiserver serves the purpose of generating the
+// AuthenticationConfiguration types for configuring the
+// Kubernetes API Server with a direct OIDC provider token authenticator.
+//
+// TODO: Remove this package once the ExternalOIDCExternalClaimsSourcing feature gate
```
liouk, line 5: Would it make sense to accompany this TODO with a jira ticket to track it? Plain TODO comments scare me 😅

pkg/controllers/externaloidc/externaloidc_controller.go
liouk, line 55: nit: we don't really need to pre-declare the variable, how about direct assignment?

pkg/controllers/externaloidc/generation/kubeapiserver/generate.go
```diff
@@ -0,0 +1,487 @@
...
+)
+
+type AuthenticationConfigurationValidator interface {
+	ValidateAuthenticationConfiguration(*apiserverv1beta1.AuthenticationConfiguration) error
+}
+
+type AuthenticationConfigurationGenerator struct {
+	configMapLister corev1listers.ConfigMapLister
+	featureGates    featuregates.FeatureGate
+	validator       AuthenticationConfigurationValidator
```
liouk: We don't seem to be using this field outside of tests; probably an oversight?

liouk: This field is never set or used in actual code apart from tests; likely an oversight? Especially for the kubeapiserver package, this introduces a regression because validations won't run AFAICS.

pkg/controllers/externaloidc/generation/oauthapiserver/generate.go
```diff
@@ -0,0 +1,490 @@
...
+
+type AuthenticationConfigurationValidator interface {
+	ValidateAuthenticationConfiguration(*authenticationv1alpha1.AuthenticationConfiguration) error
+}
+
+type AuthenticationConfigurationGenerator struct {
+	configMapLister corev1listers.ConfigMapLister
+	secretLister    corev1listers.SecretLister
+	featureGates    featuregates.FeatureGate
+	validator       AuthenticationConfigurationValidator
```
liouk: This field is never set or used in actual code apart from tests; likely an oversight?

pkg/controllers/externaloidc/generation/kubeapiserver/validation/validator.go
```diff
@@ -0,0 +1,46 @@
...
+	"strings"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/controllers/externaloidc/generation"
+	apiserverv1beta1 "k8s.io/apiserver/pkg/apis/apiserver/v1beta1"
+	"k8s.io/client-go/util/cert"
+)
+
+type AuthenticationConfigurationValidator struct{}
+
+func (acv *AuthenticationConfigurationValidator) ValidateAuthenticationConfiguration(auth *apiserverv1beta1.AuthenticationConfiguration) error {
```
liouk: As far as I can see, `ValidateAuthenticationConfiguration()` methods are almost identical in the two generator packages (only difference is really the type of DiscoveryURL, `string` vs `*string`), which makes this whole structure overcomplicated: generator -> validator -> common helpers from generation pkg (which is where most of the validation happens).

Can we structure this in a simpler way? For example, I don't see the need for the validator packages; we each generator could call a private validation method that uses `ValidateCACert()` directly.

pkg/controllers/externaloidc/externaloidc_controller.go
```diff
@@ -2,66 +2,46 @@ package externaloidc
...
 )
 
 const (
-	configNamespace                 = "openshift-config"
-	managedNamespace                = "openshift-config-managed"
-	targetAuthConfigCMName          = "auth-config"
-	authConfigDataKey               = "auth-config.json"
-	oidcDiscoveryEndpointPath       = "/.well-known/openid-configuration"
-	kindAuthenticationConfiguration = "AuthenticationConfiguration"
+	configNamespace        = "openshift-config"
```
liouk, line 28: Looks like we can drop this, as we're now using `generation.ConfigNamespace` everywhere.

liouk (changes requested):
One thing that stands out in this PR is that the two generators massively duplicate code for most of the generation logic (and the respective tests). As a note to self and other reviewers, as far as I understand this will change with #880 and is meant as a placeholder with the current PR. Notably, tests could be de-duplicated further even with #880 in place.

The rest of the comments I've added here seem to be applicable regardless of the changes from #880.

pkg/controllers/externaloidc/generation/kubeapiserver/doc.go
everettraven: Yeah, I'll create a jira ticket and link it here.

pkg/controllers/externaloidc/generation/kubeapiserver/generate.go
everettraven: Ah, good catch -  I meant to default this as part of `NewAuthenticationConfigurationGenerator()`. Will update.

pkg/controllers/externaloidc/generation/kubeapiserver/validation/validator.go
everettraven: The purpose of implementing the distinct validator packages was two-fold:
1. Abstract the validation behavior out into a mockable interface so we can test the generator logic without having to also inherently test the validation behavior that requires standing up a local http server.
2. The separation of the `kube-apiserver` and `oauth-apiserver` validator packages is so that we can evolve the validation logic of the `oauth-apiserver` package over time as needed. #880 currently doesn't add any changes to the validation logic because of the fact that there is a kube version skew between CAO and oauth-apiserver that makes it not yet possible to import the validations the oauth-apiserver runs to be used as our validation step.

I do agree that the current pathing of `generator -> validator -> common` is a bit cumbersome and is probably a bit over-abstracted. I'm happy to simplify this to `generator -> validator` by having each validation package carry a duplicate of the `ValidateCACert()` function that it calls instead of using `common.ValidateCACert()`.

>I don't see the need for the validator packages; we each generator could call a private validation method

I'm not strongly opinionated here, and I'm OK with dropping the validator packages altogether if you feel they aren't providing any value.

My intention with separating it out into a package is so that we could, over time, introduce new feature validation logic as part of a new validator that is configured based on gate state instead of making the validators aware of gate state. That being said, there is a trade off there of likely duplicating existing validation logic.

Reasoning about this a bit further, I've convinced myself that just making it an overridable private field on the generator is probably OK and making it gate aware is probably the path with less churn overall. I'll make the updates to adjust this accordingly.

pkg/controllers/externaloidc/generation/kubeapiserver/doc.go
everettraven: Updated this comment with a link to https://redhat.atlassian.net/browse/CNTRLPLANE-3454

pkg/controllers/externaloidc/externaloidc_controller.go
everettraven: We need to pre-declare the variable so it is of the local interface type rather than the package-specific generator type

pkg/controllers/externaloidc/externaloidc_controller.go
everettraven: I removed the exported `generation.ConfigNamespace` in favor of everything just explicitly declaring the namespaces it cares about.

This constant is used to influence the namespace that the ConfigMap informers this controller uses is created against.

This follows the existing pattern of all the packages defining their own constants. I would prefer that we move to a shared constants model where possible but that would be a significantly larger change that I'd consider out of scope for this work.

everettraven:
> One thing that stands out in this PR is that the two generators massively duplicate code for most of the generation logic (and the respective tests). As a note to self and other reviewers, as far as I understand this will change with #880 and is meant as a placeholder with the current PR. Notably, tests could be de-duplicated further even with #880 in place.
> 
> The rest of the comments I've added here seem to be applicable regardless of the changes from #880.

@liouk Thanks for the review! I made some pretty significant changes in response to your comments. Notably, there is no longer a separate validation package for each generator and there is no more `common.go` implementation that tries to bridge the gap between code re-use.

Instead, all the generation logic is mostly the exact same between the `generation/kubeapiserver` and `generation/oauthapiserver` packages with the key difference being the return types from the generation.

I agree that, in theory, we could wire things up such that there is less duplication in place even with the staged changes in #880 . In practice, I don't think the juice is worth the squeeze there. We already have it planned to remove the `generation/kubeapiserver` package in its entirety once the new external claims feature has been promoted and configuring the oauth-apiserver is the only configuration path we have to support. The inclusion of the `generation/kubeapiserver` package is a stop-gap for backwards compatibility until we have promoted the feature and it has shipped as GA with an official release.

If you feel strongly that we should limit duplication, I'm happy to have continued discussions here.

pkg/controllers/externaloidc/generation/kubeapiserver/validation/validator.go
everettraven: In the latest push, I've collapsed this to be such that everything is self-contained for each generator.

This increases code duplication, but should simplify the logical overhead of how a config is produced and validated.

See https://github.com/openshift/cluster-authentication-operator/pull/893#issuecomment-4442930551 for my thoughts on the code duplication aspect.

pkg/controllers/externaloidc/generation/kubeapiserver/generate.go
```diff
@@ -0,0 +1,750 @@
...
+	var resp *http.Response
+	var connErr error
+	retryCtx, cancel := context.WithTimeout(req.Context(), 10*time.Second)
+	defer cancel()
+	if err := retry.RetryOnConnectionErrors(retryCtx, func(ctx context.Context) (done bool, err error) {
+		resp, connErr = client.Do(req.WithContext(ctx))
+		return connErr == nil, connErr
+	}); err != nil {
+		return fmt.Errorf("persistent well-known GET error: %v", err)
+	}
```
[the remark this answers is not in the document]
everettraven: Good catch! I had intended for this to use `http.DefaultTransport` but got mixed up - thanks! I just pushed a fix.

liouk:
@everettraven I much prefer the approach with `validationFunc`, thanks for the change :+1: This way we can still preserve testability while keeping things simple.

Regarding duplication, that's fine with me, I'm aware of the reasons -- I know we're dropping the one package soon, but still wanted to raise this point for awareness and reference. I'm happy with how things stand now; we'll just need to keep this duplication in mind and fix any potentially arising issues in both places.

/lgtm

PR #901, opened 2026-05-27
OCPBUGS-86221: Update the well-known readiness controller to properly set cluster-operator conditions

everettraven:
As part of investigating [OCPBUGS-86221](https://redhat.atlassian.net/browse/OCPBUGS-86221) it was found that the likely culprit was the well-known readiness controller setting `Progressing=True` when a kube-apiserver endpoint was not yet accessible during a node reboot being done by the MCO.

This PR attempts to resolve this by aligning the status conditions that are set by this controller to better align with the status condition type meanings as outlined in https://github.com/openshift/api/blob/09730f85d8835712c0412a585685ddadb01e171b/config/v1/types_cluster_operator.go#L151.

To facilitate this, the following changes were made:
- `Available` gets set to `True` when _any_ of the kube-apiservers returns a valid response for the well-known endpoint.
- `Progressing` is replaced by `DegradationObserved` when _any_ of the kube-apiservers is not sufficiently serving their well-known endpoint. This controller has no operands, making `Progressing` a condition it should not be setting.
- `Degraded` is set after 5 minutes of a given `DegradationObserved` state continuing to be observed.

To ensure the behavior is working as expected, additional unit tests were added to exercise this new behavior.

## Summary by CodeRabbit

* **New Features**
  * Introduced time-based controller degradation tracking and reporting for clearer health visibility
  * Refined well-known endpoint readiness checks with improved aggregation and error classification

* **Bug Fixes / Behavior Changes**
  * Replaced previous "progressing" condition with a "degradation observed" condition for readiness reporting

* **Tests**
  * Added comprehensive unit and end-to-end tests covering endpoint scenarios and degradation timeouts


pkg/controllers/readiness/wellknown_ready_controller.go
liouk: nit: Would it be worth replacing `ioutil` with `io` since the former is deprecated, since we're touching this file?

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -165,30 +174,36 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
...
+				WithReason("AtLeastOneWellKnownEndpointAvailable")
+		} else {
+			available = available.
+				WithStatus(operatorv1.ConditionFalse).
+				WithReason("NotReady").
+				WithMessage(fmt.Sprintf("No well-known endpoints are available: %s", err.Error()))
+		}
+		degradationObserved = degradationObserved.
 			WithStatus(operatorv1.ConditionTrue)
-		if progressingErr, ok := err.(*common.ControllerProgressingError); ok {
```
liouk, line 180: nit: `common.ControllerProgressingError` doesn't look like it's used anymore; maybe worth cleaning up?

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -128,8 +137,8 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
 	// ensuring the API server will reject the Apply operation if the status is not set down below
 	available := applyoperatorv1.OperatorCondition().
 		WithType("WellKnownAvailable")
-	progressing := applyoperatorv1.OperatorCondition().
-		WithType(common.ControllerProgressingConditionName(controllerName))
```
liouk, line 132: nit: `common.ControllerProgressingConditionName` doesn't look like it's used anymore; maybe worth cleaning up?

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -197,56 +212,69 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
...
+	// so therefore we are _technically_ available but have observed a degraded state.
+	if len(errs) > 0 && len(errs) < len(ips) {
+		return true, NewControllerDegradationObservedError("AtLeastOneWellKnownEndpointUnavailable", errors.Join(errs...), 5*time.Minute)
+	}
+
+	if len(errs) > 0 {
+		return false, errors.Join(errs...)
+	}
+
+	// TODO(everettraven): verify that this is no longer necessary.
```
liouk, line 258: Are you planning in verifying this in the scope of this PR, or at a later stage? To me it also _looks_ like this is no longer necessary; if I understand the `alreadyTrueOnce` guard correctly, we won't need it any more because we are now going available when at least one endpoint is healthy, and if some are still rolling out, we're capturing that with the new partial-failure model.

If you're planning on revisiting this after this PR merges, maybe it'd be better to drop the code already and open an issue to track the verification.

everettraven: I'd like to verify this within the scope of this PR by running some payload jobs to verify that this change won't trigger any regressions in existing tests (once I tackle additional work to take care of existing CI failures here)

pkg/controllers/readiness/wellknown_ready_controller.go
everettraven: Ah, good catch. I didn't even check with it being in the `common` package - just assumed something else would be using it.

pkg/controllers/readiness/wellknown_ready_controller.go
everettraven: Done in the latest commit

pkg/controllers/readiness/wellknown_ready_controller.go
everettraven: Done in the latest push.

pkg/controllers/readiness/wellknown_ready_controller.go
everettraven: Done in the latest push.

pkg/controllers/readiness/error.go
```diff
@@ -0,0 +1,67 @@
...
+// longer than `maxAge` ago
+func (cdoe *ControllerDegradationObservedError) IsDegraded(controllerName string, lastStatus *operatorv1.OperatorStatus) bool {
+	if cdoe.maxAge <= 0 {
+		return false
+	}
+
+	lastCondition := v1helpers.FindOperatorCondition(lastStatus.Conditions, ControllerDegradationObservedConditionName(controllerName))
+	if lastCondition == nil {
+		return false
+	}
```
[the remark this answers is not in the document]
everettraven: Done in the latest push

pkg/controllers/readiness/wellknown_ready_controller.go
```diff
@@ -165,30 +174,36 @@ func (c *wellKnownReadyController) sync(ctx context.Context, controllerContext f
...
+		if degradationErr, ok := err.(*ControllerDegradationObservedError); ok {
+			if degradationErr.IsDegraded(controllerName, operatorStatus) {
+				return degradationErr.Unwrap()
 			}
-			progressing = progressingErr.ToCondition(controllerName)
+			degradationObserved = degradationErr.ToCondition(controllerName)
 			return nil
 		} else {
 			return err
 		}
```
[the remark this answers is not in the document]
everettraven: I've updated the implementation such that there is only a single case where the `ControllerDegradationObservedError` is used to make it more clear that the intention is that any errors outside of that case result in an explicit and immediate setting of the `Degraded` condition.

everettraven:
/testwith openshift/cluster-authentication-operator/e2e-agnostic https://github.com/openshift/origin/pull/31236

everettraven:
/testwith openshift/cluster-authentication-operator/master/e2e-agnostic https://github.com/openshift/origin/pull/31236

liouk:
LGTM :+1: Holding until [this TODO](https://github.com/openshift/cluster-authentication-operator/pull/901#discussion_r3326028431) is completed.

/lgtm
/hold

everettraven:
/payload-job periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-aws-ovn-upgrade

everettraven:
Payload jobs look alright. Initial run had some timeouts and the individual job had a failure unrelated to the CAO.

Removing the hold and marking verified.

/hold cancel
/verified by ci

PR #905, opened 2026-06-01
NO-JIRA: kms rotation controller

tjungblu:
## Summary by CodeRabbit

* **Chores**
  * Updated module resolution to redirect two third-party dependencies to alternative sources, ensuring compatibility with testing and build workflows.
  * These redirects change where the project fetches specific dependency versions, stabilizing builds and aiding integration testing without altering runtime behavior.


vendor/github.com/openshift/library-go/pkg/operator/encryption/controllers/encryption_rotation_controller.go
```diff
@@ -0,0 +1,293 @@
...
+func (c *encryptionRotationController) reconcileKMSPlugin(
+	ctx context.Context,
+	keyID string,
+	encryptedGRs []schema.GroupResource,
+	writeKeySecret *corev1.Secret,
+	healthReports []encryptionstatus.KMSPluginHealthReport,
+	rotations []encryptionstatus.KMSPluginRotationStatus,
+) ([]encryptionstatus.KMSPluginRotationStatus, error) {
+	// Check KEK convergence across all nodes for this keyID.
+	convergedKEKID, converged := encryptionstatus.ConvergedKEKForKeyID(healthReports, keyID)
```
ardaguclu: I know that this is draft and WIP. Just dropped a comment if it makes sense;

In order to adopt the invariants of other controllers, if there is active migration **or** one of the kms plugins are not healthy, we should return just like `!converged`.

vendor/github.com/openshift/library-go/pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -78,6 +78,25 @@ func (c *Config) HasEncryptionConfiguration() bool {
 	return c != nil && c.Encryption != nil
 }
 
+// UsesKMS returns whether the deployed encryption configuration includes KMS plugins.
+func (c *Config) UsesKMS() bool {
+	return c != nil && len(c.KMSPlugins) > 0
+}
+
+// EncryptedGroupResources returns group resources from the deployed encryption configuration.
+func (c *Config) EncryptedGroupResources() []schema.GroupResource {
```
ardaguclu: Instead of new function, I think it is better to use `ToEncryptionState` by querying key Secrets.

vendor/github.com/openshift/library-go/pkg/operator/encryption/controllers/encryption_rotation_controller.go
tjungblu: thanks Arda, much appreciated. Can you leave this here? https://github.com/openshift/library-go/pull/2237

This is just mostly slopware to see whether this kinda works.

ardaguclu: I'll extensively review https://github.com/openshift/library-go/pull/2237 tomorrow

tjungblu: it's the same thing you just reviewed ;)

ardaguclu: I didn't know that it is ready for review :), I'll allocate sufficient time for it

PR #926, opened 2026-06-19
CNTRLPLANE-2589: Migrate test/e2e-encryption-perf to OpenShift Tests Extension framework

ropatil010:
Hi Team,

Splitting the original PR: https://github.com/openshift/cluster-authentication-operator/pull/859 into smaller once to better debug and merge the PR's easily. 

PR about: 
This commit migrates the encryption performance tests to support both traditional go test and the OpenShift Tests Extension (OTE) framework, following the dual-mode pattern established in PR #859.

Changes:
- Add encryption_perf.go with Ginkgo test registration
- Refactor encryption_perf_test.go to call shared test function
- Add test/library/encryption_wrappers.go for testing.TB compatibility
- Add test/library/encryption/scenarios.go with core test implementations
- Add test/library/encryption/perf_helpers.go for performance test utilities
- Register encryption-perf suite in OTE main.go with ClusterStability: Disruptive

The tests now use testing.TB interface for framework compatibility and context.Background() instead of t.Context() to avoid Ginkgo panics.

Suite details:
- Name: openshift/cluster-authentication-operator/operator-encryption-perf/serial
- Parallelism: 1 (serial execution)
- ClusterStability: Disruptive (triggers API server rollouts)

Test: [sig-auth] authentication operator [Encryption][Serial] TestPerfEncryptionTypeAESCBC

## Summary by CodeRabbit

* **New Features**
  * Added a new encryption performance test suite to broaden coverage for OAuth API server encryption scenarios.
  * Introduced a long-running performance check that measures encryption migration behavior and token creation throughput.

* **Bug Fixes**
  * Improved validation to detect slow encryption migrations and fail when they exceed the expected time limit.

### DIFF
--- a/test/e2e-encryption-perf/encryption_perf.go
+++ b/test/e2e-encryption-perf/encryption_perf.go
@@ -1,0 +1,112 @@
+package e2e_encryption_perf
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	g "github.com/onsi/ginkgo/v2"
+	"github.com/stretchr/testify/require"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	configv1 "github.com/openshift/api/config/v1"
+	oauthapiv1 "github.com/openshift/api/oauth/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	oauthclient "github.com/openshift/client-go/oauth/clientset/versioned/typed/oauth/v1"
+	operatorlibrary "github.com/openshift/cluster-authentication-operator/test/library"
+	operatorencryption "github.com/openshift/cluster-authentication-operator/test/library/encryption"
+	library "github.com/openshift/library-go/test/library/encryption"
+)
+
+const (
+	tokenStatsKey = "created oauthaccesstokens"
+)
+
+var _ = g.Describe("[sig-auth] authentication operator", func() {
+	g.It("[Encryption][Serial] TestPerfEncryptionTypeAESCBC", func(ctx context.Context) {
+		testPerfEncryptionTypeAESCBC(ctx, g.GinkgoTB())
+	})
+})
+
+func testPerfEncryptionTypeAESCBC(ctx context.Context, tt testing.TB) {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Minute)
+	tt.Cleanup(cancel)
+	clientSet := getPerfClients(tt)
+	library.TestPerfEncryption(ctx, tt, library.PerfScenario{
+		BasicScenario: library.BasicScenario{
+			Namespace:                       "openshift-config-managed",
+			LabelSelector:                   "encryption.apiserver.operator.openshift.io/component" + "=" + "openshift-oauth-apiserver",
+			EncryptionConfigSecretName:      fmt.Sprintf("encryption-config-%s", "openshift-oauth-apiserver"),
+			EncryptionConfigSecretNamespace: "openshift-config-managed",
+			OperatorNamespace:               "openshift-authentication-operator",
+			TargetGRs:                       library.AuthTargetGRs,
+			AssertFunc:                      library.AssertTokens,
+		},
+		GetOperatorConditionsFunc: func(t testing.TB) ([]operatorv1.OperatorCondition, error) {
+			apiServerOperator, err := clientSet.OperatorClient.Get(ctx, "cluster", metav1.GetOptions{})
+			if err != nil {
+				return nil, err
+			}
+			return apiServerOperator.Status.Conditions, nil
+		},
+		AssertDBPopulatedFunc: func(t testing.TB, errorStore map[string]int, statStore map[string]int) {
+			require.Empty(t, errorStore, "db loader workers reported errors")
+
+			tokenCount, ok := statStore[tokenStatsKey]
+			require.True(t, ok, "missing oauth access tokens count stats")
+			require.GreaterOrEqual(t, tokenCount, 14000)
+			t.Logf("Created %d access tokens", tokenCount)
+		},
+		AssertMigrationTime: func(t testing.TB, migrationTime time.Duration) {
+			t.Logf("migration took %v", migrationTime)
+			expectedMigrationTime := 10 * time.Minute
+			if migrationTime > expectedMigrationTime {
+				t.Errorf("migration took too long (%v), expected it to take no more than %v", migrationTime, expectedMigrationTime)
+			}
+		},
+		DBLoaderWorkers: 3,
+		DBLoaderFunc: library.DBLoaderRepeat(1, false,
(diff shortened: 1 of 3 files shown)


ropatil010:
/assign @liouk
/retest
Failure CI profiles:
1. e2e-agnostic the failures are not related to my PR. 
2. e2e-agnostic-upgrade the failure case is consistent across other builds as well nd is not related to my PR. 
How ever i'm retriggering to double confirm. 

PASS logs for migrated case 
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/926/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-encryption-perf-serial-ote-1of2/2067937008877047808/artifacts/e2e-aws-operator-encryption-perf-serial-ote/openshift-e2e-test/artifacts/e2e.log

passed: (14m23s) 2026-06-19T13:01:08 "[sig-auth] authentication operator [Encryption][Serial] TestPerfEncryptionTypeAESCBC"
```

test/library/encryption/perf_helpers.go
```diff
@@ -0,0 +1,186 @@
...
+// controller progressing condition in a background goroutine.
+func watchForMigrationControllerProgressingConditionAsync(ctx context.Context, t testing.TB, getOperatorCondFn library.GetOperatorConditionsFuncType, migrationStartedCh chan time.Time, testStartTime time.Time) {
+	t.Helper()
+	go watchForMigrationControllerProgressingCondition(ctx, t, getOperatorCondFn, migrationStartedCh, testStartTime)
+}
+
+// watchForMigrationControllerProgressingCondition waits for the EncryptionMigrationControllerProgressing
+// condition to be set to true with reason "Migrating" and a fresh transition timestamp.
+// It validates that the condition transition occurred after the test start time to avoid accepting stale conditions.
+func watchForMigrationControllerProgressingCondition(ctx context.Context, t testing.TB, getOperatorConditionsFn library.GetOperatorConditionsFuncType, migrationStartedCh chan time.Time, testStartTime time.Time) {
```
liouk: It looks like this func is used to replace the one existing in library-go, in `vendor/github.com/openshift/library-go/test/library/encryption/perf_helpers.go:22` -- however this one checks `cond.Reason == "Migrating"` and also `cond.LastTransitionTime.Time.After(testStartTime)`, but the lib-go version only checks `cond.Status == ConditionTrue`. If the controller sets the `EncryptionMigrationControllerProgressing` condition to `True` with a reason other than `"Migrating"`, the local watcher will never fire, the `time.After(waitPollTimeout)` branch will trigger after 79 minutes, and the test will report `"failed to observe when the migration has started"`, a false failure that the original code would not have hit.

I'd suggest we stick to the same conditions.

test/library/encryption/scenarios.go
```diff
@@ -0,0 +1,247 @@
...
+// TestPerfEncryption tests encryption performance.
+// This is a local implementation that accepts testing.TB instead of *testing.T.
+// It populates the database with test data, enables encryption, and measures migration time.
+func TestPerfEncryption(tb testing.TB, scenario library.PerfScenario) {
+	tb.Logf("Starting encryption performance test for %q provider", scenario.EncryptionProvider.Type)
+
+	migrationStartedCh := make(chan time.Time, 1)
+
+	// Create a cancelable context for the watcher goroutine to ensure it stops when the test finishes
+	watcherCtx, cancel := context.WithCancel(context.Background())
```
liouk: The caller (`testPerfEncryptionTypeAESCBC`) creates a 30min timeout context, which is ignored here as we're creating a new background context. Can we accept context as an arg in `TestPerfEncryption` so that we pass it through instead of a new context?

test/e2e-encryption-perf/encryption_perf.go
```diff
@@ -0,0 +1,119 @@
...
+	operatorencryption "github.com/openshift/cluster-authentication-operator/test/library/encryption"
+	library "github.com/openshift/library-go/test/library/encryption"
+)
+
+const (
+	tokenStatsKey = "created oauthaccesstokens"
+)
+
+var _ = g.Describe("[sig-auth] authentication operator", func() {
+	g.It("[Encryption][Serial] TestPerfEncryptionTypeAESCBC", func() {
```
liouk: Instead of `func()`, we should pass in a context and reuse it to create child contexts in test funcs below.

test/library/encryption_wrappers.go
liouk, line 1: Unless I'm missing something, all of the wrappers defined in this file are simply pass-through calls, they don't do any kind of adaptation. Callers can call `localEncryption.Test*` funcs directly.

test/e2e-encryption-perf/encryption_perf.go
```diff
@@ -0,0 +1,119 @@
...
+
+			tokenCount, ok := statStore[tokenStatsKey]
+			if !ok {
+				err := errors.New("missing oauth access tokens count stats, can't continue the test")
+				require.NoError(t, err)
+			}
+			if tokenCount < 14000 {
+				err := fmt.Errorf("expected to create at least 14000 tokens but %d were created", tokenCount)
+				require.NoError(t, err)
+			}
```
liouk: We're creating errors just to require-and-fail -- this pattern is hard to read. We could require the desired outcome directly, i.e. something like:

```suggestion
			require.True(t, ok, "missing oauth access tokens count stats")
			require.GreaterOrEqual(t, tokenCount, 14000)
```

liouk (changes requested):
Overall, it looks like we're copying and adapting quite a lot of functions from library-go mainly because we're changing `*testing.T` to `testing.TB`. Is there a plan to eventually update these funcs in library-go as well, so that we don't have to copy code over for this change? If we're doing a global OTE migration, this should definitely be part of it (I'd even argue that we should have started there). This way we can reduce code duplication just for plumbing needs.

ropatil010:
> Overall, it looks like we're copying and adapting quite a lot of functions from library-go mainly because we're changing `*testing.T` to `testing.TB`. Is there a plan to eventually update these funcs in library-go as well, so that we don't have to copy code over for this change? If we're doing a global OTE migration, this should definitely be part of it (I'd even argue that we should have started there). This way we can reduce code duplication just for plumbing needs.

Thanks for the feedback, how ever i have copied the same contents of PR: https://github.com/openshift/cluster-authentication-operator/pull/859 in this PR without any modifications. 

Initiated discussion here: https://redhat-internal.slack.com/archives/C07RDCVEYJG/p1782475409270789

ropatil010:
Hi @liouk Merged the dependency PR: [#2335](https://github.com/openshift/library-go/pull/2335) 
Updated this PR accordingly. PTAL. Notified here as well: https://redhat-internal.slack.com/archives/CC3CZCQHM/p1782816690879119?thread_ts=1781885395.465479&cid=CC3CZCQHM

liouk:
The library-go changes which lead to dropping the wrappers simplified this PR a lot -- Thanks @ropatil010. 

/approve
/lgtm

PR #928, opened 2026-06-22
NO-JIRA: add kms preflight to NP to allow egress

tjungblu:
## Summary by CodeRabbit

* **Chores**
  * Added a Kubernetes `NetworkPolicy` named `allow-kms-preflight` to govern network traffic for KMS preflight workloads with egress-only rules.
  * Updated the OAuth API Server operator and its resource output so the new `NetworkPolicy` is included and listed correctly.
* **Tests**
  * Refreshed expected output fixtures and “Create” event entries to reflect the new `NetworkPolicy`.


bindata/oauth-apiserver/networkpolicy_oauth-apiserver.yaml
```diff
@@ -67,3 +67,18 @@ spec:
     # allow all egress traffic
     # required for egress to kube-apiserver pods
     - {}
+---
```
dusk125: Can you roll this into its own file, if memory serves, the StaticResourceController doesn't handle multiple definitions in a single file correctly; so if this is changed, then it won't get automatically seen by the controller.

PR #934, opened 2026-06-26
OCPBUGS-87480: Updating ose-cluster-authentication-operator-container image to be consistent with ART for 5.0

liouk:
Replaces #910

## Summary by CodeRabbit

* **Chores**
  * Updated build and CI configurations to use newer OpenShift/RHEL 9 release images.
  * Upgraded the Go toolchain version to 1.26.
  * Refreshed the container runtime base image to match the newer platform release.

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-9-release-golang-1.25-openshift-4.22
+  tag: rhel-9-release-golang-1.26-openshift-5.0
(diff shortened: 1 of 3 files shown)


.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-9-release-golang-1.25-openshift-4.22
+  tag: rhel-9-release-golang-1.26-openshift-5.0
```
ehearne-redhat, line 4: I might be wrong, but does OpenShift 5.0 run with RHEL 10?

liouk: AFAIU, this refers to the CI environment that will build the operator -- I don't see why RHEL 10 wouldn't work but I'm guessing that the ART haven't transitioned fully to 10.

This was taken as-is from the automated ART PR: https://github.com/openshift/cluster-authentication-operator/pull/910/changes

liouk:
Self-approving, routine ART-consistency PR.

/approve

PR #946, opened 2026-07-07
CNTRLPLANE-3762: Add support for component-scoped proxy

tchap:
## Add component-scoped proxy support for authentication

Adds support for a per-component proxy configuration on the `operator.openshift.io/v1 Authentication` CR, gated behind `FeatureGateAuthenticationComponentProxy`. When enabled, the authentication operator reads `spec.proxy` (HTTPProxy, HTTPSProxy, NoProxy, TrustedCA) from the Authentication CR instead of relying on the cluster-wide proxy environment variables.

Enhancement: [Proxy support for integrated auth stack](https://github.com/openshift/enhancements/blob/master/enhancements/authentication/proxy-support-for-integrated-auth-stack.md)

### Motivation

In disconnected or restricted-network environments, the cluster-wide proxy may not be appropriate for authentication traffic. A component-scoped proxy allows administrators to route OAuth/OIDC and IdP traffic through a dedicated proxy with its own CA trust, independent of the global proxy configuration.

### Changes

**Cluster-wide proxy**: no longer loaded from the `proxies.config.openshift.io/cluster` API resource. `ResolveProxy` reads it from the process environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`), which are already injected by the cluster proxy controller. This removes the direct API dependency and the `proxyLister`/`proxyInformer` from the deployment controller.

**Transport layer** (`pkg/transport`)
- `TransportForCARefWithProxy`: builds proxy-aware HTTP transports with configurable CA references.
- `LoadCAData`: exported helper that reads CA bundle bytes from a ConfigMap, checking `Data` first then `BinaryData`.

**Proxy resolution** (`pkg/controllers/common`)
- `ResolveProxy`: reads the component proxy from the Authentication CR when the feature gate is enabled; falls back to `httpproxy.FromEnvironment()` otherwise. Returns a `ResolvedProxy` with the effective proxy configuration and optional `TrustedCAName`.

**Config observer** (`pkg/controllers/configobservation/oauth`)
- IDP transport construction uses the resolved proxy settings and appends the proxy trusted CA to CA references.
- New `ObserveProxyTrustedCA` observer injects the proxy CA bundle file path into the observed config when a trusted CA is configured.

**Deployment controller** (`pkg/controllers/deployment`)
- Injects `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` env vars from the resolved proxy into the OAuth Server pod.
- Syncs the trusted CA ConfigMap from `openshift-config` to `openshift-authentication`.
- Adds a volume/mount with `Optional: true` for CA hot-reload without redeployment.

**Proxy validation controller** (`pkg/controllers/proxyconfig`)
- Validates route reachability through the resolved proxy (component or cluster-wide).
- Tests IdP endpoint connectivity and emits `IdPEndpointUnreachable` warning events for transient failures (does not set Degraded).
- Uses hash-based change detection to avoid redundant validation.

**Endpoint accessible controller** (`pkg/libs/endpointaccessible`)
- Accepts an optional proxy function for route health checks. Only the route check controller uses it (external hostname); service and endpoint checks remain proxy-free (cluster-internal addresses).

**Custom route controller** (`pkg/controllers/customroute`)
- Route availability health check uses the resolved proxy and loads the proxy trusted CA via `LoadCAData`.

### Test plan

Unit tests are in place, but e2e tests are still missing.

## Summary by CodeRabbit

* **New Features**
  * OAuth server and route/endpoint accessibility checks now support component-scoped proxy settings (with component-scoped trusted CA support) in addition to cluster-wide proxies.
  * Proxy validation and identity-provider discovery now use CA-aware, proxy-aware transports and deterministic `noProxy` merging.
* **Bug Fixes**
  * When component proxy configuration can’t be read, behavior now safely falls back to cluster-wide proxy settings.
  * Proxying is disabled when no effective proxy is set, and identity-provider reachability warnings are aggregated while avoiding unnecessary revalidation.
* **Tests**
  * Added unit coverage for proxy resolution, request proxy selection, and IdP connectivity event behavior.


pkg/transport/transport_test.go
```diff
@@ -0,0 +1,372 @@
...
+	"crypto/x509"
+	"encoding/pem"
+	"net/http"
+	"net/url"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/stretchr/testify/require"
```
ehearne-redhat, line 10: I'm wondering if we should follow assertions instead of relying on an external dependency to save a few lines of code. Especially since the other test files you have created don't use this convention. 

I'm thinking we should remove this to keep things consistent. If we need to write the additional lines for clarity, that should be okay. :)

pkg/transport/transport_test.go
```diff
@@ -0,0 +1,372 @@
...
+func TestTransportForCARef(t *testing.T) {
+	_, caPEM := makeSelfSignedCA(t)
+	_, extraPEM := makeSelfSignedCA(t)
+	emptyLister := newConfigMapLister()
+
+	ref := func(name, key string) CAReference { return CAReference{ConfigMapName: name, ConfigMapKey: key} }
+
+	t.Run("no refs no proxy returns default transport", func(t *testing.T) {
+		rt, err := TransportForCARef(emptyLister, nil, "", "", "")
+		require.NoError(t, err)
```
ehearne-redhat:
```suggestion
		if err != nil { t.Fatalf("unexpected error: %v", err) } 
```

Example change.

test-data/apply-configuration/overall/minimal-cluster/expected-output/UserWorkload/Create/cluster-scoped-resources/certificates.k8s.io/certificatesigningrequests/9806-body-system-COLON-openshift-COLON-openshift-authenticator-.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
...
+  request: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURSBSRVFVRVNULS0tLS0KTUlJQkRUQ0J0QUlCQURCU01WQXdUZ1lEVlFRREUwZHplWE4wWlcwNmMyVnlkbWxqWldGalkyOTFiblE2YjNCbApibk5vYVdaMExXOWhkWFJvTFdGd2FYTmxjblpsY2pwdmNHVnVjMmhwWm5RdFlYVjBhR1Z1ZEdsallYUnZjakJaCk1CTUdCeXFHU000OUFnRUdDQ3FHU000OUF3RUhBMElBQkJZQ2wyaXdqamgwS2xDb0JhbklLRVB4SXVaOEFuNFcKNm96TGNQdWhKWDkwSmgrQzVmLzVyWEtZRkZRdFFrMnZ2QlJYSUhyNlhJUVhYUlZWUXNOMnA4bWdBREFLQmdncQpoa2pPUFFRREFnTklBREJGQWlFQXFDZTBjOFVEQXRPdlVQSjhvRTZXcUFqdk9BUkIrakxOVG5GR3dDNFdnY1VDCklHeU15TGY2YVVTRUVERmUvQ2F5Vm5nQ2o2bDJ5a1JBLzJuckhZc0hpYUpsCi0tLS0tRU5EIENFUlRJRklDQVRFIFJFUVVFU1QtLS0tLQo=
```
ehearne-redhat, line 8: Why does the certificate need to change?

ehearne-redhat (commented):
Overall things look good - another concern that I cannot pinpoint here is if there is a mechanism we can use to interact with oauth-server to ensure that it "does what it says on the tin" i.e. hot reload. 

Maybe not something we can effectively do yet until the oauth-server logic is implemented, but definitely something to note. :)

tchap:
> Overall things look good - another concern that I cannot pinpoint here is if there is a mechanism we can use to interact with oauth-server to ensure that it "does what it says on the tin" i.e. hot reload.
> 
> Maybe not something we can effectively do yet until the oauth-server logic is implemented, but definitely something to note. :)

Yeah, and this must be handled in e2e tests.

pkg/transport/transport_test.go
tchap: I was not paying extra attention to the library being used, but it's not a new dependency and it's being used across the code base, so I would not worry too much about it.

But the import block is badly formatted 🥲

test-data/apply-configuration/overall/minimal-cluster/expected-output/UserWorkload/Create/cluster-scoped-resources/certificates.k8s.io/certificatesigningrequests/9806-body-system-COLON-openshift-COLON-openshift-authenticator-.yaml
ehearne-redhat: You may have missed this comment @tchap :)

tchap: Seems it's regenerated every time you update test fixtures, which was necessary here.

tchap: I actually haven't but I am doing too many things at once 🥲

tchap:
/retitle [CNTRLPLANE-3762](https://redhat.atlassian.net/browse/CNTRLPLANE-3762): Add support for component-scoped proxy

We are gonna merge this without e2e tests.

pkg/controllers/common/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+// proxy. Auth components connect to internal services via DNS names covered by
+// .svc and .cluster.local, so network CIDRs and api-int hostname are not needed.
+var staticNoProxyEntries = []string{".cluster.local", ".svc", "127.0.0.1", "localhost"}
+
+// mergeNoProxy combines user-provided noProxy entries with static cluster-internal
+// defaults. It performs deduplication, but the items are not sorted.
+func mergeNoProxy(userNoProxy []string) string {
+	entries := sets.New[string](staticNoProxyEntries...)
+	entries.Insert(userNoProxy...)
+	return strings.Join(entries.UnsortedList(), ",")
```
everettraven: We probably want to have deterministic ordering here to prevent oauth-server deployment churn based on changes to the ordering within this list that would result in a different environment variable value.

pkg/controllers/common/proxy_test.go
```diff
@@ -0,0 +1,383 @@
...
+
+func TestResolveProxy_FeatureGateError(t *testing.T) {
+	ch := make(chan struct{})
+	fga := featuregates.NewHardcodedFeatureGateAccessForTesting(nil, nil, ch, errors.New("not yet observed"))
+
+	_, err := ResolveProxy(fga, newOperatorAuthLister(nil))
+	if err == nil || err.Error() != "failed to get current feature gates: not yet observed" {
+		t.Fatalf("expected feature gate error, got: %v", err)
+	}
+}
```
everettraven: Looks like this is already tested as part of a table driven test here: https://github.com/openshift/cluster-authentication-operator/pull/946/changes#diff-f7a02cebac7f0f63409aaf693b9b9b8f3fd63ee4997c5a3be85a29426f878ed3R73-R78

pkg/controllers/common/proxy_test.go
```diff
@@ -0,0 +1,383 @@
...
+
+func sortedCSV(s string) string {
+	if s == "" {
+		return ""
+	}
+	parts := strings.Split(s, ",")
+	sort.Strings(parts)
+	return strings.Join(parts, ",")
+}
```
everettraven: Making the return value of the noproxy string deterministic would allow us to remove this.

If we are having to do this in testing, I'm certain we will see some kind of churn associated with the non-determinism without carrying more sorting complexity elsewhere as well.

everettraven (commented):
Leaving the review I've gotten to so far.

There is much more to try and connect here than I had initially thought. I'm going to try and step through the changes a bit more intentionally tomorrow morning with a fresh set of eyes.

pkg/controllers/common/proxy.go
tchap: Aligned in 053f60ffa

pkg/controllers/common/proxy_test.go
tchap: Removed in 053f60ffa

pkg/controllers/common/proxy_test.go
tchap: Removed in 053f60ffa

pkg/controllers/common/proxy.go
```diff
@@ -0,0 +1,105 @@
...
+// ProxyFunc returns a function that resolves the proxy URL for a given request URL.
+func (p *ResolvedProxy) ProxyFunc() transport.ProxyFunc {
+	return transport.ProxyFunc(p.Config.ProxyFunc())
+}
+
+// ResolveProxy reads the component-scoped proxy from the Authentication
+// operator CR (when the feature gate is enabled) and returns the effective proxy
+// settings. When no component proxy is configured, it falls back to the process
+// environment (which reflects the cluster-wide proxy).
+func ResolveProxy(
```
everettraven, line 40: Having gone through a lot of the code changes here, I wonder if making this a struct that contains a method equivalent to this `ResolveProxy` function would be a better path that reduces the amount of changes a bit further?

As an example, as part of the external OIDC work, we added https://github.com/openshift/cluster-authentication-operator/blob/8c68999bf7b5d6c7aaf3937bb897db5ca43c2132/pkg/controllers/common/external_oidc.go#L23-L49

Which allowed us to minimally change function signatures to accept that new struct (or an interface that struct satisfies - which is my preference for mocking in unit tests) and avoid having to make changes to things like the listers struct to add the feature gate accessor. Instead, we instantiate it once in `starter.go` like https://github.com/openshift/cluster-authentication-operator/blob/8c68999bf7b5d6c7aaf3937bb897db5ca43c2132/pkg/operator/starter.go#L138-L144 and pass it to the things that need it.

pkg/controllers/configobservation/oauth/idp_conversions.go
```diff
@@ -19,9 +19,12 @@ import (
 	osinv1 "github.com/openshift/api/osin/v1"
 
 	"github.com/openshift/cluster-authentication-operator/pkg/operator/datasync"
-	"github.com/openshift/cluster-authentication-operator/pkg/transport"
 )
 
+// transportForCABuilderFunc builds an http.RoundTripper for a given CA ConfigMap
+// reference. The cmLister and any proxy settings are captured in the closure.
+type transportForCABuilderFunc func(caConfigMapName, key string) (http.RoundTripper, error)
```
everettraven: Is there a particular reason why this transport builder function needs to be scoped specifically to this configobserver?

I had anticipated that we would probably want a more generic transport builder implementation that things call out to to get the HTTP transport configuration / roundtripper needed for the client.

tchap: The IDP observer is special because it needs to merge the IdP CA with the proxy CA, whereas other call sites only need the proxy CA directly. But I agree the builder pattern could be generalized.

pkg/controllers/common/proxy.go
tchap: I will look into this probably when I finish some of the e2e tests, too much context switching.

I agree this would be nice. I probably ended up with the messy approach because I was implementing things gradually and didn't get to the final refactoring.

tchap: I refactored the changes in dbef8867b4b435f2ee3c65923c12b5e683283732

pkg/controllers/configobservation/oauth/idp_conversions.go
tchap: I refactored the changes in dbef8867b4b435f2ee3c65923c12b5e683283732

pkg/controllers/common/proxy_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/openshift/cluster-authentication-operator/pkg/transport"
+)
+
+// transportConfig holds the cert pool being assembled by TransportOptions.
+type transportConfig struct {
+	pool *x509.CertPool
+}
+
+func (c *transportConfig) appendToPool(name string, data []byte) error {
```
everettraven: Do we need the `name` parameter here? I suspect we could just return the error without the name and wrap it from the caller to include the name information as it bubbles up the error?

pkg/controllers/common/proxy_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	if c.pool == nil {
+		c.pool = x509.NewCertPool()
+	}
+	if ok := c.pool.AppendCertsFromPEM(data); !ok {
+		return fmt.Errorf("unable to parse CA bundle from %s", name)
+	}
```
everettraven: It looks like `AppendCertsFromPEM` returns `true` if _any_ certificates were successfully appended.

I wonder if we should have more explicit parsing + adding functionality for the certificates to ensure that we aren't ever creating clients with unexpected cert pools?

I feel like as-is this might make it more difficult to debug custom CA certificate related issues than if we were to fail out due to more strict parsing validation.

pkg/controllers/common/proxy_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	Err       error
+}
+
+func (f *FakeProxyResolver) ResolveProxy() (*ResolvedProxy, error) {
+	return f.Proxy, f.Err
+}
+
+func (f *FakeProxyResolver) NewTransport(opts ...TransportOption) (*http.Transport, error) {
+	return f.Transport, f.Err
+}
```
everettraven: More of a nit, but in general I think the recommended pattern is for fakes to be in a separate sub-package (i.e `common/fake`).

pkg/controllers/common/proxy_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	operatorAuthLister operatorv1listers.AuthenticationLister,
+	configMapLister corelistersv1.ConfigMapLister,
+	featureGateAccessor featuregates.FeatureGateAccess,
+) AuthProxyResolver {
+	return AuthProxyResolver{
+		operatorAuthLister:  operatorAuthLister,
+		configMapLister:     configMapLister,
+		featureGateAccessor: featureGateAccessor,
+	}
+}
```
everettraven: Can the tests just utilize a fake informer backed by a manually created indexer?

pkg/controllers/common/proxy_test.go
```diff
@@ -0,0 +1,374 @@
...
+				ObjectMeta: metav1.ObjectMeta{Name: "cluster"},
+				Spec: operatorv1.AuthenticationSpec{
+					Proxy: operatorv1.AuthenticationProxyConfig{
+						HTTPProxy: "http://proxy.corp.example.com:3128",
+					},
+				},
+			}),
+			reqURL:       httpURL,
+			wantProxyURL: "http://proxy.corp.example.com:3128",
+		},
```
everettraven, line 300: Should we also test that the opposite scheme for these are not proxied? i.e for HTTPS proxy configured the `http://` prefixed URLs do not get proxied?

pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -91,6 +97,8 @@ func NewConfigObserver(
 		OAuthLister_:         configInformer.Config().V1().OAuths().Lister(),
 		ResourceSync:         resourceSyncer,
 		PreRunCachesSynced:   preRunCacheSynced,
+
+		ProxyResolver: proxyResolver,
```
everettraven: Is there a particular reason that these need to be added?

pkg/controllers/deployment/deployment_controller.go
```diff
@@ -60,11 +67,14 @@ type oauthServerDeploymentSyncer struct {
 	deployments       appsv1client.DeploymentsGetter
 	deploymentsLister appsv1listers.DeploymentLister
 
-	configMapLister corev1listers.ConfigMapLister
-	secretLister    corev1listers.SecretLister
-	podsLister      corev1listers.PodLister
-	proxyLister     configv1listers.ProxyLister
-	routeLister     routev1listers.RouteLister
+	configMaps            corev1client.ConfigMapsGetter
```
everettraven: Do we need to add a specific ConfigMaps getter here? Are we able to get the configmap we care about from methods available to us on the sourceConfigMapLister?

pkg/controllers/deployment/deployment_controller.go
```diff
@@ -115,8 +131,8 @@ func NewOAuthServerWorkloadController(
 
 	clusterScopedInformers := []factory.Informer{
 		configInformers.Config().V1().Ingresses().Informer(),
-		configInformers.Config().V1().Proxies().Informer(),
```
everettraven, line 118: Do we need to make sure that we are paying attention to this informer as part of the ProxyResolver implementation as well?

pkg/controllers/deployment/deployment_controller.go
```diff
@@ -331,3 +343,66 @@ func setRollingUpdateParameters(controlPlaneCount int32, deployment *appsv1.Depl
...
+	if err != nil {
+		return fmt.Errorf("failed to get component proxy CA configmap \"%s/%s\": %w", "openshift-config", trustedCAName, err)
+	}
+
+	targetCM := corev1ac.ConfigMap(componentProxyCAConfigMapName, "openshift-authentication").WithData(sourceCM.Data)
+	if _, err := c.configMaps.ConfigMaps("openshift-authentication").Apply(
+		ctx, targetCM, metav1.ApplyOptions{FieldManager: c.name, Force: true},
+	); err != nil {
+		return fmt.Errorf("failed to apply component proxy CA configmap \"%s/%s\": %w", "openshift-authentication", componentProxyCAConfigMapName, err)
+	}
```
everettraven, line 367: For syncing the proxy configmap, I wonder if we can re-use the ResourceSyncController that is available here https://github.com/openshift/cluster-authentication-operator/blob/482d82f81fc7b460eb3f4024c7fcb4c4a841aecd/pkg/operator/starter.go#L102 ?

can be used something like https://github.com/openshift/cluster-authentication-operator/blob/482d82f81fc7b460eb3f4024c7fcb4c4a841aecd/pkg/operator/starter.go#L115-L121 to sync a configmap from one namespace to another.

pkg/controllers/deployment/deployment_controller.go
```diff
@@ -331,3 +343,66 @@ func setRollingUpdateParameters(controlPlaneCount int32, deployment *appsv1.Depl
...
+
+	deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, corev1.Volume{
+		Name: componentProxyCAConfigMapName,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: componentProxyCAConfigMapName,
+				},
+				Optional: ptr.To(true),
```
everettraven: Do we actually want this to be optional? What happens if this configmap is not present before rolling out the deployment?

pkg/controllers/proxyconfig/proxyconfig_controller.go
```diff
@@ -72,37 +85,111 @@ func NewProxyConfigChecker(
...
+		case idp.Keystone != nil && len(idp.Keystone.URL) > 0:
+			urls = append(urls, idp.Keystone.URL)
+		case idp.BasicAuth != nil && len(idp.BasicAuth.URL) > 0:
+			urls = append(urls, idp.BasicAuth.URL)
+		case idp.Google != nil:
+			urls = append(urls, "https://accounts.google.com")
+		}
+	}
+	return urls
+}
```
everettraven, line 184: Is there a reason we have to do this now but didn't before? This looks like net-new logic (unless I'm missing this being moved from somewhere else).

pkg/transport/transport_test.go
```diff
@@ -0,0 +1,155 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/user"
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/openshift/cluster-authentication-operator/pkg/internal/transporttest"
+)
+
+func TestNewTransport(t *testing.T) {
```
everettraven, line 21: Is it possible for all of the test cases here to be converted to table driven tests?

pkg/controllers/deployment/deployment_controller.go
tchap: This would IMO only complicate things. RSC is made to be used for static definitions, not to be updated when Authentication CR is changed. Also this would mean we now don't know when the CM is actually sync'd while here we can sync it before updating the deployment. Doesn't feel like bending the interface and creating races is worth the few lines saved, if any.

pkg/controllers/common/proxy_resolver.go
tchap: Aligned in b037c9d100ed337e9c782db2fad8d6d679189633

pkg/controllers/deployment/deployment_controller.go
tchap: Actually the CM is always present, so no, we can remove this. Done in e08a17f53f8cb5c5aa250353d72da1de149e5845

pkg/controllers/common/proxy_resolver.go
tchap: I will look into improving this, but this is an existing pattern in the code base 😐

tchap: Implemented in a9234883b04e1eb5e363baff5afeae094c53eb2c . I only changed the approach in the new code, leaving existing code to use `AppendCertsFromPEM` as this might be a breaking change. There are quite a few call sites, can be refactored later if needed.

pkg/controllers/common/proxy_resolver.go
tchap: Split in b77f5782639039485c33bbf7a356e347a8400f0f

pkg/controllers/common/proxy_resolver.go
tchap: Refactored in a5b1ae924837b668da511a0cc5017267092b2e90

pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
tchap: What do you mean exactly? `ProxyResolver` is used by multiple config observers, so this is the way to pass it in.

pkg/controllers/deployment/deployment_controller.go
tchap: Yeah, this is one thing I changes when implementing the changes. I read proxy config from the environment simply, which is the same config you get from reading proxy config explicitly, it may just be lagging a bit more behind, i.e. until the operator is re-deployed on change. I wanted to unify proxy handling, i.e. when actually creating a transport, you load from the env, so why to do it differently somewhere else.

pkg/transport/transport_test.go
tchap: Converted in c519bddab52fe84c1d1e1ea52f851eb1e03b034d

pkg/controllers/common/proxy_test.go
tchap: Test cases added in 909adda2c269d7852b1b02d165c27acf551c81ca

pkg/controllers/deployment/deployment_controller.go
tchap: We are writing and also deleting ConfigMaps, so no, we can't just use a lister. This is for syncing `trustedCA`.

pkg/controllers/proxyconfig/proxyconfig_controller.go
tchap: It is new, but it's mentioned in the EP: 

https://github.com/openshift/enhancements/blob/master/enhancements/authentication/proxy-support-for-integrated-auth-stack.md#operator-process

An event is to be emitted when the IdP is unreachable via the proxy.

pkg/controllers/deployment/deployment_controller.go
everettraven: >RSC is made to be used for static definitions, not to be updated when Authentication CR is changed.

Where do you see that it is made to be used for static definitions?

I've seen this used to sync something conditionally based on the observed state of something in the past. We actually do this within this operator and it is done as part of the encryption controller logic.

>Also this would mean we now don't know when the CM is actually sync'd while here we can sync it before updating the deployment.

I don't think we need to perform the full sync to know when the configmap is actually synced. Realistically, all we care about is that the configmap exists within the namespace we are creating the deployment in so it can be referenced as a volume mount.

> Doesn't feel like bending the interface and creating races is worth the few lines saved, if any

It isn't really about saving a few lines. Instead it is about using the common functionality that was designed specifically for syncing configuration resources from one namespace to another that has been used for many years by controller running in production.

pkg/controllers/proxyconfig/proxyconfig_controller.go
everettraven: Ah, thanks for the reminder - I had forgotten that we added the bit about actually hitting the IdP endpoints as well instead of just hitting the oauth-server routes.

pkg/controllers/configobservation/configobservercontroller/observe_config_controller.go
everettraven: I'm just hesitant to add things to the "listers" struct that are not really related to listers.

How we have done this in the past when only a handful of observers need some kind of persistent functionality through a reference, we create an initializer function and call that to get the config observer.

For example, this is one that I did in the past for cluster-kube-apiserver-operator: https://github.com/openshift/cluster-kube-apiserver-operator/blob/cb003ede291505e062c3e0523e1d760eaec00339/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L133-L135

tchap: Makes sense, aligned in ea719e956d10264625ac0b1874f2a650c6b818d1

pkg/controllers/deployment/deployment_controller.go
tchap: > Where do you see that it is made to be used for static definitions?
> 
> I've seen this used to sync something conditionally based on the observed state of something in the past. We actually do this within this operator and it is done as part of the encryption controller logic.

Ok, this is not precise. You can dynamically configure RSC. You can pass `emptyResourceLocation` to mark the CM as to be deleted.
 
> I don't think we need to perform the full sync to know when the configmap is actually synced. Realistically, all we care about is that the configmap exists within the namespace we are creating the deployment in so it can be referenced as a volume mount.

Ok, but how do we know that? RSC is asynchronous. So the CM copy doesn't have to exist at the point when we are creating the deployment.
 
> It isn't really about saving a few lines. Instead it is about using the common functionality that was designed specifically for syncing configuration resources from one namespace to another that has been used for many years by controller running in production.

Yeah, I understand this in principle, but I think that passing around one controller into another (can be done via an interface, no problem) and calling the methods instead of a simple API call doesn't seem like too much of an added value to me. We would be reconfiguring RSC in every deployment sync run.

Or were you thinking about placing that sync logic somewhere else? In any case, the async argument still holds IMO. This seems to create more problems than it solves.

everettraven: >Ok, but how do we know that? RSC is asynchronous. So the CM copy doesn't have to exist at the point when we are creating the deployment.

We have a configmap lister for the namespace we are deploying in available to us right? Try to get the configmap we care about and if we get a `NotFound` error back we know it doesn't exist and can decide to not create the deployment yet.

Alternatively, we just rely on eventual consistency and let the deployment set in the progressing state until the pods are able to be created because the configmap now exists.

IMO the additional get call is reasonable enough protection if we want an earlier signal and to use it as a "we should / should not call the sync function on the RSC".

tchap: Implemented in 687c775b72b050b2cb0aa373c5fb0850086ef9f8 . I chose to return an error when the CM is not there yet and let it be swallowed by the Degraded inertia.

pkg/controllers/common/proxy_resolver.go
```diff
@@ -0,0 +1,173 @@
...
+
+// WithCA appends raw CA certificate bytes to the transport's trust pool.
+// The name identifies the source for error messages.
+func WithCA(name string, data []byte) TransportOption {
+	return func(cfg *transportConfig, _ corelistersv1.ConfigMapLister) error {
+		if len(data) == 0 {
+			return nil
+		}
+		if err := cfg.appendToPool(data); err != nil {
+			return fmt.Errorf("%s: %w", name, err)
```
everettraven: Could we make this error message a bit more descriptive? maybe somthing like:

```suggestion
			return fmt.Errorf("appending CA certificates to cert pool for source %q: %w", name, err)
```

pkg/controllers/common/proxy_resolver.go
```diff
@@ -0,0 +1,173 @@
...
+	return func(cfg *transportConfig, cmLister corelistersv1.ConfigMapLister) error {
+		if len(name) == 0 {
+			return nil
+		}
+		data, err := transport.LoadCAData(cmLister, name, key)
+		if err != nil {
+			return err
+		}
+		if err := cfg.appendToPool(data); err != nil {
+			return fmt.Errorf("configmap %q: %w", name, err)
```
everettraven: Similar comment on error descriptiveness.

pkg/controllers/common/proxy_resolver.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	if len(proxy.TrustedCAName) > 0 {
+		proxyCA, err := transport.LoadCAData(r.configMapLister, proxy.TrustedCAName, "ca-bundle.crt")
+		if err != nil {
+			return nil, err
+		}
+		if err := cfg.appendToPool(proxyCA); err != nil {
+			return nil, fmt.Errorf("proxy trustedCA %q: %w", proxy.TrustedCAName, err)
```
everettraven: Similar comment on error descriptiveness

pkg/controllers/configobservation/oauth/observe_idps.go
```diff
@@ -10,80 +10,83 @@ import (
...
 )
 
 var identityProvidersMounts = []string{"volumesToMount", "identityProviders"}
 
-func ObserveIdentityProviders(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
-	identityProvidersPath := []string{"oauthConfig", "identityProviders"}
-	defer func() {
-		ret = configobserver.Pruned(ret, identityProvidersPath, identityProvidersMounts)
-	}()
+func NewObserveIdentityProviders(proxyResolver common.ProxyResolver) configobserver.ObserveConfigFunc {
```
everettraven, line 24: I think it would probably be less disruptive to the code, and more clear, to do something like:

```go
type identityProviderObserver struct {
    proxyResolver common.ProxyResolver
}

func NewObserveIdentityProviders(proxyResolver common.ProxyResolver) configobserver.ObserveConfigFunc {
    return (&identityProviderObserver{
        proxyResolver: proxyResolver,
    }).ObserveIdentityProviders
}

// This is the same as the existing `ObserveIdentityProviders` function, just updated to be a method on `identityProviderObserver`
func (ipo *identityProviderObserver) ObserveIdentityProviders(...) (...) {...}
```

everettraven (commented):
Overall this looks a lot better to me.

A few more smaller comments. Other than these this looks ready to go to me.

pkg/controllers/common/proxy_resolver.go
tchap: Improved in 48b4f7e544b2af8209f5d132576370da33e7745e

pkg/controllers/common/proxy_resolver.go
tchap: Improved in 48b4f7e544b2af8209f5d132576370da33e7745e

pkg/controllers/common/proxy_resolver.go
tchap: Improved in 86f43ca1841944f20399eda411cbf2ec31892520

pkg/controllers/configobservation/oauth/observe_idps.go
tchap: Look better, refactored in 57bb2b67f0951cbb37ad6864dbef10b1b53999d0

pkg/controllers/deployment/deployment_controller.go
tchap: This has been removed as we use RSC now.

tchap:
I am going to block this until I run my e2e tests, just to be sure.

/hold

everettraven (approved):
/lgtm
/approve

tchap:
I run the e2e tests being developed on my end, they passed.

/unhold

tchap:
This can get merged as soon as CI is finished and passing.

/verified by CI

tchap:
This seems unrelated at first.

/retest

PR #950, opened 2026-07-20
[WIP] CNTRLPLANE-3851: Oauth server proxy config e2e

ehearne-redhat:
Continues work from and closes https://github.com/openshift/origin/pull/31398 . Kept in separate file for now. Plan is to add tests to `test/e2e-component-proxy/component_proxy.go` into https://github.com/openshift/cluster-authentication-operator/pull/949 when consensus reached on test status. 

## Summary by CodeRabbit

* **New Features**
  * Added component-scoped HTTP proxy support for OAuth/OIDC and health checks, including feature-gated configuration and no-proxy handling.
  * Added trusted CA bundle support for proxied requests, including syncing into the OAuth server namespace and hot-reload behavior.

* **Bug Fixes**
  * Improved proxy-aware route availability and endpoint accessibility checks, including proper TLS trust and consolidated reachability warning events.
  * Refined rollout behavior to account for resolved proxy settings and avoid unnecessary redeployments.

* **Tests**
  * Expanded component-proxy end-to-end coverage and added unit/integration test cases for proxy resolution, transports, and trusted-CA observation.


test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,553 @@
...
+
+	g.By("Setting only httpsProxy in component proxy config")
+	operatorAuth.Spec.Proxy = operatorv1.AuthenticationProxyConfig{
+		HTTPSProxy: proxyURL,
+	}
+	_, err = clients.OperatorClient.OperatorV1().Authentications().Update(ctx, operatorAuth, metav1.UpdateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.By("Waiting for operator to reconcile proxy config")
+	err = test.WaitForClusterOperatorAvailableNotProgressingNotDegraded(t, clients.ConfigClient.ConfigV1(), "authentication")
```
tchap: I think that in these cases we should use `WaitForOperatorToPickUpChanges`. Actually none of these solutions is bullet-proof, because you can just miss the `Progressing` condition. I am kinda surprised that function actually works since the status polling is done only every 10 seconds.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,553 @@
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.By("Verifying oauth-server has HTTPS_PROXY but not HTTP_PROXY")
+	envVars := test.GetOAuthServerProxyEnvVars(t, clients.KubeClient)
```
tchap: This function is actually now redundant, use `VerifyOAuthServerDeploymentProxyConfig`.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,553 @@
...
+				return true, nil
+			}
+		}
+		t.Logf("expected group %q not found in groups: %v", expectedGroup, ssr.Status.UserInfo.Groups)
+		return false, nil
+	})
+	o.Expect(err).NotTo(o.HaveOccurred(), "OIDC login flow should succeed")
+}
+
+func testPartialProxyEnvVars() {
```
tchap: The proxy env var testing cases are the same, they just differ in `AuthenticationProxyConfig` and checking the OAuth Server env. So I would just create a single function that accepts these two as arguments.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,553 @@
...
+		g.GinkgoWriter.Println("cleaning up: removing proxy NetworkPolicies")
+		networkPolicyCleanup()
+	})
+
+	g.By("Waiting for operator to reconcile proxy config")
+	err = test.WaitForClusterOperatorAvailableNotProgressingNotDegraded(t, clients.ConfigClient.ConfigV1(), "authentication")
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.By("Creating Keycloak test user and group")
+	group := fmt.Sprintf("ocp-test-proxy-login-group")
```
tchap: There is no dynamic param here.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,553 @@
...
+	g.By("Removing component-scoped proxy config")
+	operatorAuth, err = clients.OperatorClient.OperatorV1().Authentications().Get(ctx, "cluster", metav1.GetOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	operatorAuth.Spec.Proxy = operatorv1.AuthenticationProxyConfig{}
+	_, err = clients.OperatorClient.OperatorV1().Authentications().Update(ctx, operatorAuth, metav1.UpdateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.By("Waiting for operator to reconcile proxy removal")
+	err = test.WaitForClusterOperatorAvailableNotProgressingNotDegraded(t, clients.ConfigClient.ConfigV1(), "authentication")
+	o.Expect(err).NotTo(o.HaveOccurred())
```
tchap, line 224: Might make sense to remove the proxy and network policies here to make sure the requests are not going though. Or check that in some other way.

Also `WaitForClusterOperatorAvailableNotProgressingNotDegraded` is yet again probably not the right function here, because it may unblock immediately.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,553 @@
...
+
+	g.By("Performing OIDC login flow via direct IdP connectivity after proxy removal")
+	assertOIDCLogin(t, kubeConfig, *clients, username, password, group)
+}
+
+func testTrustedCAHotReload() {
```
tchap, line 239: I am not sure this test is right. As you rotate the CA, you need to redeploy Squid, which deploys it with another CA. The helper function now actually returns the CA associated with Squid. So you need to deploy Squid and store that CA in the ConfigMap. Then delete and redeploy, get new CA, update ConfigMap and check again.

test/e2e-component-proxy/component_proxy_oidc_login.go
ehearne-redhat: What does `VerifyOAuthServerDeploymentProxyConfig` return that I can use as verification? The above function actually gets the env vars for me to verify. There is no error returned for me to use for verifying. Let me know your thoughts.

tchap: These helpers get `t` as an arg so they just fail for you.

test/e2e-component-proxy/component_proxy_oidc_login.go
ehearne-redhat: Maybe I should also include trustedCA too.

test/e2e-component-proxy/component_proxy_oidc_login.go
ehearne-redhat: Could you clarify what you mean? Are you referring to `fmt.Sprintf()` ? If so, that is actually not there anymore. Or, do you want each test to have a unique username/password/group ?

test/e2e-component-proxy/component_proxy_oidc_login.go
ehearne-redhat: I believe `testDirectIdPFallback()` does not add network policies but it does add a proxy. The idea behind this is that if a user adds a proxy, and then removes it, they should be able to log in successfully. 

I have a check in `testNoProxy()` that might do the job here. We can ensure that there are no entries that contain the proxy hostname.

test/e2e-component-proxy/component_proxy_oidc_login.go
ehearne-redhat: That makes sense. I'll work on that...

test/e2e-component-proxy/component_proxy_oidc_login.go
tchap: Yeah, I was referring to `fmt.Sprintf`.

test/e2e-component-proxy/component_proxy_oidc_login.go
tchap: I mean that removing the proxy and checking login works does not really check how that login is happening, so we should ensure that once proxy is removed, login is not using that proxy actually.

test/e2e-component-proxy/component_proxy_oidc_login.go
ehearne-redhat: Actually, reading this back again I am seeing a problem with that approach. We cannot redeploy squid proxy for the purpose of this test. Doing so requires a complete auth proxy config update with `httpProxy` and would trigger oauth-server redeploy. I would need to have a way to create new CA, attach to existing squid proxy, and then update auth config CR with just the new CA to test this. I'll need to look into this a little further...

tchap: Really? Aren't we using a service name there?

ehearne-redhat: From what I can see with the service name, it is dependent on the namespace. The namespace is randomised. So it requires a full redeployment. 

I have some logic to hopefully combat this: we can recreate the certs, update `tls-squid` secret and tell squid to reconfigure itself in the pod via `oc exec` which should keep the service. We can also restart the pod which should still retain the service name if preferred.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,609 @@
...
+	test.CheckFeatureGateEnabledOrSkip(t, clients.ConfigClient, features.FeatureGateAuthenticationComponentProxy)
+
+	g.By("Waiting for authentication operator to be stable before test")
+	err := test.WaitForClusterOperatorAvailableNotProgressingNotDegraded(t, clients.ConfigClient.ConfigV1(), "authentication")
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.By("Deploying Squid forward proxy")
+	httpProxyURL, _, _, proxyNamespace, proxyCleanup := test.DeploySquidProxy(t, clients.KubeClient)
+	g.DeferCleanup(func() {
+		g.GinkgoWriter.Println("cleaning up: removing Squid proxy")
```
tchap: This is now included in `proxyCleanup`, so you can just do `g.DeferCleanup(proxyCleanup)`

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,609 @@
...
+	})
+	g.It("[Serial][Operator][ComponentProxy] should hot-reload mounted CA file on change when spec.proxy.trustedCA is set", func() {
+		testTrustedCAHotReload()
+	})
+	g.It("[Serial][Operator][ComponentProxy] should bypass proxy for noProxy hosts", func() {
+		testNoProxy()
+	})
+})
+
+func enableDirectAccessGrants(kcClient *test.KeycloakClient, clientID string) {
```
tchap: `nit`: I somehow prefer to put the helpers after tests themselves so that I can firstly see the tests.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,609 @@
...
+
+	username := "proxy-login-test-user"
+	password := "proxy-login-test-password"
+	o.Expect(kcClient.CreateUser(username, "", password, []string{group}, nil)).To(o.Succeed())
+
+	g.By("Performing full OIDC login flow through component proxy")
+	assertOIDCLogin(t, kubeConfig, *clients, username, password, group)
+
+	g.By("Verifying traffic went through the Squid proxy")
+	err = test.WaitForSquidProxyTraffic(t, clients.KubeClient, proxyNamespace, 5*time.Minute)
```
tchap: I am not sure this really needs to be checked and that it's actually checking anything, because the operator does all sorts of things that go through the proxy, like health checks, IdP config reload...

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,609 @@
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+}
+
+func testDirectIdPFallback() {
```
tchap: I am wondering whether we can't just append this to `testProxyOIDCLoginFlow` since there we test the flow with proxy. So we can just remove it after and do the flow again? I am not sure what kind of cleanup we would have to do, perhaps only log out the user?

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,609 @@
...
+
+	g.By("Deploying NetworkPolicy to restrict Keycloak ingress to proxy namespace only")
+	keycloakNamespace := setup.Namespace
+	networkPolicyCleanup := test.DeployProxyNetworkPolicies(t, clients.KubeClient, proxyNamespace, keycloakNamespace)
+	g.DeferCleanup(func() {
+		g.GinkgoWriter.Println("cleaning up: removing proxy NetworkPolicies")
+		networkPolicyCleanup()
+	})
+
+	configMapName := "e2e-proxy-trusted-ca"
```
tchap, line 283: This should be a new `g.By` block, I think.

test/e2e-component-proxy/component_proxy_oidc_login.go
```diff
@@ -0,0 +1,609 @@
...
+	o.Expect(err).NotTo(o.HaveOccurred())
+	cm.Data["ca-bundle.crt"] = string(newCACertPEM)
+	_, err = clients.KubeClient.CoreV1().ConfigMaps("openshift-config").Update(ctx, cm, metav1.UpdateOptions{})
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	g.By("Verifying OIDC login works after CA rotation")
+	assertOIDCLogin(t, kubeConfig, *clients, username, password, group)
+
+	g.By("Verifying traffic went through the Squid proxy")
+	err = test.WaitForSquidProxyTraffic(t, clients.KubeClient, proxyNamespace, 5*time.Minute)
```
tchap: Again not entirely sure this really checks anything since the networkpolicy is there, but I am asking this question myself repeatedly when reviewing and writing these tests...

test/e2e-component-proxy/component_proxy_oidc_login.go
tchap: I guess we can leave it here, it's an additional check, albeit not very precise...

PR #960, opened 2026-07-29
CNTRLPLANE-3762: Clear IdP validation hash when all identity providers are removed

tchap:
When all IdPs were removed from the OAuth config, the stale lastIdPValidationHash was preserved. Re-adding the same IdPs with the same proxy config produced a matching hash, causing validateIdPConnectivity to skip validation entirely.

Clear the hash in the no-IdP early return so re-added IdPs are always validated.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Identity Provider connectivity validation to refresh correctly after all Identity Provider endpoints are removed and later re-added.
  - Prevented stale validation results from being reused after configuration changes.
  - Ensured unreachable endpoint checks run again and generate the expected notification.


everettraven (approved):
/lgtm

PR #964, opened 2026-08-06
CNTRLPLANE-3762: Watch openshift-config ConfigMaps in OAuth route health check controller

tchap:
The OAuthServerRouteEndpointAccessibleController reads the proxy
trusted CA from a ConfigMap in the openshift-config namespace, but
did not watch that namespace's ConfigMap informer as a trigger. This
meant that after a proxy CA rotation, the controller would not
re-sync until its periodic resync (1-2 minutes), delaying recovery.

Add the openshift-config ConfigMap informer to the controller's
trigger list so it re-syncs immediately when the proxy CA content
changes. The informer is already running (used for the lister),
so this adds no new API watches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Proxy trusted-CA configuration changes are now detected automatically, triggering reconciliation when the relevant configuration is updated.


everettraven (approved):
/lgtm
/approve

PR #972, opened 2026-08-18
OCPBUGS-111095: Do not cache non-definitive OIDC token endpoint errors during password grant check

ShazaAldawamneh:
Fix permanent caching of challenge: false when the OIDC token endpoint returns a transient JSON error (e.g. invalid_client_credentials, server_error, temporarily_unavailable) during the dummy ROPC probe
Only cache definitive results: invalid_grant (grants supported), unsupported_grant_type / unauthorized_client (grants not supported); all other JSON errors skip the cache so the next sync loop retries.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OIDC password-grant detection by distinguishing definitive OAuth responses.
  - Correctly allows password grants for `invalid_grant` responses and disallows them for unsupported grant types.
  - Prevents temporary, malformed, or unrecognized errors from being cached, allowing later successful checks to determine the correct state.

### DIFF
--- a/pkg/controllers/configobservation/oauth/idp_conversions.go
+++ b/pkg/controllers/configobservation/oauth/idp_conversions.go
@@ -440,7 +440,15 @@ func checkOIDCPasswordGrantFlow(
 	}
 
 	if errVal, ok := respMap["error"]; ok {
-		oidcPasswordChecks[secret.ResourceVersion] = errVal == "invalid_grant" // wrong password, but password grants allowed
+		switch errVal {
+		case "invalid_grant":
+			oidcPasswordChecks[secret.ResourceVersion] = true
+		case "unsupported_grant_type":
+			oidcPasswordChecks[secret.ResourceVersion] = false
+		default:
+			klog.Warningf("OIDC token endpoint returned non-definitive error %q, not caching result", errVal)
+			return false, nil
+		}
 	} else {
 		_, ok = respMap["access_token"] // in case we managed to hit the correct user
 		oidcPasswordChecks[secret.ResourceVersion] = ok
--- a/pkg/controllers/configobservation/oauth/idp_conversions_test.go
+++ b/pkg/controllers/configobservation/oauth/idp_conversions_test.go
@@ -351,6 +351,79 @@ func TestCheckOIDCPasswordGrantFlowCaching(t *testing.T) {
 		require.True(t, oidcPasswordChecks["test-version-1"])
 	})
 
+	t.Run("non-definitive JSON errors are not cached", func(t *testing.T) {
+		oidcPasswordChecks = map[string]bool{}
+
+		for _, errCode := range []string{"invalid_client_credentials", "server_error", "temporarily_unavailable", "unauthorized_client"} {
+			responseContent = fmt.Sprintf(`{"error": %q}`, errCode)
+			shouldError = false
+			result, err := checkOIDCPasswordGrantFlow(
+				secretLister,
+				&proxyResolver,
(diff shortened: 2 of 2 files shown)


liouk (changes requested):
Overall looks good to me -- just one thing, I noticed there's no test for `unauthorized_client`; this is low risk as it lives in the same switch case as `unsupported_grant_type`, but let's add one for completeness.

pkg/controllers/configobservation/oauth/idp_conversions.go
```diff
@@ -440,7 +440,15 @@ func checkOIDCPasswordGrantFlow(
 	}
 
 	if errVal, ok := respMap["error"]; ok {
-		oidcPasswordChecks[secret.ResourceVersion] = errVal == "invalid_grant" // wrong password, but password grants allowed
+		switch errVal {
+		case "invalid_grant":
+			oidcPasswordChecks[secret.ResourceVersion] = true
+		case "unsupported_grant_type", "unauthorized_client":
```
everettraven: For `unauthorized_client`, I'm not sure this is a permanent failure case. Looking at https://www.rfc-editor.org/info/rfc6749/#section-5.2 it looks like `unauthorized_client` just means that the client as configured is not currently able to utilize this grant type, but the IdP side configuration could be updated such that the client IS able to use it in the future.

We may want this error case to also fall under a non-definitive style of error.

pkg/controllers/configobservation/oauth/idp_conversions.go
```diff
@@ -440,7 +440,15 @@ func checkOIDCPasswordGrantFlow(
...
+			oidcPasswordChecks[secret.ResourceVersion] = false
+		default:
+			klog.Warningf("OIDC token endpoint returned non-definitive error %q (%s), not caching result", errVal, tokenURL)
```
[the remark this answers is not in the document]
everettraven: +1 to omitting the tokenURL from the message, it isn't necessary.

I don't mind the errVal being provided in the message.

pkg/controllers/configobservation/oauth/idp_conversions.go
```diff
@@ -440,7 +440,15 @@ func checkOIDCPasswordGrantFlow(
 	}
 
 	if errVal, ok := respMap["error"]; ok {
-		oidcPasswordChecks[secret.ResourceVersion] = errVal == "invalid_grant" // wrong password, but password grants allowed
+		switch errVal {
+		case "invalid_grant":
+			oidcPasswordChecks[secret.ResourceVersion] = true
+		case "unsupported_grant_type", "unauthorized_client":
+			oidcPasswordChecks[secret.ResourceVersion] = false
+		default:
```
everettraven, line 448: What is the potential impact here if we don't cache the results for at least some amount of time?

How often will we end up trying to re-verify whether or not the password grant type is allowed? How much churn in configuration of the oauth-server will that result in?

ShazaAldawamneh: The non-cached path only stays active while the IdP keeps returning a non-definitive error. Once the IdP recovers (the common case, these are transient), the next probe gets invalid_grant which IS cached, and probing stops. The sync loop interval provides natural rate limiting in the meantime. If the IdP is persistently broken, the repeated probes are arguably desirable, they're how the operator detects recovery without requiring a pod restart or secret change.

everettraven (approved):
/lgtm
/approve

everettraven (approved):
/lgtm

